Issue #7, opened 2023-03-11
DefaultAzureCredential failed to retrieve a token from the included credentials


spedditi:
> Please provide us with the following information:
> ---------------------------------------------------------------

### This issue is for a: (mark with an `x`)
```
- [ ] bug report -> please search issues before submitting
- [ ] feature request
- [ ] documentation issue or request
- [ ] regression (a behavior that used to work and stopped in a new release)
```

### Minimal steps to reproduce
>
When  i run azd up ( part of predocs.ps1),  i am getting following error and search index gptkbindex didn't get created as part of this process. I created gptkbindex manually.  Data files are not getting uploaded to storage account. 
### Any log messages given by the failure
>
DefaultAzureCredential failed to retrieve a token from the included credentials.
Attempted credentials:
        EnvironmentCredential: EnvironmentCredential authentication unavailable. Environment variables are not fully configured.
Visit https://aka.ms/azsdk/python/identity/environmentcredential/troubleshoot to troubleshoot.this issue.
        ManagedIdentityCredential: ManagedIdentityCredential authentication unavailable, no response from the IMDS endpoint.
        SharedTokenCacheCredential: SharedTokenCacheCredential authentication unavailable. No accounts were found in the cache.
        AzureCliCredential: Azure CLI not found on path
        AzurePowerShellCredential: Az.Account module >= 2.2.0 is not installed
To mitigate this issue, please refer to the troubleshooting guidelines here at https://aka.ms/azsdk/python/identity/defaultazurecredential/troubleshoot.
Traceback (most recent call last):
  File "………...Documents\workspace\lia\azure-search-openai-demo\scripts\prepdocs.py", line 208, in <module>
    create_search_index()
  File "………...Documents\workspace\lia\azure-search-openai-demo\scripts\prepdocs.py", line 146, in create_search_index
    if args.index not in index_client.list_index_names():
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "………...AppData\Local\Packages\PythonSoftwareFoundation.Python.3.11_qbz5n2kfra8p0\LocalCache\local-packages\Python311\site-packages\azure\core\paging.py", line 132, in __next__
    return next(self._page_iterator)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #10, opened 2023-03-13
Update prepdocs hook to use AzureDeveloperCliCredential

vhvb1989:
The `prepdocs` hook uses `DefaultAzureCredential`, which might incorrectly use another source of auth for Azure.
For example, DAC might use an environment credential that is set within the system, or the default account set for the `Azure CLI`, if `az` was previously login.

This change makes sure that the hook uses the same auth that was used when provisioning the template, by calling `azd` for an auth token.

The `TenantId` is important to support accounts with access to multiple tenants, as `azd` would return `home-tenant` if no tenant id is provided.

fix: https://github.com/Azure-Samples/azure-search-openai-demo/issues/7


infra/main.bicep
```diff
@@ -248,3 +248,6 @@ output AZURE_SEARCH_SERVICE string = searchServices.outputs.name
 output AZURE_STORAGE_ACCOUNT string = storage.outputs.name
 output AZURE_STORAGE_CONTAINER string = containerName
 output BACKEND_URI string = backend.outputs.uri
+// Tenant id is required by any external app ( like hooks ) to authtenticate 
+// to the same tenant.
```
jongio, line 252: Don't need this comment.

PR #11, opened 2023-03-14
Install script dependencies into virtual environment...

johanste:
...and call python directly to prep documents...

## Purpose
Remove dependency on powershell on posix systems + install doc preparation scripts into a virtual environment.

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[ ] Bugfix
[X] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
```

## What to Check
* ...

### DIFF
--- a/azure.yaml
+++ b/azure.yaml
@@ -8,16 +8,30 @@ services:
     project: ./app/backend
     language: py
     host: appservice
+  scripts:
+    project: ./scripts
+    language: py
+    host: ""
 hooks:
     postprovision:
-      shell: pwsh
-      run: ./scripts/prepdocs.ps1
-      interactive: true
-      continueOnError: false
+      windows:
+        shell: pwsh
+        run: ./scripts/prepdocs.ps1
+        interactive: true
+        continueOnError: false
+      posix:
+        shell: sh
+        run: scripts/scripts_env/bin/python scripts/prepdocs.py 'data/*' --storageaccount $AZURE_STORAGE_ACCOUNT --container $AZURE_STORAGE_CONTAINER --searchservice $AZURE_SEARCH_SERVICE --index $AZURE_SEARCH_INDEX -v
+        interactive: true
+        continueOnError: false
     predeploy:
+      windows:
         shell: pwsh
         run:  cd ./app/frontend;npm install;npm run build
         interactive: true
         continueOnError: false
-
-      
+      posix:
+        shell: sh
+        run:  cd ./app/frontend;npm install;npm run build
+        interactive: true
+        continueOnError: false


azure.yaml
```diff
@@ -8,16 +8,30 @@ services:
...
-      interactive: true
-      continueOnError: false
+      windows:
+        shell: pwsh
+        run: ./scripts/prepdocs.ps1
+        interactive: true
+        continueOnError: false
+      posix:
+        shell: sh
+        run: scripts/scripts_env/bin/python scripts/prepdocs.py 'data/*' --storageaccount $AZURE_STORAGE_ACCOUNT --container $AZURE_STORAGE_CONTAINER --searchservice $AZURE_SEARCH_SERVICE --index $AZURE_SEARCH_INDEX -v
```
jongio, line 24: For consistency, I'd like to create a .sh file for this.

azure.yaml
```diff
@@ -8,16 +8,30 @@ services:
...
         run:  cd ./app/frontend;npm install;npm run build
         interactive: true
         continueOnError: false
-
-      
+      posix:
+        shell: sh
+        run:  cd ./app/frontend;npm install;npm run build
+        interactive: true
+        continueOnError: false
```
jongio, line 37: Why dup?

azure.yaml
jongio, line 24: I don't think we can guarantee the script_env venv.  Why not have the venv create in the .sh file?

azure.yaml
```diff
@@ -8,16 +8,30 @@ services:
     project: ./app/backend
     language: py
     host: appservice
+  scripts:
```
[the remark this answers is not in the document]
johanste: To install the requirements for the doc prep script into a virtual environment. From what I understand, `azd restore` will install each "service"'s requirements into its own virtual environment.

azure.yaml
johanste: Makes sense.

azure.yaml
johanste: Different shells. Is there a better way to do it?

azure.yaml
johanste: I assumed that we want to make use of `azd restore` - isn't that why it creates a virtual environment per "service" in the first place? And since it is `azd restore` that creates the virtual environment, I'd think that there is some way to plumb the environment name through here....

azure.yaml
jongio: The default section will default to appropriate shell for the detected OS.

azure.yaml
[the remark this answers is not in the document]
jongio: What happens when you call `azd deploy`?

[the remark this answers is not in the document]
johanste: Need to try. The docs are a bit unclear - the host is required, but empty string is an allowed value. The documentation says that it is optional and uses app service if omitted.

What is the intent? I don't want to base things on assumptions on current behavior. And is there a story for dev-only dependencies/local only tools?

azure.yaml
johanste: That does not seem to be the case:

```sh
$> azd deploy

ERROR: failed running pre hooks: failed running scripts for hooks 'azd deploy', hook configuration for 'predeploy' is invalid, unable to determine script type. Ensure 'Shell' parameter is set in configuration options
```

jongio: I was mistaken.

johanste:
Superseded by #15

PR #15, opened 2023-03-14
Install scripts into scripts/.venv virtual environment. Add sh shell …

johanste:
…support for azd hooks when running posix/non-windows OS

## Purpose

Avoid installing python packages globally + drops requirement to have PowerShell in non-windows machines.

```
[ ] Yes
[ ] No
```

## Pull Request Type

```
[ ] Bugfix
[X] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
```

## What to Check
* ...

### DIFF
--- a/azure.yaml
+++ b/azure.yaml
@@ -10,14 +10,27 @@ services:
     host: appservice
 hooks:
     postprovision:
-      shell: pwsh
-      run: ./scripts/prepdocs.ps1
-      interactive: true
-      continueOnError: false
+      windows:
+        shell: pwsh
+        run: ./scripts/prepdocs.ps1
+        interactive: true
+        continueOnError: false
+      posix:
+        shell: sh
+        run: ./scripts/postprovision.sh
+        interactive: true
+        continueOnError: false
     predeploy:
+      windows:
         shell: pwsh
         run:  cd ./app/frontend;npm install;npm run build
         interactive: true
         continueOnError: false
+      posix:
+        shell: sh
+        run:  cd ./app/frontend;npm install;npm run build
+        interactive: true
+        continueOnError: false
+
 
       
--- a/scripts/postprovision.sh
+++ b/scripts/postprovision.sh
@@ -1,0 +1,30 @@
+ #!/bin/sh
+
+if ! command -v python -v &> /dev/null
+then
+    echo 'Python not found on $PATH'
+    exit -1
+fi
+
+echo 'Creating python virtual environment "scripts/.venv"'
+python -m venv scripts/.venv
+
+
+echo 'Installing dependencies from "requirements.txt" into virtual environment'
+./scripts/.venv/bin/python -m pip install -r scripts/requirements.txt
+
+if [ "$AZURE_STORATE_ACCOUNT" == "" ]
+then
+    echo 'Environment variables not set - this script assumes that it runs as part of an azd postprovision hook.'
+    echo 'You can run the script manually by passing in the appropriate command line parameters after'
+    echo 'activting the scripts/.venv virtual environment.'
+    echo 'For more information about virtual environments, see https://docs.python.org/3/tutorial/venv.html'
+    echo
+    exit -1
+fi
+
+echo 'Running "prepdocs.py"'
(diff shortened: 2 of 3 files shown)


azure.yaml
```diff
@@ -10,14 +10,27 @@ services:
...
-      interactive: true
-      continueOnError: false
+      windows:
+        shell: pwsh
+        run: ./scripts/prepdocs.ps1
+        interactive: true
+        continueOnError: false
+      posix:
+        shell: sh
+        run: ./scripts/postprovision.sh
```
jongio, line 20: please name prepdocs.sh

scripts/postprovision.sh
```diff
@@ -0,0 +1,30 @@
...
+
+if [ "$AZURE_STORATE_ACCOUNT" == "" ]
+then
+    echo 'Environment variables not set - this script assumes that it runs as part of an azd postprovision hook.'
+    echo 'You can run the script manually by passing in the appropriate command line parameters after'
+    echo 'activting the scripts/.venv virtual environment.'
+    echo 'For more information about virtual environments, see https://docs.python.org/3/tutorial/venv.html'
+    echo
+    exit -1
+fi
```
jongio, line 24: I load them automatically so it can be run outside of a hook and not ask the user to load, see the ps1 version.

scripts/prepdocs.ps1
```diff
@@ -8,5 +8,6 @@ foreach ($line in $output) {
 
 Write-Host "Environment variables set."
 
-pip install -r ./scripts/requirements.txt
-python ./scripts/prepdocs.py '.\data\*' --storageaccount $env:AZURE_STORAGE_ACCOUNT --container $env:AZURE_STORAGE_CONTAINER --searchservice $env:AZURE_SEARCH_SERVICE --index $env:AZURE_SEARCH_INDEX -v
\ No newline at end of file
+python -m venv scripts/.venv
+scripts/scripts/.venv/scripts/python pip install -r ./scripts/requirements.txt
```
jongio, line 12: why scripts/scripts and not /scripts

azure.yaml
johanste: I'd like to avoid using the same name as the python script. This script only works in the context of a hook or in a shell where environment variables have been set from `azd env get-values`. It also does more/different things than the prepdocs.py script does (e.g. create the virtual environment)

Is there a specific reason that it should have the same name?

scripts/prepdocs.ps1
johanste: Typo. Fixed.

scripts/postprovision.sh
johanste: We also pip install dependencies in the script ps1 version. It's a bit of Russian dolls at the moment with cli calling a hook script calling a init script - not to mention and the hook script has dual purpose (restore execution environment + execute init script). This is why I tried to use `azd restore` in #11.  

I think it would be better to load environment variables from `azd` in the `prepdocs.py` script than wrapping `prepdocs.py` script in another script just to set environment variables. But that is a bigger change. I intended to follow up with a clean-up of the `prepdocs.py` script (I ran across a few bugs in there while doing this work), but I wanted to stop the pollution of the global environment first.

azure.yaml
jongio: Because it does the same thing, just a diff format.  This file won't always be run post-provision.  It is there to prep the docs.

scripts/postprovision.sh
jongio: I would rather take the time now to do the best option, even if a bigger change.

azure.yaml
johanste: The primary purpose of `postprovision.sh` is to install the requirements for the `prepdocs.py` script/CLI. I expect users to run `prepdocs.py` directly to prep docs for subsequent runs. This is what allows them to specify which documents to index without copying them into the data folder etc.

jongio:
We decided to cover this in this PR: https://github.com/Azure-Samples/azure-search-openai-demo/pull/16

PR #16, opened 2023-03-14
Codespace/Remote-Container Support

jongio:
This PR adds codespace and remote-container support.


app/start.sh
```diff
@@ -0,0 +1,60 @@
...
+    echo "Failed to build frontend"
+    exit $?
+fi
+
+echo ""
+echo "Starting backend"
+echo ""
+
+cd ../backend
+xdg-open http://127.0.0.1:5000
```
pablocastro, line 55: Does this work well inside a codespace? What does it do since there isn't anything to open in the local console?

pablocastro (commented):
One more note: to run on linux/mac, prepdocs.ps1 also needs to be updated to use / instead of \ in paths.

jongio:
> 

@johanste is created a .sh file for linux, we'll update to that.

.vscode/tasks.json
```diff
@@ -12,12 +12,15 @@
         {
             "label": "Start App (Script)",
             "type": "shell",
-            "command": "start.cmd",
+            "command": "${workspaceFolder}/app/start.sh",
+            "windows": {
+                "command": "start.cmd"
```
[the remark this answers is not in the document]
jongio: Changed...

app/start.sh
jongio: This experience for me is that it doesn't error in Codespaces, but pops a dialog with a link to click to open the site.

app/start.sh
```diff
@@ -0,0 +1,60 @@
...
+    echo "Failed to load environment variables from azd environment"
+    exit $?
+fi
+
+echo ""
+echo "Restoring backend python packages"
+echo ""
+
+cd backend
+pip install -r requirements.txt
```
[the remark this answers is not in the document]
jongio: @johanste is checking on best approach here...

app/start.sh
jongio: Resolving as it doesn't error for me.

app/backend/requirements.txt
```diff
@@ -1,4 +1,4 @@
-azure-identity==1.12.0
+azure-identity==1.13.0b3
```
johanste, line 1: Why do we need a dependency of a preview version of `azure-identity`?

app/start.sh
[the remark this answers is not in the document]
johanste: At the very least, we need to run pip from the virtual environment. Creating the virtual environment (on line 19) doesn't activate it.

app/backend/requirements.txt
jongio: Because it has AzureDeveloperCredential, which is needed for this to work locally and remotely.

app/start.sh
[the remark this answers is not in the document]
jongio: Done

app/start.ps1
```diff
@@ -0,0 +1,60 @@
...
+
+
+Write-Host 'Creating python virtual environment "backend/backend_env"'
+python -m venv backend/backend_env
+
+Write-Host ""
+Write-Host "Restoring backend python packages"
+Write-Host ""
+Set-Location backend
+./backend_env/scripts/pip3 install -r requirements.txt
```
johanste, line 26: This is only going to work on windows. Which is fine for the hooks. Not sure how much we care about non-windows users running the PowerShell scripts outside the hook, but if we do, we also have to check the `backend_env/bin` folder...

johanste: We should be consistent in using `python -m pip` (from the virtual environment) to install packages

johanste (approved):
A couple of minor comments. The change to `python -m pip` is probably the most consequential

PR #25, opened 2023-03-17
Existing resources, tenant id, codespaces/linux support

jongio:
This PR will:

- Allow the use of existing search, openai, and storage resources
- Add codespaces and linux support
- Use python venvs for backend and scripts
- Fix multi-tenant issues


infra/main.bicep
```diff
@@ -32,16 +44,28 @@ var resourceToken = toLower(uniqueString(subscription().id, environmentName, loc
...
 
 // Organize resources in a resource group
-resource rg 'Microsoft.Resources/resourceGroups@2021-04-01' = {
+resource resourceGroup 'Microsoft.Resources/resourceGroups@2021-04-01' = {
   name: !empty(resourceGroupName) ? resourceGroupName : '${abbrs.resourcesResourceGroups}${environmentName}'
   location: location
   tags: tags
 }
 
+resource openAiResourceGroup 'Microsoft.Resources/resourceGroups@2021-04-01' existing = if (!empty(openAiResourceGroupName)) {
```
briansmi, line 53: Just reading the code, are we moving to a model where there is a RG per resource type or will they all just be mapped to the AZURE_RESOURCE_GROUP by default? I assume this explicit mapping is needed for the scenario where users are bringing their own resources that might not be in the same group?

jongio: We default to a single resource group, but allow the user to override with an existing resource group.  

The scenario we don't support is creating the solution from scratch using more than one resource group.  The "other" resource groups need to already exist.

jongio, line 53: @briansmi - Do you think it is common for folks to have existing openai, search, or storage that aren't in the same locations?  Right now, the template is easy if all are in the same location, but more difficult if they are in different locations.  There would be more work required to support resources in different locations than the location the application is deployed to.

Is that common enough scenario that we want to make it very easy to support that?

app/start.sh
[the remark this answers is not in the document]
jongio: To include sh or just ps1 depends on target audience, some posix people will never install ps1.  The pre-req is there because that was the only option, which it can be updated to not require it for linux folks.

.vscode/tasks.json
```diff
@@ -12,12 +12,15 @@
...
+            "command": "${workspaceFolder}/app/start.sh",
+            "windows": {
+                "command": "pwsh ${workspaceFolder}/app/start.ps1"
+              },
             "presentation": {
                 "reveal": "silent"
             },
             "options": {
-                "cwd": "${workspaceFolder}/app"
+                "cwd": "${workspaceFolder}/app",
```
briansmi, line 23: extra comma?

infra/main.bicep
jongio: Created issue #33 to track

app/start.sh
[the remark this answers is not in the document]
jongio: For now, I'm updating the README to say PowerShell is optional.

.vscode/tasks.json
jongio: Fixed

app/start.sh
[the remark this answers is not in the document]
jongio: Done https://github.com/Azure-Samples/azure-search-openai-demo/pull/25/commits/91eb8fe75f95931ac08f3ec00e2d9ad0fcc81fa1

pablocastro (approved):

PR #29, opened 2023-03-18
Avoid padding error when using an encoded Azure Storage Services key

swarchitex:
## Purpose
Corrects an issue where the environment variable preprocessing may remove the trailing "==" of an encoded key (which was obtained from the Azure Storage Service, in encoded format). This happens when the **AZURE_STORAGE_KEY** environment variable is set on **.azure/**<_environment name_>**/.env**.

The change ensures that the storage key is properly terminated and avoids an **Incorrect padding** error on **binascii.a2b_base64()** while trying to create the credentials to connect to the service. If and when this happens the upload fails and the whole script is aborted; nothing is uploaded.

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[X] Bugfix
[X] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

Prep work:
Execute azd provision in order to create the infrastructure.
Once the services have been created, obtain the Cognitive Search Service key and Storage Service key. Use the main admin keys to be certain.
Add the environment variables AZURE_STORAGE_KEY and AZURE_SEARCH_KEY to the **.azure/**<_environment name_>**/.env** file, and populate with the values obtained. Preferrably use quotes for the values.

Test:
```
execute **scripts/prepdocs.ps1** on the PowerShell environment -or- **azd up**
``` 

## What to Check
Blobs have been uploaded to **container** on Storage Services, and an index has been created in Search Services.

### DIFF
--- a/scripts/prepdocs.ps1
+++ b/scripts/prepdocs.ps1
@@ -8,5 +8,6 @@ foreach ($line in $output) {
 
 Write-Host "Environment variables set."
 
-pip install -r ./scripts/requirements.txt
+pip install -r ./scripts/requirements.txt --quiet
+# Add if having Azure authentication issues: --searchkey $env:AZURE_SEARCH_KEY  --storagekey $env:AZURE_STORAGE_KEY
 python ./scripts/prepdocs.py './data/*' --storageaccount $env:AZURE_STORAGE_ACCOUNT --container $env:AZURE_STORAGE_CONTAINER --searchservice $env:AZURE_SEARCH_SERVICE --index $env:AZURE_SEARCH_INDEX -v
--- a/scripts/prepdocs.py
+++ b/scripts/prepdocs.py
@@ -36,9 +36,9 @@ args = parser.parse_args()
 
 # Use the current user identity to connect to Azure services unless a key is explicitly set for any of them
 default_creds = DefaultAzureCredential() if args.searchkey == None or args.storagekey == None else None
-search_creds = default_creds if args.searchkey == None else AzureKeyCredential(args.searchkey)
+search_creds = default_creds if args.searchkey == None else AzureKeyCredential(args.searchkey.strip())
 if not args.skipblobs:
-    storage_creds = default_creds if args.storagekey == None else args.storagekey
+    storage_creds = default_creds if args.storagekey == None else args.storagekey.strip().strip("=")+"=="
 
 def blob_name_from_file_page(filename, page):
     return os.path.splitext(os.path.basename(filename))[0] + f"-{page}" + ".pdf"


swarchitex:
> 

vhvb1989:
@swarchitex 

> This happens when the AZURE_STORAGE_KEY environment variable is set on .azure/<environment name>/.env.

How are you setting `AZURE_STORAGE_KEY`?  Are you calling `azd env set AZURE_STORAGE_KEY` ?  Or are you setting `AZURE_STORAGE_KEY` as a bicep output?

I'd like to understand where is the value losing the padding and fix it, instead of manually inserting padding.

swarchitex:
This is a verbose answer. Sorry about that.

**Background**
I had to resort to using they keys because something on my environment is caching more than one AAD identity and throws an error something about being unable to identify the credentials to use, due to the fact that there is more than one cached (and it states so on the error message when the .ps1 script tries to ascertain the identity to use). This _might_ be due to the fact that I have setup (not joined) a different AAD tenant on my Windows installation (my production/customer's) vs the one I am logging in (my dev) on **azd login**.

**Prep work**
The storage key is obtained after the script finishes provisioning and then fails due to issue just described. This storage key being [64]bit encoded should have an "==" terminator and has a fixed length of [80?] chars. 

**Attempted resolution**
Once the script fails for the first time, then the keys are obtained from the Azure portal and then added to the **\<environment name\>\\.env** file as follows. I am pasting the keys exactly as on the script. They are not valid anymore.

**AZURE_STORAGE_KEY="xvbh4OaJFIwSj9svOCqgOmio04pnWTXe6yAN/LsckDDkhPbR5DYYadU8ejtBiSkTjg4fsvgQqgTw+AStCG0PIw=="
AZURE_SEARCH_KEY="VIDFyypwdyTTqx6jrcl01oW2V70TYMPecAn3D7LuwPAzSeDddstP"**

Once the keys are on the env file, the **prepdocs.ps1** is updated as follows:

python ./scripts/prepdocs.py './data/*' **--searchkey $env:AZURE_SEARCH_KEY  --storagekey $env:AZURE_STORAGE_KEY** --storageaccount $env:AZURE_STORAGE_ACCOUNT ........

**Issue**
However, this fails with a "_padding error_" while decoding the storage key, just before the script starts uploading the split files.

The problem lies on lines 3-7 of **prepdocs.ps1**. That loop populates the PS environment variables array with the parameters passed to the script. In order to do this, it uses the **=** character to detect the variables (**$line.Split("=")**). That is where the trailing **=** characters are lost in the storage key environment variable and therefore breaks the decoding code.

**Solution**
My proposed solution is to, in case the loop above is changed in the future, strip any left-over **=** characters and then adding the two **=** terminator characters required for the string to be properly formatted.

vhvb1989:
@swarchitex, instead of settings the keys within the `.env` file (the azd-environment), can you set them as system environment variables and then read them from `prepdocs.ps` before calling `prepdocs.py` ?   the values are only used by the `prepdocs.py` script, so, you don't really need them inside the azd-environment.

scripts/prepdocs.ps1
```diff
@@ -8,5 +8,6 @@ foreach ($line in $output) {
 
 Write-Host "Environment variables set."
 
-pip install -r ./scripts/requirements.txt
+pip install -r ./scripts/requirements.txt --quiet
+# Add if having Azure authentication issues: --searchkey $env:AZURE_SEARCH_KEY  --storagekey $env:AZURE_STORAGE_KEY
```
vhvb1989, line 12: add comments about setting `AZURE_STORAGE_KEY` and `AZURE_SEARCH_KEY` as system environment variables.  Instead of adding them to `.env` file.  That would fix the issue on previous lines where the padding is lost, and you don't need to add the padding back.

PR #37, opened 2023-03-21
Use Azure Form Recognizer as document preprocessing to extract text, tables, and document layout

anatolip:
## Purpose

Use Azure Form Recognizer as document preprocessing to extract text, tables, and document layout.

List of changes:
- Add Azure Form Recognizer as a primary way to extract information from the documents in prepdocs.py.
- Add support for tables in preprocessing to provide HTML tables as input for ChatGPT.
- Moved existing PyPdf local PDF parser under --localpdfparser flag since it only support text extraction from digital PDFs.
- Add Azure Form Recognizer into deployment configs.
- Modify prompt to return HTML tables if possible. Same question: What is cost difference between plans?
- Added new section into Benefit_Options.pdf sample (see page 4) with image of the table for plans cost comparison to illustrate extraction text from image and extraction of the tables by Azure Form Recognizer.

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

## How to Test
```
Re-run deployment to create and configure Azure Form Recognizer resource.
Re-run document pre-processing script to use Azure Form Recognizer to extract text, tables and document layout.

Suggested question:
What is cost difference between plans?
It should return Table from Benefit_Options.pdf (page 4) in HTML format.
```


scripts/prepdocs.py
```diff
@@ -74,18 +96,74 @@ def remove_blobs(filename):
...
+    html = "<table>"
+    rows = [sorted([cell for cell in table.cells if cell.row_index == i], key=lambda cell: cell.column_index) for i in range(table.row_count)]
+    for row_cells in rows:
+        html += "<tr>"
+        for cell in row_cells:
+            tag = "th" if (cell.kind == "columnHeader" or cell.kind == "rowHeader") else "td"
+            cell_spans = ""
+            if cell.column_span > 1: cell_spans += f" colSpan={cell.column_span}"
+            if cell.row_span > 1: cell_spans += f" rowSpan={cell.row_span}"
+            html += f"<{tag}{cell_spans}>{cell.content}</{tag}>"
```
pablocastro, line 109: Is cell.content already HTML-escaped? Do we want it to be? On one hand there might be valid cases where GPT can generate additional markup within cells, on the other hand unescaped strings could have "<" ">" or other characters that will make it invalid HTML. The FE already sanitizes it for security, but there's still the issue of breaking the format. A simple solution would be to just do html.escape() on this, a more complicated one would be to see what are well-formed tags and what's just accidental characters that need to be escaped. I would propose that we just do html.escape() for now.

scripts/prepdocs.py
```diff
@@ -74,18 +96,74 @@ def remove_blobs(filename):
...
+        html +="</tr>"
+    html += "</table>"
+    return html
+
+def get_document_text(filename):
+    offset = 0
+    page_map = []
+    if args.localpdfparser:
+        reader = PdfReader(filename)
+        pages = reader.pages
```
pablocastro, line 119: A bit unfortunate that we're opening/parsing each PDF twice (here and in upload_blobs). No the end of the world if it'll keep the code simple since this is an example.

scripts/prepdocs.py
```diff
@@ -41,24 +45,42 @@
...
+        for i in range(len(pages)):
+            blob_name = blob_name_from_file_page(filename, i)
+            if args.verbose: print(f"\tUploading blob for page {i} -> {blob_name}")
+            f = io.BytesIO()
+            writer = PdfWriter()
+            writer.add_page(pages[i])
+            writer.write(f)
+            f.seek(0)
+            blob_container.upload_blob(blob_name, f, overwrite=True)
+    else:
```
pablocastro, line 80: We might want to not upload blobs in the non-PDF case (and maybe print a warning) if Form Recognizer is not enabled, since we won't be able to parse these.

README.md
pablocastro, line 1: When using Form Recognizer more file types will "just work" right? Could we list some of the common types that would now work? We also may need to clarify that for non-PDFs we'd still not be splitting the source docs into individual pages, so citations won't be ideal if the documents are more of a page or two long without an improvement in that area.

chuwik (commented):
UX changes look good. Deferring to Pablo to review & approve backend changes.

README.md
```diff
@@ -22,7 +22,7 @@ The repo includes sample data so it's ready to try end to end. In this sample ap
...
 
-> **AZURE RESOURCE COSTS** by default this sample will create Azure App Service and Azure Cognitive Search resources that have a monthly cost. You can switch them to free versions of each of them if you want to avoid this cost by changing the parameters file under the infra folder (though there are some limits to consider; for example, you can have up to 1 free Cognitive Search resource per subscription.)
+> **AZURE RESOURCE COSTS** by default this sample will create Azure App Service and Azure Cognitive Search resources that have a monthly cost, as well as Form Recognizer resource that has cost per document page. You can switch them to free versions of each of them if you want to avoid this cost by changing the parameters file under the infra folder (though there are some limits to consider; for example, you can have up to 1 free Cognitive Search resource per subscription, and free Form Recognizer resouse only analyze first 2 pages of the document.)
```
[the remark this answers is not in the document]
anatolip: Fixed. Thanks.

app/backend/approaches/chatreadretrieveread.py
```diff
@@ -10,7 +10,8 @@
 class ChatReadRetrieveReadApproach(Approach):
     prompt_prefix = """<|im_start|>system
 Assistant helps the company employees with their healthcare plan questions, and questions about the employee handbook. Be brief in your answers.
-Answer ONLY with the facts listed in the list of sources below. If there isn't enough information below, say you don't know. Do not generate answers that don't use the sources below. If asking a clarifying question to the user would help, ask the question. 
+Answer ONLY with the facts listed in the list of sources below. If there isn't enough information below, say you don't know. Do not generate answers that don't use the sources below. If asking a clarifying question to the user would help, ask the question.
+For tabular information return it as html table. Do not return markdown format.
```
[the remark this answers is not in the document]
anatolip: Fixed. Thanks.

app/backend/approaches/readretrieveread.py
```diff
@@ -21,6 +21,7 @@ class ReadRetrieveReadApproach(Approach):
     template_prefix = \
 "You are an intelligent assistant helping Contoso Inc employees with their healthcare plan questions and employee handbook questions. " \
 "Answer the question using only the data provided in the information sources below. " \
+"For tabular information return it as html table. Do not return markdown format. " \
```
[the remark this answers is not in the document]
anatolip: Fixed. Thanks.

app/backend/approaches/retrievethenread.py
```diff
@@ -13,6 +13,7 @@ class RetrieveThenReadApproach(Approach):
 "You are an intelligent assistant helping Contoso Inc employees with their healthcare plan questions and employee handbook questions. " + \
 "Use 'you' to refer to the individual asking the questions even if they ask with 'I'. " + \
 "Answer the following question using only the data provided in the sources below. " + \
+"For tabular information return it as html table. Do not return markdown format. "  + \
```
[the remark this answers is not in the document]
anatolip: Fixed. Thanks.

scripts/prepdocs.py
anatolip: Yes, I've seen this issue but I've left it as is for now to make code simpler.

scripts/prepdocs.py
anatolip: Yes, I've seen this issue but I've left it as is for now to make code simpler.

README.md
anatolip: Currently images (i.e. jpg/png) would work fine in preprocessing and frontend but yes tiffs will not be split into pages and they also will not be rendered in the frontend. If there are enough demand for extending this demo into image formats, we can have a separate commit for it.

pablocastro (approved):

infra/main.bicep
```diff
@@ -274,6 +308,9 @@ output AZURE_OPENAI_RESOURCE_GROUP string = openAiResourceGroup.name
 output AZURE_OPENAI_GPT_DEPLOYMENT string = gptDeploymentName
 output AZURE_OPENAI_CHATGPT_DEPLOYMENT string = chatGptDeploymentName
 
+output AZURE_FORMRECOGNIZER_SERVICE string = formrecognizer.outputs.name
```
jongio, line 311: Please keep consistent case for formRecognizer, camel

infra/core/ai/formrecognizer.bicep
```diff
@@ -0,0 +1,26 @@
...
+  properties: {
+    customSubDomainName: customSubDomainName
+    publicNetworkAccess: publicNetworkAccess
+  }
+  sku: sku
+}
+
+output endpoint string = account.properties.endpoint
+output id string = account.id
+output name string = account.name
```
jongio, line 26: Instead of this, you can just reference the cognitiveservices.bicep module and pass it kind:FormRecognizer paramter in main.bicep.

If you want an explicit formrecognizer.bicep file, then you can also do the same thing.  That way we are reusing the cognitiveservices.bicep file and can have consistency there.

infra/main.bicep
anatolip: Thanks for feedback. Addressed in PR: https://github.com/Azure-Samples/azure-search-openai-demo/pull/49

infra/core/ai/formrecognizer.bicep
anatolip: Thanks for feedback. Addressed in PR: https://github.com/Azure-Samples/azure-search-openai-demo/pull/49

PR #56, opened 2023-03-27
Add Markdown Render Support to GPT completions

### LINKED ISSUE #30: Markdown is not rendered as HTML in the frontend
[](url)
Screenshot 2023-03-19 at 4 21 11 PM

praveen-palanisamy:
## Purpose

The frontend on the `main` branch currently doesn't support rendering Markdown in the model's responses/completions. This PR adds support for rendering Markdown while retaining support for raw HTML code in the completions.
Fixes #30 
|Current|This PR|
|--------|---------|
|without-markdown-support |markdown-support_cropped|

```
[ ] Yes
[x] No
```
## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone https://github.com/Azure-Samples/azure-search-openai-demo
cd azure-search-openai-demo
git checkout -b praveen-palanisamy-add-markdown-render-support main
git pull https://github.com/praveen-palanisamy/azure-search-openai-demo add-markdown-render-support
```

* Test the code
Run backend & frontend on Azure or locally and test with a prompt that would result in GPT's response/completion containing Markdown syntax. Example prompt:

```
Showcase markdown syntax features
```

## What to Check
Verify that the UI renders Markdown syntax as expected.

## Other Information

* Uses `react-markdown` which is safe by default (no `dangerouslySetInnerHTML` or XSS attacks)


kishorerv93 (approved):

kishorerv93:
@praveen-palanisamy  approved this pull request. It is working as expected.

ToadRider:
bumping this so that  @jongio / @chuwik / @pablocastro  can review this to get it implemented.

app/frontend/src/components/Answer/Answer.tsx
```diff
@@ -58,7 +60,9 @@ export const Answer = ({
             </Stack.Item>
 
             <Stack.Item grow>
-                <div className={styles.answerText} dangerouslySetInnerHTML={{ __html: sanitizedAnswerHtml }}></div>
+                <div className={styles.answerText}>
+                    <ReactMarkdown children={sanitizedAnswerHtml} rehypePlugins={[rehypeRaw]} />
```
pamelafox, line 64: @praveen-palanisamy I'm curious why you brought in the rehypePlugins? I notice a similar repo brings in two other plugins instead, remarkGfm and supersub:
https://github.com/microsoft/sample-app-aoai-chatGPT/blob/354b62b3485ce8270798dd625c4ce904bd311695/frontend/src/components/Answer/Answer.tsx#L63

I'm trying to determine what plugins we need for this repo. Thanks!

pamelafox:
I suspect that the actual Markdown returned by the second question included backticks, and thats why the Markdown renders as code. Here's a deployed version which includes the markdown renderer:

https://app-backend-dsqwnx2uoc57g.azurewebsites.net/

app/frontend/src/components/Answer/Answer.tsx
praveen-palanisamy: Thanks for looking into this PR. Yes, preserving the citations and rendering them appropriately is one of the main reasons for choosing this `rehype` plugin.

praveen-palanisamy:
@pamelafox : Resolved the merge conflicts with the latest `main` branch. Please let me know if there's anything pending on this PR or if it's ready to finalize/merge.

pamelafox:
@praveen-palanisamy Thanks for updating to main! I've integrated this change in this demo app here to test it out:
https://app-backend-dsqwnx2uoc57g.azurewebsites.net/

Now that we've added streaming to the repo, this change does introduce a slight visual hiccup when it's interpreting certain incremental Markdown, and some of those can be jarring. I screen capped it happening:

Glitch - turned list item into a heading:
Screenshot 2023-11-01 at 1 44 53 PM

Fixed a moment after - back into a normal list item:
Screenshot 2023-11-01 at 1 45 11 PM

I haven't dug into precisely the underlying Markdown causing the glitch, but I think we'd want some solution before merging into main, to avoid that jarring effect. It looks like it may be related to citations, and we might need to make sure that citations get streamed a whole citation at a time, but I'm not certain that's the case.

If you have time to look into it, let us know what you find out.

pamelafox:
cc @SteveJSteiner who authored the streaming change (see comment above on interaction with this PR)

praveen-palanisamy:
That's a good observation! If we have Markdown rendering enabled while streaming (partial fragments/chunks of) output text, that's an inevitable artifact, I think. I have observed this behavior with the official ChatGPT webapp and also with the Bing (Enterprise) Chat. A sample repro below where the partial markdown bold text takes a few more bytes to be complete and render properly. This output didn't include Markdown headers to match with your above screenshot, but you can see the behavior:

https://github.com/Azure-Samples/azure-search-openai-demo/assets/4770482/7b2d60b2-ebd5-422b-8ae0-1789afcd825a

If this is an unacceptable UX, one way to handle this is to add another post-processing step in `AnswerParser.tsx` to the `run_with_streaming` output fragments (from `chat.completion.chunk`) to contain complete/closed Markdown tags before displaying on the UI.

SteveJSteiner:
Yes, I have seen chatGPT with some visual ticks due to re-rendering of markdown. Table as the output hits this.  I can also reproduce some oddities in Bing as well as it creates annotated lists.

@pamelafox - what kind of bar do you wish to apply for these cases?  Adding markdown support seems highly worthwhile.
Perhaps fix the 'worst' of the visual hiccups as one-offs?

Can the markdown support be disabled if a user customizing the demo repo finds it too jarring for their end customers?

praveen-palanisamy:
@pamelafox : Could you please provide your final comments to finalize this PR soon?

pamelafox:
Okay, so I think what we could do is just change our CSS so that headings inside the chat boxes aren't actually very large, that way it isn't such a jarring visual glitch. How about that?

mrmueller:
Wanted to add the feedback that markdown support is heavily requested by our users. We have many tables in our output.
Some UI glitches during streamig are very acceptable for us, while not properly formatted table are not.

alexmedranosan:
Hi @pamelafox . Thanks for the great work guys! Do you know why this PR was not closed?

pamelafox:
I've added remark-gfm, which will render the tables, since that seems to be a desire of the folks following this PR.

I've also changed our MarkdownViewer to use the same package (react-markdown), for consistency. The markdown viewer originally used react-markdown in the PR, but I asked them to change to marked for package size reasons. However, the react-markdown package size seems to have decreased, and its snyk score increased, and it has more functionality, so it seems okay to use it in both places.

Some screenshots:

Screenshot 2024-08-19 at 10 37 20 AM

Screenshot 2024-08-19 at 10 36 26 AM

pamelafox:
I also updated the system prompts which were previously discouraging markdown. Now they say nothing, with the assumption that LLMs tend toward markdown.

pamelafox (approved):

app/backend/approaches/chatreadretrieveread.py
```diff
@@ -57,7 +57,7 @@ def __init__(
     def system_message_chat_conversation(self):
         return """Assistant helps the company employees with their healthcare plan questions, and questions about the employee handbook. Be brief in your answers.
         Answer ONLY with the facts listed in the list of sources below. If there isn't enough information below, say you don't know. Do not generate answers that don't use the sources below. If asking a clarifying question to the user would help, ask the question.
-        For tabular information return it as an html table. Do not return markdown format. If the question is not in English, answer in the language used in the question.
+        If the question is not in English, answer in the language used in the question.
```
mattgotteiner, line 60: Did we need to explicitly ask for markdown in the prompt?

pamelafox: In my experience, LLMs tend to produce Markdown, but likely would produce even more if we explicitly asked for it. Perhaps I should add "If your answer includes lists, headings, tables, or emphasis, use Markdown syntax."

mattgotteiner (approved):
This is a surprisingly small change, especially when you don't consider the test regenerations. Great!

pamelafox:
@mattgotteiner For the record, I did just run evaluations on the prompt change, and there are no change in metrics. Of course, we don't have a "how markdowny is this?" metric currently, but answer quality seems unaffected.

Screenshot 2024-08-21 at 2 48 57 PM

pamelafox:
Thanks so much @praveen-palanisamy ! Sorry it took more than a year to merge.

PR #58, opened 2023-03-27
[README] Add multi-tenant, use-device-code for azd login (optional)

jimmyliao:
## Purpose

* Add multi-tenant, use-device-code for azd login in README.md.

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
```

## What to Check
* ...

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ You can run this repo virtually by using GitHub Codespaces or VS Code Remote Con
 1. Create a new folder and switch to it in the terminal
 1. Run `azd login` 
    1. If you're in multi-tenant organization, please use `azd login --tenant-id {YOUR_TENANT_ID}`
-   1. If you're using dev-container, please use `azd login --use-device-code`
+   1. If you're using VS Code DevContainer or GitHub Codespaces, please use `azd login --tenant-id {YOUR_TENANT_ID} --use-device-code`
 1. Run `azd init -t azure-search-openai-demo`
     * For the target location, the regions that currently support the models used in this sample are **East US** or **South Central US**. For an up-to-date list of regions and models, check [here](https://learn.microsoft.com/en-us/azure/cognitive-services/openai/concepts/models)
 
--- a/README.md
+++ b/README.md
@@ -50,7 +50,9 @@ You can run this repo virtually by using GitHub Codespaces or VS Code Remote Con
 #### Project Initialization
 
 1. Create a new folder and switch to it in the terminal
-1. Run `azd login`
+1. Run `azd login` 
+   1. If you're in multi-tenant organization, please use `azd login --tenant-id {YOUR_TENANT_ID}`
+   1. If you're using dev-container, please use `azd login --use-device-code`
 1. Run `azd init -t azure-search-openai-demo`
     * For the target location, the regions that currently support the models used in this sample are **East US** or **South Central US**. For an up-to-date list of regions and models, check [here](https://learn.microsoft.com/en-us/azure/cognitive-services/openai/concepts/models)
 
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ You can run this repo virtually by using GitHub Codespaces or VS Code Remote Con
 1. Create a new folder and switch to it in the terminal
 1. Run `azd login` 
    1. If you're in multi-tenant organization, please use `azd login --tenant-id {YOUR_TENANT_ID}`
-   1. If you're using VS Code DevContainer or GitHub Codespaces, please use `azd login --tenant-id {YOUR_TENANT_ID} --use-device-code`
+   1. If you're using the VS Code DevContainer or GitHub Codespaces, please use `azd login --tenant-id {YOUR_TENANT_ID} --use-device-code`
 1. Run `azd init -t azure-search-openai-demo`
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -50,7 +50,9 @@ You can run this repo virtually by using GitHub Codespaces or VS Code Remote Con
 #### Project Initialization
 
 1. Create a new folder and switch to it in the terminal
-1. Run `azd login`
+1. Run `azd login` 
+   1. If you're in multi-tenant organization, please use `azd login --tenant-id {YOUR_TENANT_ID}`
+   1. If you're using dev-container, please use `azd login --use-device-code`
```
jongio, line 55:
```suggestion
   1. If you're using VS Code DevContainer or GitHub Codespaces, please use `azd login --use-device-code`
```

jimmyliao: I add `azd login --tenant-id {YOUR_TENANT_ID} --use-device-code` initially, will combine with this.

jimmyliao: Refined. @jongio , @vhvb1989

README.md
```diff
@@ -52,7 +52,7 @@ You can run this repo virtually by using GitHub Codespaces or VS Code Remote Con
 1. Create a new folder and switch to it in the terminal
 1. Run `azd login` 
    1. If you're in multi-tenant organization, please use `azd login --tenant-id {YOUR_TENANT_ID}`
-   1. If you're using dev-container, please use `azd login --use-device-code`
+   1. If you're using VS Code DevContainer or GitHub Codespaces, please use `azd login --tenant-id {YOUR_TENANT_ID} --use-device-code`
```
jongio, line 55: If you're using the VS Code DevContainer or GitHub Codespaces, please use `azd login --tenant-id {YOUR_TENANT_ID} --use-device-code`

jimmyliao: > If you're using the VS Code DevContainer or GitHub Codespaces, please use `azd login --tenant-id {YOUR_TENANT_ID} --use-device-code`

refined. #fd4cb76

jimmyliao: @jongio I'd done for the refinement. Please have time review. Thanks.

jimmyliao: Hi @jongio , any required needed from requester side?

jongio:
Deferring to @vhvb1989

README.md
```diff
@@ -50,7 +50,9 @@ You can run this repo virtually by using GitHub Codespaces or VS Code Remote Con
 #### Project Initialization
 
 1. Create a new folder and switch to it in the terminal
-1. Run `azd login`
+1. Run `azd login` 
+   1. If you're in multi-tenant organization, please use `azd login --tenant-id {YOUR_TENANT_ID}`
+   1. If you're using the VS Code DevContainer or GitHub Codespaces, please use `azd login --tenant-id {YOUR_TENANT_ID} --use-device-code`
```
vhvb1989, line 55: well, it is not mandatory to use `--tenant-id XXXX` on multi-tenant scenarios. 

I'm guessing you ran into some situation where azd was not listing the subscription you needed, and instead, it was listing subscription from other(s) tenants.

By default, azd knows how to handle multi-tenant scenarios. However, each tenant might be requiring additional auth-steps, like a 2FA, for granting the access. AFAIK, 2FA can't be shared across tenants, which means that if you authenticate and provide 2FA for tenant-A, it will not be enough for accessing tenant B, if tenant B requires 2FA as well.

So, when you run `azd auth login` w/o a tenant flag, azd will pick the tenant based on the information provided by the Web-Browser. If the browser has an active session with tenant A, for your account, but an expired token for tenant B (which requires 2FA to refresh), azd will not take you to resolve 2FA for tenant B.  If there isn't any active session for your account, you would see the full login-process with your home-tenant (where your account belongs).

In case of the flag `--use-device-code`, there are also other implications, as your tenant might have specific policies for authenticating devices. Some tenants, for example, do not allow Linux devices, or require devices to be enrolled with the company.

So, I don't think you should ask customers to use these `flags` by default.  I see this as some sort of `troubleshooting` approach, where using `--tenant-id` flag is a way how you would force azd to login to a tenant. Like, in my previous example, by running `azd auth login --tenant-id tenantB`,  you would see the 2FA challenge from tenantB, which would update your browser session.  After that, you could run again `azd auth login` (w/o tenantId) and azd would manage multiple tenants just fine, getting the sessions from the browser.

I will create an issue for azd to provide a full guide on login and better troubleshooting.  Then, I would recommend referring to that documentation instead of asking folks to use the flags by default.

vhvb1989: https://github.com/Azure/azure-dev/issues/3166

vhvb1989: Added my comments @jongio .  LMK what you think

jongio: Yeah, only ask them to give the tenant if using multi-tenant.  I think the doc here:
image


Should read:

If using multi-tenant and codespaces, then supply the tenant id.  If that was the intention.

pamelafox: Should we close this PR?

pamelafox: We're going to close this PR since we haven't heard back. We are going to be adding some stuff about "--tenant-id" to the login doc specifically since you can have a multi-tenant situation there.

PR #113, opened 2023-04-13
Add speech recognizer and synthesis on browser interface

sowu880:
## Purpose

Enable speech input and output for browser interface.

```
[ ] Yes
[x ] No
```

## Pull Request Type

```
[ ] Bugfix
[x ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```


unmuntean:
tried to implement this, i get blank screen with the error "Uncaught TypeError: HF is not a constructor
    <anonymous> QuestionInput.tsx:16 
" . The lines with the problem are :
const SpeechRecognition = (window as any).speechRecognition|| (window as any).webkitSpeechRecognition;
const recognition = new SpeechRecognition();

kishorerv93:
I get `500 Internal Server Error` when tried to use this. 

Do we just need Speech Region and Speech Key?

sowu880:
> tried to implement this, i get blank screen with the error "Uncaught TypeError: HF is not a constructor QuestionInput.tsx:16 " . The lines with the problem are : const SpeechRecognition = (window as any).speechRecognition|| (window as any).webkitSpeechRecognition; const recognition = new SpeechRecognition();
> 
> EDIT: Mozilla and other browsers usually don't support webkit Speech, had to overwrite default browser settings.

Fix the bug: Add try catch for speech recognition constructor. Web Speech API is supported for the following browsers. Recognition can not be used on Mozilla and other browsers but will not throw exception.
image

sowu880:
Hi, could you help review the PR？ Thanks a lot.

kishorerv93:
@sowu880 

This code and integration of the speech is increasing the time of processing the request. A simple change should be let the text generate and display the result and let speech complete in the background without causing the further delay.

sowu880:
> integration of the speech is increasing the time of processing the request

Updated. For now, text will display without waiting speech generation.

pamelafox:
@sowu880 It seems that this PR doesn't include the creation of the speech resource. Can that be included as an optional resource in the Bicep files? Also, instead of using a key, can it used the ManagedIdentity credential? We are trying to avoid the use of API keys for security reasons.

pamelafox:
Please also look at `CONTRIBUTING.md` to see how you can run linters and write tests on your code.

sowu880:
> @sowu880 It seems that this PR doesn't include the creation of the speech resource. Can that be included as an optional resource in the Bicep files? Also, instead of using a key, can it used the ManagedIdentity credential? We are trying to avoid the use of API keys for security reasons.

Hi, we designed that user should use their existing speech resource rather than creating since it's not required.  I'm not sure it's necessary to add in Bicep file for creation.

sowu880:
> @sowu880 It seems that this PR doesn't include the creation of the speech resource. Can that be included as an optional resource in the Bicep files? Also, instead of using a key, can it used the ManagedIdentity credential? We are trying to avoid the use of API keys for security reasons.


> @sowu880 It seems that this PR doesn't include the creation of the speech resource. Can that be included as an optional resource in the Bicep files? Also, instead of using a key, can it used the ManagedIdentity credential? We are trying to avoid the use of API keys for security reasons.

Our SDK only support key for now.

sowu880:
> @sowu880 It seems that this PR doesn't include the creation of the speech resource. Can that be included as an optional resource in the Bicep files? Also, instead of using a key, can it used the ManagedIdentity credential? We are trying to avoid the use of API keys for security reasons.

Our SDK only support key for now.

pamelafox:
@sowu880 I see "auth_token" on https://learn.microsoft.com/en-us/python/api/azure-cognitiveservices-speech/azure.cognitiveservices.speech.speechconfig?view=azure-python , is that a different kind of auth token than the kind you can get from AzureDefaultCredential?
I found this snippet that seemed to use it like that:
https://github.com/csiebler/azure-cognitive-services-snippets/blob/a60a9a8c06c00ea52e0eccb702cba456f3547e07/aad-authentication/speech.py#L12

pamelafox:
For this repo, we aim to have all resources created by Bicep, so that deploys can be replicable. We can still support optional features in Bicep, you can take a look at how Application Insights was added, via a bool parameter and conditionals in Bicep.

sowu880:
> @sowu880 I see "auth_token" on https://learn.microsoft.com/en-us/python/api/azure-cognitiveservices-speech/azure.cognitiveservices.speech.speechconfig?view=azure-python , is that a different kind of auth token than the kind you can get from AzureDefaultCredential? I found this snippet that seemed to use it like that: https://github.com/csiebler/azure-cognitive-services-snippets/blob/a60a9a8c06c00ea52e0eccb702cba456f3547e07/aad-authentication/speech.py#L12

Hi, these comments have been fixed, Here are updates:
1. The speech resource will be created from main.bicep by default. 
2. Customers can set 'useSpeechResource' to false if don't need speech. 
3. Customers can still use their own speech resource by setting speechServiceName and speechResourceGroupName the same as openai resource.
4. AAD auth is used and key auth is removed. Aad token will be refreshed if not valid the same way with openai token.
5. Tests are added in test_app.py

arsalanmubeen (approved):

arsalanmubeen:
I'm currently working on implementing text-to-speech functionality, and I'm encountering an error in the Chat.tsx and Ask.tsx files. Specifically, the issue arises at the line:
 
 speechUrl = await getSpeechApi(parsedResponse.answer);
 
I've reviewed the code, but I'm unable to identify the root cause of the error. Your assistance in resolving this matter would be greatly appreciated.

Thank you in advance.

sowu880:
@arsalanmubeen Hi, could you share more error details or log?

DeepAsmani:
Hi @pamelafox , Can you help me how set up speech to text functionality in current working Chat/Ask BOT.
Thanks!

daptatea:
This is a very handy feature and I would very much appreciate it if this Pull Request was revisited. I notice that feature is available in the https://github.com/Azure-Samples/chat-with-your-data-solution-accelerator repository if someone want inspiration for how to do it!

pamelafox:
@daptatea Will be merged soon!

pamelafox:
This can be tried out here: https://app-backend-5hhse4yls5chk.azurewebsites.net/

pamelafox:
@zedhaque Is #2 something that you think should be configured on a per-app setting or a per-user setting? There was another setting originally in this PR for "speak all answers" that really felt like it should be a user setting, so I removed it to simplify the PR and defer the decision on user settings UI.

For #1) I agree, I'll do that, we need to make it easy to opt out of a default of en-US.

pamelafox:
@zedhaque I've made your suggested changes to split input/output and add a voice option. I also made named the input as INPUT_BROWSER and output as OUTPUT_AZURE as I could imagine us adding INPUT_AZURE or OUTPUT_BROWSER in the future.

john0isaac:
@pamelafox I deployed a version that solely depends on the [Web Speech API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Speech_API) in the recognition and synthesis of the speech it's for free.
You can test it here:
https://dfbsfb-lh4hrrtgs4a42-appservice.azurewebsites.net/

This is the PR where I added it:
https://github.com/khelanmodi/build-24-langchain-vcore/pull/47

It's based on the same changes you have here for the speech recognition part but depends on the same tool (Web Speech API) for speech synthesis instead of the Azure Speech API.

You might ask why is the synthesized voice bad.
This is the default en-us voice, it's called David and is available on most browsers.
You can use better voices from the list available here: https://mdn.github.io/dom-examples/web-speech-api/speak-easy-synthesis/
But each browser has its own set of available voices.
When you change the browser using this URL the list of available voices will change which is why I settled for the default one as it's available on most browsers but I think with some extra work this can be customized or even added as a drop-down to the developer settings.

sowu880:
@john0isaac @pamelafox 
The reason we use Azure Speech API because of the voice quality and prosody, and our speech team have released many conversational voice recently. [Try new voices](https://techcommunity.microsoft.com/t5/ai-azure-ai-services-blog/9-more-realistic-ai-voices-for-conversations-now-generally/ba-p/4099471). And many of our new voices can beat all competitors in the current marketing. That the reason why we highly recommend to use azure speech resource and we have a big team to support and maintain these product voice.

My suggestion is merge this "speak out" feature first. Now, and then we can continuously upgrade it on other requirements.

app/frontend/src/pages/chat/Chat.tsx
```diff
@@ -77,10 +82,12 @@ const Chat = () => {
                 setRetrievalMode(RetrievalMode.Text);
             }
             setShowUserUpload(config.showUserUpload);
+            setShowSpeechInput(config.showSpeechInput);
+            setShowSpeechOutput(config.showSpeechOutput);
         });
     };
 
-    const handleAsyncRequest = async (question: string, answers: [string, ChatAppResponse][], setAnswers: Function, responseBody: ReadableStream<any>) => {
+    const handleAsyncRequest = async (question: string, answers: [string, ChatAppResponse][], responseBody: ReadableStream<any>) => {
```
pamelafox, line 90: Removed unused setAnswers function from signature and call below

pamelafox:
@john0isaac Thank you for sharing that, super helpful. I just tried it out and it even works in Edge on Mac (where the browser Speech Recognition does *not* work yet, sadly). I do agree with @sowu880 that the Azure voices are much more fluid, and I also selected a default for this PR that has the broadest language support possible, since developers use this repo across many languages.

So I think we should get this PR merged, and then could you send a PR to add a USE_SPEECH_OUTPUT_BROWSER option? That should be fairly compatible with the way I've modularized this PR, I think. Either the SpeechOutput component could take an additional `answer` prop and a `enableBrowserOutput` bool, or there could be a different SpeechOutputBrowser vs SpeechOutputAzure component.

I've asked @mattgotteiner to take a look at this PR now, since it's a large change and large changes can use multiple eyes.

infra/main.bicep
```diff
@@ -48,6 +48,11 @@ param azureOpenAiApiVersion string = ''
 
 param openAiServiceName string = ''
 param openAiResourceGroupName string = ''
+
+param speechResourceGroupName string = ''
+param speechResourceGroupLocation string = location
```
mattgotteiner, line 53: Recommend adding AZURE_SPEEECH_LOCATION so that existing speech services can be used

infra/main.bicep
```diff
@@ -173,7 +182,6 @@ param useLocalHtmlParser bool = false
 var abbrs = loadJsonContent('abbreviations.json')
 var resourceToken = toLower(uniqueString(subscription().id, environmentName, location))
 var tags = { 'azd-env-name': environmentName }
```
mattgotteiner, line 184: why was computerVisionName removed?

infra/main.parameters.json
```diff
@@ -149,6 +149,18 @@
...
+    "useSpeechInputBrowser": {
+      "value": "${USE_SPEECH_INPUT_BROWSER=false}"
+    },
+    "useSpeechOutputAzure": {
+      "value": "${USE_SPEECH_OUTPUT_AZURE=false}"
+    },
+    "speechServiceName": {
+      "value": "${AZURE_SPEECH_SERVICE}"
+    },
+    "speechResourceGroupName": {
```
mattgotteiner, line 161: Recommend adding speech location here as a parameter

app/backend/app.py
```diff
@@ -229,10 +244,51 @@ def config():
...
+    if speech_token is None or speech_token.expires_on < time.time() + 60:
+        speech_token = await current_app.config[CONFIG_CREDENTIAL].get_token(
+            "https://cognitiveservices.azure.com/.default"
+        )
+        current_app.config[CONFIG_SPEECH_TOKEN] = speech_token
+
+    request_json = await request.get_json()
+    text = request_json["text"]
+    try:
+        auth_token = (
```
mattgotteiner, line 268: Please link to the  doc on why we are constructing the auth token this way
https://learn.microsoft.com/en-us/azure/ai-services/speech-service/how-to-configure-azure-ad-auth?tabs=portal&pivots=programming-language-python#speechrecognizer-speechsynthesizer-intentrecognizer-conversationtranscriber

app/backend/app.py
```diff
@@ -229,10 +244,51 @@ def config():
...
+            "aad#" + current_app.config[CONFIG_SPEECH_SERVICE_ID] + "#" + current_app.config[CONFIG_SPEECH_TOKEN].token
+        )
+        speech_config = SpeechConfig(auth_token=auth_token, region=current_app.config[CONFIG_SPEECH_REGION])
+        speech_config.speech_synthesis_voice_name = current_app.config[CONFIG_SPEECH_VOICE]
+        speech_config.speech_synthesis_output_format = SpeechSynthesisOutputFormat.Audio16Khz32KBitRateMonoMp3
+        synthesizer = SpeechSynthesizer(speech_config=speech_config, audio_config=None)
+        result: SpeechSynthesisResult = synthesizer.speak_text_async(text).get()
```
mattgotteiner, line 275: Does the .get() call block? Looks like other things in this call use await

app/backend/app.py
```diff
@@ -229,10 +244,51 @@ def config():
...
+        if result.reason == ResultReason.SynthesizingAudioCompleted:
+            return result.audio_data, 200, {"Content-Type": "audio/mp3"}
+        elif result.reason == ResultReason.Canceled:
+            cancellation_details = result.cancellation_details
+            current_app.logger.error(
+                "Speech synthesis canceled: %s %s", cancellation_details.reason, cancellation_details.error_details
+            )
+            raise Exception("Speech synthesis canceled. Check logs for details.")
```
mattgotteiner, line 283: Should we include the reason why in the exception? Not sure if we are still including that in the error page @pamelafox

pamelafox: We generally don't include reasons in the user-sent errors unless we think they're user-solvable. Instead we log those out. Seems better security-wise.

app/backend/app.py
pamelafox: It is not an async function, from Python's perspective. They implemented their own version of Promise's, it seems:

```
    def speak_text_async(self, text: str) -> ResultFuture:
        """
        Performs synthesis on plain text in a non-blocking (asynchronous) mode.

        :returns: A future with SpeechSynthesisResult.
        """
        async_handle = _spx_handle(0)
        c_text = _c_str(text)
        text_length = ctypes.c_uint32(len(c_text))
        _call_hr_fn(fn=_sdk_lib.synthesizer_speak_text_async, *[self._handle, c_text, text_length, ctypes.byref(async_handle)])

        def resolve_future(handle: _spx_handle):
            result_handle = _spx_handle(0)
            _call_hr_fn(fn=_sdk_lib.synthesizer_speak_async_wait_for, *[handle, max_uint32, ctypes.byref(result_handle)])
            _sdk_lib.synthesizer_async_handle_release(handle)
            return result_handle
        return ResultFuture(async_handle, resolve_future, SpeechSynthesisResult)
```

```
class ResultFuture():
    """
    The result of an asynchronous operation.
    """

    def __init__(self, async_handle, get_function: Callable[[_spx_handle], _spx_handle], wrapped_type):
        """
        private constructor
        """
        self._handle = async_handle
        self.__get_function = get_function
        self.__wrapped_type = wrapped_type
        self.__resolved = False

    def get(self):
        """
        Waits until the result is available, and returns it.
        """
        if not self.__resolved:
            result_handle = self.__get_function(self._handle)
            if self.__wrapped_type is not None:
                self.__result = self.__wrapped_type(result_handle)
            self.__resolved = True
        return self.__result if self.__wrapped_type is not None else None
```

pamelafox:
I've added parameters that allow for overriding location, resource group, service name, and sku. Also renamed some parameters for greater consistency. All feedback has been addressed, I believe.

mattgotteiner (approved):

PR #135, opened 2023-04-19
Update README.md

msdpalam:
## Purpose

* ...
Found that without the py launcher in the path, the `azd up` is failing. I feel that we need to add a bit more clarification on py launcher dependency for azd up, so adding the following: "Important - Please note that azd up would look for py launcher, make sure py launcher is in the path system environment variable. After successful installation of Python (with py launcher selected in the installer), put the py location (i.e., C:\Windows) in the system path variable.

Please review and let me know if we can add this change. It took me some time to figure this out, hopefully will save some end user time seeing the same behavior. One of my customers had this issue. Even though I had Python in path, azd up kept failing and only way to fix this was to put the py launcher location in the path system variable.

```
[ ] Yes
[ x] No
```

## Pull Request Type
Documentation

```
[ ] Bugfix
[ ] Feature
[ x] Documentation content changes
[ ] Other... Please describe:
```

## How to Test
*  Get the code
No code changes

* Test the code

```
```

## What to Check
* ...
Review the information added in the README.md and test if needed. Please reach out, if I can help with the clarification

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,8 @@ The repo includes sample data so it's ready to try end to end. In this sample ap
 - [Azure Developer CLI](https://aka.ms/azure-dev/install)
 - [Python 3+](https://www.python.org/downloads/)
     - **Important**: Python and the pip package manager must be in the path in Windows for the setup scripts to work.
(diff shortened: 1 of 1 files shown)


tonybaloney:
I don't have the py launcher and was able to run this template successfully, what OS are you running on?

PR #153, opened 2023-04-26
Fix path to env file in README

openjck:
## Purpose
Fix the path the the .env file in the "Sharing Environments" section of the README. I believe this new path is correct, but please correct me if I'm wrong.

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
* Follow the the updated documentation.

## What to Check
* The "Sharing Environments" section of the README


pamelafox (approved):
"./azure" also works in my environment (zsh), but I think ".azure" is clearer. Thanks for the PR!

PR #207, opened 2023-05-16
FIX: Prompt override error in the Front end

leongj:
## Purpose
Currently, if you try to use the "Prompt Override" in Developer Tools, you will get a Python error coming back in the Chat response. This is because in `chatreadretrieveread.py` the `injected_prompt` is not being populated in the `prompt_override.format` call (as it is in the other cases of the `if` statement). In this case, `injected_prompt` is irrelevant, so I have set it to `""` as per the un-overriden case.

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Open Developer Tools and enter something into the "Override prompt template".
* Ask a question.
* Response should return without an error.

## What to Check
* Response should return without an error.

## Other Information
n/a

### DIFF
--- a/app/backend/approaches/chatreadretrieveread.py
+++ b/app/backend/approaches/chatreadretrieveread.py
@@ -92,7 +92,7 @@ Search query:
         elif prompt_override.startswith(">>>"):
             prompt = self.prompt_prefix.format(injected_prompt=prompt_override[3:] + "\n", sources=content, chat_history=self.get_chat_history_as_text(history), follow_up_questions_prompt=follow_up_questions_prompt)
         else:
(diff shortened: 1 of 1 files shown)


pamelafox:
Sorry we missed this PR for so long. I've sent an updated fix in this PR https://github.com/Azure-Samples/azure-search-openai-demo/pull/606 so I'll close this one. Thanks so much!

Issue #208, opened 2023-05-16
403 Forbidden Error Thrown with message "Failed: Deploying service backend"


iamramengirl:
> Please provide us with the following information:
> ---------------------------------------------------------------

### This issue is for a: (mark with an `x`)
```
- [X] bug report -> please search issues before submitting
- [ ] feature request
- [X] documentation issue or request
- [ ] regression (a behavior that used to work and stopped in a new release)
```

### Minimal steps to reproduce
> Followed steps in the main page.

### Any log messages given by the failure
> A customer has attempted to deploy this application. They were able to successfully provision all the resources. However, a 403 forbidden error was thrown at the step [azd deploy](https://nam06.safelinks.protection.outlook.com/?url=https%3A%2F%2Flearn.microsoft.com%2Fen-us%2Fazure%2Fdeveloper%2Fazure-developer-cli%2Freference%23azd-deploy&data=05%7C01%7Cmaurivera%40microsoft.com%7Cd86b4ef4c67a41783d3308db52fae73d%7C72f988bf86f141af91ab2d7cd011db47%7C1%7C0%7C638195009871797105%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C3000%7C%7C%7C&sdata=D3P9mmtrtGf9FwMqIJr4z8838jiJa8rRAH8O0PodaAU%3D&reserved=0) which I believe is related to the deployment (publish) of the web application / pushing the code to the app service.

![image](https://github.com/Azure-Samples/azure-search-openai-demo/assets/19453096/fc59943a-3763-416c-844d-710c7ea3b509)

Specifically, the error occurred during the deployment of the application service backend.

![image](https://github.com/Azure-Samples/azure-search-openai-demo/assets/19453096/a5ae0e70-4846-44f2-941c-44f4a9ff2173)

### Expected/desired behavior
> Successful deployment of application.

### OS and Version?
> Windows 7, 8 or 10. Linux (which distribution). macOS (Yosemite? El Capitan? Sierra?)

### azd version?
> run `azd version` and copy paste here.

### Versions
>

### Mention any other details that might be useful

> ---------------------------------------------------------------
> Thanks! We'll be in touch soon.

iamramengirl:
Following up on this. Thank you!

vhvb1989:
[...]

Issue #189, opened 2023-05-09
Error Running prepdocs.py


sqlshep:
> Please provide us with the following information:
> ---------------------------------------------------------------

### This issue is for a: (mark with an `x`)
```
- [ X] bug report -> please search issues before submitting
- [ ] feature request
- [ ] documentation issue or request
- [ ] regression (a behavior that used to work and stopped in a new release)
```

### Minimal steps to reproduce
>. Follow the steps in the git repo for new or for existing services.  Fails at azd up

### Any log messages given by the failure
>Running "prepdocs.py"
Ensuring search index gptkbindex exists
Traceback (most recent call last):
  File "./scripts/prepdocs.py", line 299, in <module>
    create_search_index()
  File "./scripts/prepdocs.py", line 237, in create_search_index
    if args.index not in index_client.list_index_names():
  File "/Users/shepsheppard/scripts/.venv/lib/python3.8/site-packages/azure/core/paging.py", line 124, in __next__
    return next(self._page_iterator)
  File "/Users/shepsheppard/scripts/.venv/lib/python3.8/site-packages/azure/core/paging.py", line 76, in __next__
    self._response = self._get_next(self.continuation_token)
  File "/Users/shepsheppard/scripts/.venv/lib/python3.8/site-packages/azure/search/documents/indexes/_generated/operations/_indexes_operations.py", line 520, in get_next
    raise HttpResponseError(response=response, model=error)
azure.core.exceptions.HttpResponseError: () Authorization failed.
Code: 
Message: Authorization failed.

ERROR: failed running post hooks: 'postprovision' hook failed with exit code: '1', Path: './scripts/prepdocs.sh'. : exit status 1

### Expected/desired behavior
> Not fail

### OS and Version?
>  macOS Ventura 

### Versions
> 13.3.1(a)

### Mention any other details that might be useful

> ---------------------------------------------------------------
> Thanks! We'll be in touch soon.

Goro2030:
Same error here. Maybe has to do with permissions?

vikramhn:
https://github.com/Azure-Samples/azure-search-openai-demo/issues/79#issuecomment-1507899444 
"Search Index Data Contributor" role was already there. What fixed it for me was to add "Search Service Contributor" role to my ID.

Goro2030:
What solved it to me, was to add me as "OWNER" of the "gptkb-d3v5axy2dfrvs" service. That fixed the problem!

[...]

Issue #78, opened 2023-03-31
Authorization failed


gitpsk:
Requirement already satisfied: charset-normalizer<4,>=2 in ./scripts/.venv/lib/python3.9/site-packages (from requests>=2.18.4->azure-core<2.0.0,>=1.23.0->azure-ai-formrecognizer==3.2.1->-r ./scripts/requirements.txt (line 4)) (3.1.0)
Requirement already satisfied: idna<4,>=2.5 in ./scripts/.venv/lib/python3.9/site-packages (from requests>=2.18.4->azure-core<2.0.0,>=1.23.0->azure-ai-formrecognizer==3.2.1->-r ./scripts/requirements.txt (line 4)) (3.4)
Requirement already satisfied: urllib3<1.27,>=1.21.1 in ./scripts/.venv/lib/python3.9/site-packages (from requests>=2.18.4->azure-core<2.0.0,>=1.23.0->azure-ai-formrecognizer==3.2.1->-r ./scripts/requirements.txt (line 4)) (1.26.15)
Requirement already satisfied: oauthlib>=3.0.0 in ./scripts/.venv/lib/python3.9/site-packages (from requests-oauthlib>=0.5.0->msrest>=0.6.21->azure-ai-formrecognizer==3.2.1->-r ./scripts/requirements.txt (line 4)) (3.2.2)
Running "prepdocs.py"
Ensuring search index gptkbindex exists
Traceback (most recent call last):
  File "/workspaces/azure-search-openai-demo/./scripts/prepdocs.py", line 299, in <module>
    create_search_index()
  File "/workspaces/azure-search-openai-demo/./scripts/prepdocs.py", line 237, in create_search_index
    if args.index not in index_client.list_index_names():
  File "/workspaces/azure-search-openai-demo/scripts/.venv/lib/python3.9/site-packages/azure/core/paging.py", line 132, in __next__
    return next(self._page_iterator)
  File "/workspaces/azure-search-openai-demo/scripts/.venv/lib/python3.9/site-packages/azure/core/paging.py", line 76, in __next__
    self._response = self._get_next(self.continuation_token)
  File "/workspaces/azure-search-openai-demo/scripts/.venv/lib/python3.9/site-packages/azure/search/documents/indexes/_generated/operations/_indexes_operations.py", line 520, in get_next
    raise HttpResponseError(response=response, model=error)
azure.core.exceptions.HttpResponseError: () Authorization failed.
Code: 
[...]

PR #220, opened 2023-05-18
Fix non-subscription owner scenarios by adding Search Service Contributor role

pablocastro:
## Purpose
Listing and creating search indexes requires that the user has either Owner or Search Service Contributor RBAC roles. For cases where the user is not subscription owner, this change makes things work by adding the Search Service Contributor role for the user running "azd up".

```
[ ] Yes
[x] No
```
## Pull Request Type

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

### DIFF
--- a/infra/main.bicep
+++ b/infra/main.bicep
@@ -268,6 +268,16 @@ module searchContribRoleUser 'core/security/role.bicep' = {
   }
 }
 
+module searchSvcContribRoleUser 'core/security/role.bicep' = {
+  scope: searchServiceResourceGroup
+  name: 'search-svccontrib-role-user'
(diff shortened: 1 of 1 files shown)


pablocastro:
This should address issues #208, #186, #40, #189, #78

vhvb1989 (approved):

BitcoinCode:
In my case I am running the repo on Ubuntu and the main issue was that the .venv file is starting with (.) which makes it invisible the other issue I was trying to create the virtual environment on external drive that did not have the correct format so I have moved the repo to the ubuntu machine

PR #221, opened 2023-05-18
Update prepdocs.sh to use python3

pamelafox:
## Purpose

It's generally safer to use `python3` than `python` since the latter is an alias (and does not work on my Mac). `python3` should exist, however, if Python 3 is involved.

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[X] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
* Run `azd up` on Mac
* Don't see error about python: command not found

### DIFF
--- a/scripts/prepdocs.sh
+++ b/scripts/prepdocs.sh
@@ -12,7 +12,7 @@ $(azd env get-values)
 EOF
 
 echo 'Creating python virtual environment "scripts/.venv"'
-python -m venv scripts/.venv
+python3 -m venv scripts/.venv
 
 echo 'Installing dependencies from "requirements.txt" into virtual environment'
 ./scripts/.venv/bin/python -m pip install -r scripts/requirements.txt
--- a/app/start.sh
+++ b/app/start.sh
@@ -17,7 +17,7 @@ if [ $? -ne 0 ]; then
 fi
 
 echo 'Creating python virtual environment "backend/backend_env"'
-python -m venv backend/backend_env
+python3 -m venv backend/backend_env
 
 echo ""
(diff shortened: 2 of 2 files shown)


scripts/prepdocs.sh
```diff
@@ -12,7 +12,7 @@ $(azd env get-values)
 EOF
 
 echo 'Creating python virtual environment "scripts/.venv"'
-python -m venv scripts/.venv
+python3 -m venv scripts/.venv
```
vhvb1989, line 15: Would you mind doing the same for: https://github.com/Azure-Samples/azure-search-openai-demo/blob/main/app/start.sh#L20 ?

vhvb1989: I think this is a good improvement.

In the future, we might want to update the sh script to do auto-detect..  Similar to what we do for the pwsh version (https://github.com/Azure-Samples/azure-search-openai-demo/blob/main/scripts/prepdocs.ps1#L19)

vhvb1989 (approved):

scripts/prepdocs.sh
pamelafox: Sure, updated start.sh. I'm curious what situations there are where python is preferred over python3, as we currently have in our internal Python team style guide to prefer the explicit python3 command. I suppose someone might map python to a particular install of python3, and thats the situation you're trying to support?

tonybaloney: PEP394 is the best description of the aliases that Linux distributions follow https://peps.python.org/pep-0394/

PR #229, opened 2023-05-22
Update read, retrieve, read approach

mattmsft:
## Purpose

The lookup tool needs to parse the input to find the employee name. This parsed name should not be discarded, instead it should be used to lookup the employee

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[X] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code
* Try the following question using the read, retrieve, read approach. Employee2's information should be retrieved
What is Employee2's health plan?

### DIFF
--- a/app/backend/approaches/readretrieveread.py
+++ b/app/backend/approaches/readretrieveread.py
@@ -108,5 +108,5 @@ class EmployeeInfoTool(CsvLookupTool):
         self.func = self.employee_info
         self.employee_name = employee_name
 
-    def employee_info(self, unused: str) -> str:
-        return self.lookup(self.employee_name)
+    def employee_info(self, name: str) -> str:
+        return self.lookup(name)


app/backend/approaches/readretrieveread.py
```diff
@@ -108,5 +108,5 @@ def __init__(self, employee_name: str):
         self.func = self.employee_info
         self.employee_name = employee_name
 
-    def employee_info(self, unused: str) -> str:
-        return self.lookup(self.employee_name)
+    def employee_info(self, name: str) -> str:
+        return self.lookup(name)
```
tonybaloney, line 112: This makes sense, please change the type signature of the base method (`lookup`) to be `str` instead of `Optional[str]` since it will always return a str and if you're just returning the value of `lookup` then the signature should match that of `func` which is `Callable[..., str]`

https://github.com/Azure-Samples/azure-search-openai-demo/blob/main/app/backend/lookuptool.py#L19

tonybaloney: I'll address this post #339

tonybaloney (approved):

PR #263, opened 2023-06-02
Keep the Approach base class up to date

gukoff:
## Purpose

Approach base class was not up to date with the newer `run()` schemas.

Using `abc`, we get more confidence that all subclasses conform with the declared schema.

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

### DIFF
--- a/app/backend/approaches/approach.py
+++ b/app/backend/approaches/approach.py
@@ -1,6 +1,14 @@
+from abc import abstractmethod, ABC
 from typing import Any
 
 
-class Approach:
+class ChatApproach(ABC):
+    @abstractmethod
+    def run(self, history: list[dict], overrides: dict[str, Any]) -> Any:
+        ...
+
+
+class AskApproach(ABC):
+    @abstractmethod
     def run(self, q: str, overrides: dict[str, Any]) -> Any:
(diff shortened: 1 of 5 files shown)


tonybaloney:
Please rebase with the new base signatures. I think that would make Chat and Ask the following:

```python
def run(self, history: Sequence[dict[str, str]], overrides: dict[str, Any]) -> Any:
  ...
```

```python
def run(self, q: str, overrides: dict[str, Any]) -> Any:
 ...
```

tonybaloney (changes requested):
Please rebase with the new base signatures. I think that would make Chat and Ask the following:

```python
def run(self, history: Sequence[dict[str, str]], overrides: dict[str, Any]) -> Any:
  ...
```

```python
def run(self, q: str, overrides: dict[str, Any]) -> Any:
 ...
```

gukoff:
@tonybaloney done!

pamelafox:
@gukoff Thanks for this PR, I ended up making the same changes in a recent PR that changed the interface, so we don't need to merge this anymore.

PR #325, opened 2023-06-16
Add GHA for AZD and Python backend, add basic tests for API

tonybaloney:
This PR adds some automated validation to the project

- Checks the AZD/bicep compilation
- Checks the AZD template for best practice using Defender to DevOps
- Runs a small test in Python

I noticed a couple of things in the Python code that need fixing, will submit in a separate PR


tonybaloney:
Tests are failing , it was using the VS Code Authentication locally, let me work out an authentication model that the tests can mock out.

.github/workflows/backend-validation.yaml
```diff
@@ -0,0 +1,37 @@
...
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v3
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      working-directory: app/backend
+      run: |
+        python -m pip install --upgrade pip
+        pip install -r requirements.txt
+        pip install -r requirements-test.txt
```
pamelafox, line 29: Any reason not to have requirements-test.txt recursively include requirements.txt? (I'm just curious, as that's what I typically do).

.github/workflows/backend-validation.yaml
```diff
@@ -0,0 +1,37 @@
...
+
+    - name: Run pytest
+      working-directory: app/backend
+      run: python -m pytest
+      env:
+        AZURE_CLIENT_ID: 00000000-0000-0000-0000-000000000000
+        AZURE_CLIENT_SECRET: Not-The-Client-Secret
+        AZURE_TENANT_ID: 00000000-0000-0000-0000-000000000000
```
pamelafox, line 37: This is still failing, right? https://github.com/Azure-Samples/azure-search-openai-demo/actions/runs/5286159980/jobs/9565327057?pr=325
If this app was constructed using the app factory pattern, I think it'd be easier to pass in a mocked credential.
Are there any env variables that will get it past the Default credential stage?

tonybaloney: I eventually got the app factory pattern working so I could patch out `DefaultAzureCredential` during the tests.

PR #339, opened 2023-06-20
Fix some of the type annotations

tonybaloney:
These were identified by adding the following to `.vscode/settings.json`

```javascript
    "python.analysis.diagnosticMode": "workspace",
    "python.analysis.autoSearchPaths": true,
    "python.analysis.typeCheckingMode": "basic"
```

### DIFF
--- a/app/backend/approaches/approach.py
+++ b/app/backend/approaches/approach.py
@@ -2,5 +2,5 @@ from typing import Any
 
 
 class Approach:
-    def run(self, q: str, use_summaries: bool) -> Any:
+    def run(self, q: str, overrides: dict[str, Any]) -> Any:
         raise NotImplementedError
--- a/app/backend/approaches/approach.py
+++ b/app/backend/approaches/approach.py
@@ -1,3 +1,6 @@
+from typing import Any
+
+
 class Approach:
-    def run(self, q: str, use_summaries: bool) -> any:
+    def run(self, q: str, use_summaries: bool) -> Any:
         raise NotImplementedError
--- a/app/backend/approaches/chatreadretrieveread.py
+++ b/app/backend/approaches/chatreadretrieveread.py
@@ -1,3 +1,5 @@
+from typing import Any, Sequence
+
 import openai
 from azure.search.documents import SearchClient
 from azure.search.documents.models import QueryType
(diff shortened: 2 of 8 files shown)


app/backend/approaches/approach.py
```diff
@@ -1,3 +1,6 @@
+from typing import Any
+
+
 class Approach:
-    def run(self, q: str, use_summaries: bool) -> any:
```
tonybaloney, line 2: There is no such type as `any`

app/backend/approaches/readdecomposeask.py
```diff
@@ -42,7 +42,7 @@ def search(self, q: str, overrides: dict) -> str:
             self.results = [doc[self.sourcepage_field] + ":" + nonewlines(doc[self.content_field][:500]) for doc in r]
         return "\n".join(self.results)
 
-    def lookup(self, q: str) -> str:
+    def lookup(self, q: str) -> Optional[str]:
```
tonybaloney, line 45: Since the last line of this method returns `None`, `str` is not a valid annotation as it can return None

app/backend/lookuptool.py
```diff
@@ -1,13 +1,13 @@
...
+from pathlib import Path
 from langchain.agents import Tool
 from langchain.callbacks.manager import Callbacks
 from typing import Optional
 
 class CsvLookupTool(Tool):
     data: dict[str, str] = {}
 
-    def __init__(self, filename: path, key_field: str, name: str = "lookup", 
+    def __init__(self, filename: str | Path, key_field: str, name: str = "lookup",
```
tonybaloney, line 10: `path` is not a type. `open()` accepts both str and Path from path lib, so this is a more obvious annotation of what the method is expecting

app/backend/approaches/readdecomposeask.py
pamelafox: Since you changed this, employee_info() should have same return type?

app/backend/approaches/chatreadretrieveread.py
```diff
@@ -48,7 +50,7 @@ def __init__(self, search_client: SearchClient, chatgpt_deployment: str, gpt_dep
         self.sourcepage_field = sourcepage_field
         self.content_field = content_field
 
-    def run(self, history: list[dict], overrides: dict) -> any:
+    def run(self, history: Sequence[dict[str, str]], overrides: dict[str, Any]) -> Any:
```
pamelafox, line 53: Similar to other comment, mypy doesn't like how this signature disagrees with Approach() signature.

pamelafox (changes requested):
I ran mypy over the files to see if it agreed. It's a bit pickier, so I tried to only comment on the stuff you changed, and you still may not feel the need to address its issues, but I do think a couple points are valid.

app/backend/approaches/readdecomposeask.py
tonybaloney: that's being addressed in another PR

pamelafox (approved):

PR #345, opened 2023-06-20
Use other approach for specifying default values

pamelafox:
## Purpose

This PR proposes a slightly simpler way of passing in default values to azd for the gpt parameters. It seems to work for me, but I'd love confirmation from someone on azd team (@vhvb1989) that this functionality is expected to work.

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[X] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test

*  Run azd up on a fresh environment
*  Run azd up on a fresh environment but with gpt names overriden via azd env set

### DIFF
--- a/infra/main.bicep
+++ b/infra/main.bicep
@@ -37,10 +37,10 @@ param formRecognizerResourceGroupLocation string = location
 
 param formRecognizerSkuName string = 'S0'
 
-param gptDeploymentName string = 'davinci'
+param gptDeploymentName string = '' // See main.parameters.json for default
 param gptDeploymentCapacity int = 30
 param gptModelName string = 'text-davinci-003'
-param chatGptDeploymentName string = 'chat'
+param chatGptDeploymentName string = '' // See main.parameters.json for default
 param chatGptDeploymentCapacity int = 30
 param chatGptModelName string = 'gpt-35-turbo'
 
--- a/infra/main.bicep
+++ b/infra/main.bicep
@@ -37,10 +37,10 @@ param formRecognizerResourceGroupLocation string = location
 
 param formRecognizerSkuName string = 'S0'
 
-param gptDeploymentName string = '' // See main.parameters.json for default
+param gptDeploymentName string // Set in main.parameters.json
 param gptDeploymentCapacity int = 30
 param gptModelName string = 'text-davinci-003'
-param chatGptDeploymentName string = '' // See main.parameters.json for default
+param chatGptDeploymentName string // Set in main.parameters.json
 param chatGptDeploymentCapacity int = 30
 param chatGptModelName string = 'gpt-35-turbo'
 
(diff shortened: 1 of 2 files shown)


infra/main.parameters.json
```diff
@@ -45,10 +45,10 @@
       "value": "${AZURE_STORAGE_RESOURCE_GROUP}"
     },
     "chatGptDeploymentName": {
-      "value": "${AZURE_OPENAI_CHATGPT_DEPLOYMENT}"
+      "value": "${AZURE_OPENAI_CHATGPT_DEPLOYMENT=chat}"
```
pamelafox, line 48: @vhvb1989 This is the functionality I'm wondering about, this way of specifying default values if the env var isn't found.

vhvb1989: Aren't you setting the default already with `param chatGptDeploymentName string = 'chat'` within the bicep file?

Having the default on bicep means that no matter who deploy the file, the default value will be used.
Having the default value on `main.parameters.json` means that only azd would know about that default value.

I'd go one way or the other, as having both might lead to sync issues, where someone might forget to update both values when updating the default value.

Any other thoughts about this @weikanglim ?

pamelafox: The goal is for this to support a few scenarios:

* Devs can specify the parameter by using `azd env set` before running `azd up`, and that gets passed in via main.parameters.json . Just setting `param = 'chat'` doesn't seem to support that scenario - I think because the empty env variable overrides it to an empty string? That's why we have the current workaround with checking if empty() and setting a new variable to 'chat' instead. Is it supposed to work?
* If devs don't specify the parameter, then it is instead set to "chat". That is satisfied by `param = 'chat'` as long as there isn't also main.parameters.json.

weikanglim: The syntax works as you expect it @pamelafox. Also agree with @vhvb1989 that it may be best to not have multiple places specifying defaults.

I think it reads better for me if the defaulting + the env var was both defined in `parameters.json` as the single line for the source of truth. This rule also avoids encountering possible errors when the variable isn't a string type.

pamelafox: Okay, I've updated this PR to keep the default values in main.parameters.json and I put a comment about that in main.bicep. Makes sense to DRY it up.

vhvb1989 (approved):

infra/main.bicep
```diff
@@ -37,10 +37,10 @@ param formRecognizerResourceGroupLocation string = location
 
 param formRecognizerSkuName string = 'S0'
 
-param gptDeploymentName string = ''
+param gptDeploymentName string = '' // See main.parameters.json for default
```
weikanglim, line 40:
```suggestion
param gptDeploymentName string
```

I think you'd want to remove this completely and make it a required parameter that is always set using the parameters file

jongio:
It should work, are you seeing any issues?

pamelafox:
@jongio This does seem to work, I just haven't figured out where it's documented, so I didn't know if I was relying on something that was actually supported.

PR #350, opened 2023-06-21
Validate python version and add a current dir to linux terminal.

piassa157:
## Purpose

A bug fix, when running the script in a Linux terminal, runs the risk of losing the file path for predocs.sh. For example, let's say I'm here /usr/dev/azuredemo... and I execute the command 'azd up'. The terminal will lose the path './scripts/' created by the venv. So, './scripts/.venv' won't find './scripts/fileOne.py' and vice versa.

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[x] Code style update (formatting, local variables)
[ ] Other... Please describe:
```

## How to Test
*  Get the code
ATTENTION: using a linux terminal. 
```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
```

* Test the code

To see a full flow of sample with correction in linux terminal.
```
azd up
```

To see a script runing solo in linux terminal.
It is necessary to be in the project's root directory.
```
./scripts/prepdocs.sh
```

## What to Check

prepdocs.sh

## Other Information
To verify the fix, use a WSL terminal in Windows or a Bash terminal emulator.

### DIFF
--- a/scripts/prepdocs.sh
+++ b/scripts/prepdocs.sh
@@ -4,6 +4,8 @@ echo ""
 echo "Loading azd .env file from current environment"
 echo ""
 
+current_dir=$(pwd)
+
 while IFS='=' read -r key value; do
     value=$(echo "$value" | sed 's/^"//' | sed 's/"$//')
     export "$key=$value"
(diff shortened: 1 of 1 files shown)


pamelafox:
Thanks for the contribution! We've elected not to make this change as it doesn't seem to be affecting a large number of developers.

PR #385, opened 2023-07-06
Upgrade sample to use ChatCompletion API

srbalakr:
## Purpose
Upgrade Sample to use ChatCompletion API of gpt-35-turbo model
* ...

```
[x ] Yes
[ ] No
```

## Pull Request Type
1. Update openAI lib to use chatcompletion api
2. Update api version to latest
3. Contract changes for chatcompletion scenario
4. use openAI tiktok lib to validate token

```
[ ] Bugfix
[x ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
```

## What to Check
* ...

### DIFF
--- a/app/backend/approaches/chatreadretrieveread.py
+++ b/app/backend/approaches/chatreadretrieveread.py
@@ -155,7 +155,7 @@ Search query:
         """
         Calculate the number of tokens required to encode a message.
         Args:
-            message (any): The message to encode, represented as a dictionary.
+            message (dict): The message to encode, represented as a dictionary.
             model (str): The name of the model to use for encoding.
         Returns:
             int: The total number of tokens required to encode the message.
--- a/app/backend/approaches/chatreadretrieveread.py
+++ b/app/backend/approaches/chatreadretrieveread.py
@@ -94,8 +94,6 @@ Search query:
         prompt_override = overrides.get("prompt_template")
         messages = self.get_messages_from_history(prompt_override=prompt_override, follow_up_questions_prompt=follow_up_questions_prompt,history=history, sources=content)
 
-        print(messages)
-
         # STEP 3: Generate a contextual and content specific answer using the search results and chat history
         chatCompletion = openai.ChatCompletion.create(
             deployment_id=self.chatgpt_deployment,
--- a/app/backend/approaches/chatreadretrieveread.py
+++ b/app/backend/approaches/chatreadretrieveread.py
@@ -105,7 +105,7 @@ Search query:
         
         chatContent = chatCompletion.choices[0].message.content
 
-        return {"data_points": results, "answer": chatContent, "thoughts": f"Searched for:<br>{q}<br><br>Prompt:<br>" + prompt.replace('\n', '<br>')}
+        return {"data_points": results, "answer": chat_content, "thoughts": f"Searched for:<br>{q}<br><br>Prompt:<br>" + prompt.replace('\n', '<br>')}
     
     def get_chat_history_as_text(self, history: Sequence[dict[str, str]], include_last_turn: bool=True, approx_max_tokens: int=1000) -> str:
         history_text = ""
--- a/app/backend/approaches/chatreadretrieveread.py
+++ b/app/backend/approaches/chatreadretrieveread.py
@@ -95,7 +95,7 @@ Search query:
         messages = self.get_messages_from_history(prompt_override=prompt_override, follow_up_questions_prompt=follow_up_questions_prompt,history=history, sources=content)
 
         # STEP 3: Generate a contextual and content specific answer using the search results and chat history
-        chatCompletion = openai.ChatCompletion.create(
+        chat_completion = openai.ChatCompletion.create(
             deployment_id=self.chatgpt_deployment,
             model=self.chatgpt_model,
             messages=messages, 
@@ -103,7 +103,7 @@ Search query:
             max_tokens=1024, 
             n=1)
         
-        chatContent = chatCompletion.choices[0].message.content
+        chat_content = chat_completion.choices[0].message.content
 
         return {"data_points": results, "answer": chat_content, "thoughts": f"Searched for:<br>{q}<br><br>Prompt:<br>" + prompt.replace('\n', '<br>')}
     
@@ -127,37 +127,53 @@ Search query:
             system_message = prompt_override.format(follow_up_questions_prompt=follow_up_questions_prompt)
 
         messages.append({"role":self.SYSTEM, "content": system_message})
-        token_count += self.num_tokens_from_messages(messages, self.chatgpt_model)
+        token_count += self.num_tokens_from_messages(messages[-1], self.chatgpt_model)
         
         #latest conversation
-        userContent = history[-1]["user"] + " \nSources:" + sources
+        user_content = history[-1]["user"] + " \nSources:" + sources
         messages.append({"role": self.USER, "content": user_content})
-        token_count += self.num_tokens_from_messages(messages, self.chatgpt_model)
+        token_count += token_count + self.num_tokens_from_messages(messages[-1], self.chatgpt_model)
 
         '''
         Enqueue in reverse order
         if limit exceeds truncate old messages 
         leaving system message behind
+        Keep track of token count for each conversation
+        If token count exceeds limit, break
         '''
         for h in reversed(history[:-1]):
             if h.get("bot"):
                 messages.insert(1, {"role": self.ASSISTANT, "content" : h.get("bot")})
+                token_count += self.num_tokens_from_messages(messages[1], self.chatgpt_model)
             messages.insert(1, {"role": self.USER, "content" : h.get("user")})
-            token_count = token_count + self.num_tokens_from_messages(messages, self.chatgpt_model)
+            token_count += self.num_tokens_from_messages(messages[1], self.chatgpt_model)
             if token_count > approx_max_tokens*4:
                 break
-
         return messages
     
-    def num_tokens_from_messages(self, messages, model: str):
+    '''
+    Source: https://github.com/openai/openai-cookbook/blob/main/examples/How_to_count_tokens_with_tiktoken.ipynb
+    Adapted: https://learn.microsoft.com/en-us/azure/cognitive-services/openai/how-to/chatgpt?pivots=programming-language-chat-completions#managing-conversations
+
+    Method takes in a single conversation and calculate prompt tokens
+    for chat api 
+
+    Keys role and content are accounted seperately.
+
+    Values of content are encoded by model type and calculated the length.
+    
+    This gives close proximity of token length measurement used in gpt models
+
+    message = {"role":"assistant", "content":"how can I assist you?"}
+    '''
+    def num_tokens_from_messages(self, message: any, model: str):
         encoding = tiktoken.encoding_for_model(self.get_oai_chatmodel_tiktok(model))
         num_tokens = 0
-        for message in messages:
-            num_tokens += 2  # every message follows {role/name}\n{content}\n
-            for key, value in message.items():
-                num_tokens += len(encoding.encode(value))
-                if key == "name":  # if there's a name, the role is omitted
-                    num_tokens += -1  # role is always required and always 1 token
+        num_tokens += 2  # every message follows {role/name}\n{content}\n
+        for key, value in message.items():
+            num_tokens += len(encoding.encode(value))
+            if key == "name":  # if there's a name, the role is omitted
+                num_tokens += -1  # role is always required and always 1 token
         return num_tokens
 
     def get_oai_chatmodel_tiktok(self, aoaimodel: str):
--- a/app/backend/approaches/chatreadretrieveread.py
+++ b/app/backend/approaches/chatreadretrieveread.py
@@ -131,7 +131,7 @@ Search query:
         
         #latest conversation
         userContent = history[-1]["user"] + " \nSources:" + sources
-        messages.append({"role": self.USER, "content": userContent})
+        messages.append({"role": self.USER, "content": user_content})
         token_count += self.num_tokens_from_messages(messages, self.chatgpt_model)
 
         '''
--- a/app/backend/approaches/chatreadretrieveread.py
+++ b/app/backend/approaches/chatreadretrieveread.py
@@ -115,8 +115,10 @@ Search query:
                 break    
         return history_text
     
-    # Generate messages needed for chat Completion api
     def get_messages_from_history(self, prompt_override, follow_up_questions_prompt, history: Sequence[dict[str, str]], sources: str, approx_max_tokens: int = 1000) -> []:
+        '''
+        Generate messages needed for chat Completion api
+        '''
         messages = []
         token_count = 0
         if prompt_override is None:
@@ -160,7 +162,7 @@ Search query:
         Returns:
             int: The total number of tokens required to encode the message.
         Example:
-            message = {'role': 'user', 'name': 'John', 'content': 'Hello, how are you?'}
+            message = {'role': 'user', 'content': 'Hello, how are you?'}
             model = 'gpt-3.5-turbo'
             num_tokens_from_messages(message, model)
             output: 11
--- a/app/backend/approaches/chatreadretrieveread.py
+++ b/app/backend/approaches/chatreadretrieveread.py
@@ -19,7 +19,7 @@ class ChatReadRetrieveReadApproach(Approach):
     (answer) with that prompt.
     """
     system_message_chat_conversation = """Assistant helps the company employees with their healthcare plan questions, and questions about the employee handbook. Be brief in your answers.
-Answer ONLY with the facts listed in the list of Sources:. If there isn't enough information below, say you don't know. Do not generate answers that don't use the sources below. If asking a clarifying question to the user would help, ask the question.
+Answer ONLY with the facts listed in the list of sources below. If there isn't enough information below, say you don't know. Do not generate answers that don't use the sources below. If asking a clarifying question to the user would help, ask the question.
 For tabular information return it as an html table. Do not return markdown format.
 Each source has a name followed by colon and the actual information, always include the source name for each fact you use in the response. Use square brackets to reference the source, e.g. [info1.txt]. Don't combine sources, list each source separately, e.g. [info1.txt][info2.pdf].
 {follow_up_questions_prompt}
@@ -130,7 +130,7 @@ Search query:
         token_count = token_count + self.num_tokens_from_messages(messages, self.chatgpt_model)
         
         #latest conversation
-        userContent = history[-1]["user"] + "\nSources:" + sources
+        userContent = history[-1]["user"] + " \nSources:" + sources
         messages.append({"role": self.USER, "content": userContent})
         token_count = token_count + self.num_tokens_from_messages(messages, self.chatgpt_model)
 
@@ -153,12 +153,11 @@ Search query:
         encoding = tiktoken.encoding_for_model(self.get_oai_chatmodel_tiktok(model))
         num_tokens = 0
         for message in messages:
-            num_tokens += 4  # every message follows <im_start>{role/name}\n{content}<im_end>\n
+            num_tokens += 2  # every message follows {role/name}\n{content}\n
             for key, value in message.items():
                 num_tokens += len(encoding.encode(value))
                 if key == "name":  # if there's a name, the role is omitted
                     num_tokens += -1  # role is always required and always 1 token
-        num_tokens += 2  # every reply is primed with <im_start>assistant
         return num_tokens
 
     def get_oai_chatmodel_tiktok(self, aoaimodel: str):
--- a/app/backend/approaches/chatreadretrieveread.py
+++ b/app/backend/approaches/chatreadretrieveread.py
@@ -99,11 +99,6 @@ Search query:
         else:
             prompt = prompt_override.format(sources=content, chat_history=self.get_chat_history_as_text(history), follow_up_questions_prompt=follow_up_questions_prompt)
 
-        print("history: \n")
-        print(history)
-        print("prompt: \n")
-        print(prompt)
-
         messages = self.get_messages_from_prompt(prompt)
 
         # STEP 3: Generate a contextual and content specific answer using the search results and chat history
@@ -128,6 +123,7 @@ Search query:
                 break    
         return history_text
     
+    # Generate messages needed for chat Completion api
     def get_messages_from_prompt(self, prompt: str) -> []:
         messages = []
         for line in prompt.splitlines():
--- a/app/backend/approaches/chatreadretrieveread.py
+++ b/app/backend/approaches/chatreadretrieveread.py
@@ -18,18 +18,6 @@ class ChatReadRetrieveReadApproach(Approach):
     top documents from search, then constructs a prompt with them, and then uses OpenAI to generate an completion
     (answer) with that prompt.
     """
-    prompt_prefix = """<|im_start|>system
-Assistant helps the company employees with their healthcare plan questions, and questions about the employee handbook. Be brief in your answers.
-Answer ONLY with the facts listed in the list of sources below. If there isn't enough information below, say you don't know. Do not generate answers that don't use the sources below. If asking a clarifying question to the user would help, ask the question.
-For tabular information return it as an html table. Do not return markdown format.
-Each source has a name followed by colon and the actual information, always include the source name for each fact you use in the response. Use square brackets to reference the source, e.g. [info1.txt]. Don't combine sources, list each source separately, e.g. [info1.txt][info2.pdf].
-{follow_up_questions_prompt}
-{injected_prompt}
-Sources:
-{sources}
-<|im_end|>
-{chat_history}
-"""
     system_message_chat_conversation = """Assistant helps the company employees with their healthcare plan questions, and questions about the employee handbook. Be brief in your answers.
 Answer ONLY with the facts listed in the list of Sources:. If there isn't enough information below, say you don't know. Do not generate answers that don't use the sources below. If asking a clarifying question to the user would help, ask the question.
 For tabular information return it as an html table. Do not return markdown format.
@@ -57,9 +45,10 @@ Question:
 Search query:
 """
 
-    def __init__(self, search_client: SearchClient, chatgpt_deployment: str, gpt_deployment: str, sourcepage_field: str, content_field: str):
+    def __init__(self, search_client: SearchClient, chatgpt_deployment: str, chatgpt_model: str, gpt_deployment: str, sourcepage_field: str, content_field: str):
         self.search_client = search_client
         self.chatgpt_deployment = chatgpt_deployment
+        self.chatgpt_model = chatgpt_model
         self.gpt_deployment = gpt_deployment
         self.sourcepage_field = sourcepage_field
         self.content_field = content_field
@@ -117,7 +106,7 @@ Search query:
         # STEP 3: Generate a contextual and content specific answer using the search results and chat history
         chatCompletion = openai.ChatCompletion.create(
             deployment_id=self.chatgpt_deployment,
-            model="gpt-3.5-turbo",
+            model=self.chatgpt_model,
             messages=messages, 
             temperature=overrides.get("temperature") or 0.7, 
             max_tokens=1024, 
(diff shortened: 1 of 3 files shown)


app/backend/approaches/chatreadretrieveread.py
```diff
@@ -117,3 +122,16 @@ def get_chat_history_as_text(self, history: Sequence[dict[str, str]], include_la
             if len(history_text) > approx_max_tokens*4:
                 break    
         return history_text
+    
+    # Generate messages needed for chat Completion api
+    def get_messages_from_prompt(self, prompt: str) -> []:
+        messages = []
+        for line in prompt.splitlines():
+            if line.startswith("<|im_start|>"):
+                index = "<|im_start|>".__len__()
```
pamelafox, line 131: I'm not sure why this is using the dunder method, I would typically use the built-in len().

Also, since line 131 will always yield the same answer, you can move it out of the loop, e.g.:

START = "<|im_start|>"
START_LEN = len(START)

messages = []
for line in prompt.splitlines():
    if line.startswith(START):
      role = line[START_LEN:]

That might be over-optimization though.

app/backend/approaches/chatreadretrieveread.py
```diff
@@ -99,16 +99,21 @@ def run(self, history: Sequence[dict[str, str]], overrides: dict[str, Any]) -> A
...
-            engine=self.chatgpt_deployment, 
-            prompt=prompt, 
+        chatCompletion = openai.ChatCompletion.create(
+            deployment_id=self.chatgpt_deployment,
+            model="gpt-3.5-turbo",
+            messages=messages, 
             temperature=overrides.get("temperature") or 0.7, 
             max_tokens=1024, 
             n=1, 
             stop=["<|im_end|>", "<|im_start|>"])
```
pamelafox, line 112: Does "<|im_start|>" still exist in the messages? Or just "<|im_end|>"? Just checking that the stop parameter is still correct.

app/backend/approaches/chatreadretrieveread.py
```diff
@@ -99,16 +99,21 @@ def run(self, history: Sequence[dict[str, str]], overrides: dict[str, Any]) -> A
...
 
+        messages = self.get_messages_from_prompt(prompt)
+
         # STEP 3: Generate a contextual and content specific answer using the search results and chat history
-        completion = openai.Completion.create(
-            engine=self.chatgpt_deployment, 
-            prompt=prompt, 
+        chatCompletion = openai.ChatCompletion.create(
+            deployment_id=self.chatgpt_deployment,
+            model="gpt-3.5-turbo",
```
pamelafox, line 107: Developers may override the model (like to use gpt-4), so it'd be nice to not hardcode it. It could be an additional environment variable, like we do for the other sample-

https://github.com/microsoft/sample-app-aoai-chatGPT/blob/main/app.py#L42

app/backend/approaches/chatreadretrieveread.py
pamelafox: Is it still necessary to use the im_start/im_end approach at all in this file? I wonder if we can move to a list for everything, or if thats incompatible with some aspect.

pablocastro:
@srbalakr aside from individual comments: I wonder if we should consider moving all OpenAI calls to use the chat API, so we can use Turbo by default for everything (or those that want to, can use GPT-4). It'll need a bit more rework, but it means we can use a single model deployment in the Azure OpenAI for all cases except embeddings.

app/backend/approaches/chatreadretrieveread.py
bharatsb: Hi @srbalakr, thank you for raising PR.

I have few suggestions to reduce the number of tokens in the generated `messages` array:
1. Have a single `role` and `content` pair for each role.
2. Strip leading and trailing whitespaces in each line.
3. Ignore any empty lines if first or last line of content for each role.

In current implementation, each empty line will be added as `{'role': role, 'content': ''}` in `messages` array, e.g., for `system` role, it will be 11 tokens for `{'role': system, 'content': ''}`.

For a given sample prompt,
```
prompt = """\
<|im_start|>system
A

B


C
<|im_end|>
<|im_start|>user

Who are you?

<|im_end|>
<|im_start|>assistant

"""
````

**Current implementation output:**
```
[{'role': 'system', 'content': 'A'}, {'role': 'system', 'content': ''}, {'role': 'system', 'content': 'B'}, {'role': 'system', 'content': ''}, {'role': 'system', 'content': ''}, {'role': 'system', 'content': 'C'}, {'role': 'user', 'content': ''}, {'role': 'user', 'content': 'Who are you?'}, {'role': 'user', 'content': ''}, {'role': 'assistant', 'content': ''}]
```
Tokens = 133

**Expected output:**
```
[{'role': 'system', 'content': 'A\n\nB\n\n\nC'}, {'role': 'user', 'content': 'Who are you?'}]
```
Tokens = 44

Comment (srbalakr) [reply app/backend/approaches/chatreadretrieveread.py]: @pamelafox  @bharatsb thank you. I have refactored the code to use tiktok lib to estimate tokens. It makes code efficient and simpler. Help take a look.

File: app/backend/approaches/chatreadretrieveread.py (hunk 2 above)
Comment (srbalakr) [reply app/backend/approaches/chatreadretrieveread.py]: We can remove stop keywords since new conversation templates do not expect it.

File: app/backend/approaches/chatreadretrieveread.py (hunk 3 above)
Comment (srbalakr) [reply app/backend/approaches/chatreadretrieveread.py]: agreed. I have added a new environment variable to taken in model name and pass to downstream components. However we need to update the value from provisioning. I will updat them in my next PR

**Comment by srbalakr** [comment]
> @srbalakr aside from individual comments: I wonder if we should consider moving all OpenAI calls to use the chat API, so we can use Turbo by default for everything (or those that want to, can use GPT-4). It'll need a bit more rework, but it means we can use a single model deployment in the Azure OpenAI for all cases except embeddings.

@pablocastro  agreed. Since this is my first PR, let me check in with current changes. I will put up a follow up PR to update biceps and all other completion api to chat completion api.

File: app/backend/approaches/chatreadretrieveread.py [hunk 4]
```diff
@@ -117,3 +114,55 @@ def get_chat_history_as_text(self, history: Sequence[dict[str, str]], include_la
...
+            if token_count > approx_max_tokens*4:
+                break
+
+        return messages
+    
+    def num_tokens_from_messages(self, messages, model: str):
+        encoding = tiktoken.encoding_for_model(self.get_oai_chatmodel_tiktok(model))
+        num_tokens = 0
+        for message in messages:
+            num_tokens += 4  # every message follows <im_start>{role/name}\n{content}<im_end>\n
```
chuwik, line 156: Outdated, <im_start> and <im_end> not used anymore

app/backend/approaches/chatreadretrieveread.py
```diff
@@ -117,3 +114,55 @@ def get_chat_history_as_text(self, history: Sequence[dict[str, str]], include_la
...
+            for key, value in message.items():
+                num_tokens += len(encoding.encode(value))
+                if key == "name":  # if there's a name, the role is omitted
+                    num_tokens += -1  # role is always required and always 1 token
+        num_tokens += 2  # every reply is primed with <im_start>assistant
```
chuwik, line 161: Outdated, <im_start> not used anymore

chuwik:
We should also update notebooks/chat-read-retrieve-read.ipynb

app/backend/approaches/chatreadretrieveread.py
```diff
@@ -1,31 +1,30 @@
...
     Simple retrieve-then-read implementation, using the Cognitive Search and OpenAI APIs directly. It first retrieves
     top documents from search, then constructs a prompt with them, and then uses OpenAI to generate an completion
     (answer) with that prompt.
     """
-
-    prompt_prefix = """<|im_start|>system
-Assistant helps the company employees with their healthcare plan questions, and questions about the employee handbook. Be brief in your answers.
-Answer ONLY with the facts listed in the list of sources below. If there isn't enough information below, say you don't know. Do not generate answers that don't use the sources below. If asking a clarifying question to the user would help, ask the question.
+    system_message_chat_conversation = """Assistant helps the company employees with their healthcare plan questions, and questions about the employee handbook. Be brief in your answers.
+Answer ONLY with the facts listed in the list of Sources:. If there isn't enough information below, say you don't know. Do not generate answers that don't use the sources below. If asking a clarifying question to the user would help, ask the question.
```
chuwik, line 22: Does 'Sources:' work like this? We use 'sources below' later in the system message, and before these changes.

app/backend/approaches/chatreadretrieveread.py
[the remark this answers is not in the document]
srbalakr: this change looks outdated, could you refresh the PR ?

app/backend/approaches/chatreadretrieveread.py
srbalakr: yes it works. Just as act as Keyword trigger and gptTurbo was able to pick up. But for parity, I have reverted back.

srbalakr:
> We should also update notebooks/chat-read-retrieve-read.ipynb

Ah, missed it. Thank you. Let me put up a follow up PR along with other chat completion api's

app/backend/approaches/chatreadretrieveread.py
```diff
@@ -117,3 +114,54 @@ def get_chat_history_as_text(self, history: Sequence[dict[str, str]], include_la
...
+        Enqueue in reverse order
+        if limit exceeds truncate old messages 
+        leaving system message behind
+        '''
+        for h in reversed(history[:-1]):
+            if h.get("bot"):
+                messages.insert(1, {"role": self.ASSISTANT, "content" : h.get("bot")})
+            messages.insert(1, {"role": self.USER, "content" : h.get("user")})
+            token_count = token_count + self.num_tokens_from_messages(messages, self.chatgpt_model)
+            if token_count > approx_max_tokens*4:
```
pamelafox, line 147: where does 4 come from?

app/backend/approaches/chatreadretrieveread.py
```diff
@@ -117,3 +114,54 @@ def get_chat_history_as_text(self, history: Sequence[dict[str, str]], include_la
...
+                break
+
+        return messages
+    
+    def num_tokens_from_messages(self, messages, model: str):
+        encoding = tiktoken.encoding_for_model(self.get_oai_chatmodel_tiktok(model))
+        num_tokens = 0
+        for message in messages:
+            num_tokens += 2  # every message follows {role/name}\n{content}\n
```
pamelafox, line 156: does the 2 come from the two new lines?

app/backend/approaches/chatreadretrieveread.py
```diff
@@ -117,3 +114,54 @@ def get_chat_history_as_text(self, history: Sequence[dict[str, str]], include_la
...
+            if h.get("bot"):
+                messages.insert(1, {"role": self.ASSISTANT, "content" : h.get("bot")})
+            messages.insert(1, {"role": self.USER, "content" : h.get("user")})
+            token_count = token_count + self.num_tokens_from_messages(messages, self.chatgpt_model)
+            if token_count > approx_max_tokens*4:
+                break
+
+        return messages
+    
+    def num_tokens_from_messages(self, messages, model: str):
```
pamelafox, line 152: I'm curious, does this cause any notable decrease in performance?

pamelafox, line 152: It'd be helpful to type messages as well, as a list of dicts.

app/backend/approaches/chatreadretrieveread.py
```diff
@@ -117,3 +114,54 @@ def get_chat_history_as_text(self, history: Sequence[dict[str, str]], include_la
...
+        encoding = tiktoken.encoding_for_model(self.get_oai_chatmodel_tiktok(model))
+        num_tokens = 0
+        for message in messages:
+            num_tokens += 2  # every message follows {role/name}\n{content}\n
+            for key, value in message.items():
+                num_tokens += len(encoding.encode(value))
+                if key == "name":  # if there's a name, the role is omitted
```
pamelafox, line 159: This gets a bit hard to follow. Perhaps this function could have a docstring with an example?

app/backend/approaches/chatreadretrieveread.py
```diff
@@ -117,3 +114,54 @@ def get_chat_history_as_text(self, history: Sequence[dict[str, str]], include_la
...
+        token_count = 0
+        if prompt_override is None:
+            system_message = self.system_message_chat_conversation.format(injected_prompt="", follow_up_questions_prompt=follow_up_questions_prompt)
+        elif prompt_override.startswith(">>>"):
+            system_message = self.system_message_chat_conversation.format(injected_prompt=prompt_override[3:] + "\n", follow_up_questions_prompt=follow_up_questions_prompt)
+        else:
+            system_message = prompt_override.format(follow_up_questions_prompt=follow_up_questions_prompt)
+
+        messages.append({"role":self.SYSTEM, "content": system_message})
+        token_count = token_count + self.num_tokens_from_messages(messages, self.chatgpt_model)
```
pamelafox, line 130:
```suggestion
        token_count += self.num_tokens_from_messages(messages, self.chatgpt_model)
```

app/backend/approaches/chatreadretrieveread.py
```diff
@@ -117,3 +114,54 @@ def get_chat_history_as_text(self, history: Sequence[dict[str, str]], include_la
...
+        
+        #latest conversation
+        userContent = history[-1]["user"] + " \nSources:" + sources
+        messages.append({"role": self.USER, "content": userContent})
+        token_count = token_count + self.num_tokens_from_messages(messages, self.chatgpt_model)
```
pamelafox, line 135:
```suggestion
        token_count += self.num_tokens_from_messages(messages, self.chatgpt_model)
```

app/backend/approaches/chatreadretrieveread.py
```diff
@@ -117,3 +114,54 @@ def get_chat_history_as_text(self, history: Sequence[dict[str, str]], include_la
...
+        elif prompt_override.startswith(">>>"):
+            system_message = self.system_message_chat_conversation.format(injected_prompt=prompt_override[3:] + "\n", follow_up_questions_prompt=follow_up_questions_prompt)
+        else:
+            system_message = prompt_override.format(follow_up_questions_prompt=follow_up_questions_prompt)
+
+        messages.append({"role":self.SYSTEM, "content": system_message})
+        token_count = token_count + self.num_tokens_from_messages(messages, self.chatgpt_model)
+        
+        #latest conversation
+        userContent = history[-1]["user"] + " \nSources:" + sources
```
pamelafox, line 133:
```suggestion
        user_content = history[-1]["user"] + " \nSources:" + sources
```

app/backend/approaches/chatreadretrieveread.py
```diff
@@ -117,3 +114,54 @@ def get_chat_history_as_text(self, history: Sequence[dict[str, str]], include_la
...
+        messages.append({"role": self.USER, "content": userContent})
```
pamelafox, line 134:
```suggestion
        messages.append({"role": self.USER, "content": user_content})
```

app/backend/approaches/chatreadretrieveread.py
```diff
@@ -92,23 +92,20 @@ def run(self, history: Sequence[dict[str, str]], overrides: dict[str, Any]) -> A
...
+            deployment_id=self.chatgpt_deployment,
+            model=self.chatgpt_model,
+            messages=messages, 
             temperature=overrides.get("temperature") or 0.7, 
             max_tokens=1024, 
-            n=1, 
-            stop=["<|im_end|>", "<|im_start|>"])
+            n=1)
+        
+        chatContent = chatCompletion.choices[0].message.content
```
pamelafox, line 106:
```suggestion
        chat_content = chatCompletion.choices[0].message.content
```

app/backend/approaches/chatreadretrieveread.py
```diff
@@ -92,23 +92,20 @@ def run(self, history: Sequence[dict[str, str]], overrides: dict[str, Any]) -> A
...
 
-        return {"data_points": results, "answer": completion.choices[0].text, "thoughts": f"Searched for:<br>{q}<br><br>Prompt:<br>" + prompt.replace('\n', '<br>')}
+        return {"data_points": results, "answer": chatContent, "thoughts": f"Searched for:<br>{q}<br><br>Prompt:<br>" + prompt.replace('\n', '<br>')}
```
pamelafox, line 108:
```suggestion
        return {"data_points": results, "answer": chat_content, "thoughts": f"Searched for:<br>{q}<br><br>Prompt:<br>" + prompt.replace('\n', '<br>')}
```

pamelafox:
Excited for this change! Nice to not have to use the im_start/im_end anymore.

I left some comments, mostly around clarity around how token counting works, but no blockers.
 
I also noted some style issues around variable names. There's no linting/precommit setup for this repo yet, but until then, it'd be great if you could enable linting in your editor. The Python team's samples typically use ruff for linting, black for whitespace formatting, and isort for import sorting. The Pylance extension catches a lot of issues as well, however.

app/backend/approaches/chatreadretrieveread.py
srbalakr: Retained from previous code. Logic not to exceed 4096 tokens.

app/backend/approaches/chatreadretrieveread.py
srbalakr: Nice catch. there was a bug. I resolved it. Idea is to count token per conversation and not looping entire array every time.

app/backend/approaches/chatreadretrieveread.py
srbalakr: thats right

app/backend/approaches/chatreadretrieveread.py
srbalakr: its a single conversation chunk.I will leave as it is now.

app/backend/approaches/chatreadretrieveread.py
```diff
@@ -92,23 +92,20 @@ def run(self, history: Sequence[dict[str, str]], overrides: dict[str, Any]) -> A
...
             temperature=overrides.get("temperature") or 0.7, 
             max_tokens=1024, 
-            n=1, 
-            stop=["<|im_end|>", "<|im_start|>"])
+            n=1)
+        
+        chat_content = chat_completion.choices[0].message.content
 
-        return {"data_points": results, "answer": completion.choices[0].text, "thoughts": f"Searched for:<br>{q}<br><br>Prompt:<br>" + prompt.replace('\n', '<br>')}
+        return {"data_points": results, "answer": chat_content, "thoughts": f"Searched for:<br>{q}<br><br>Prompt:<br>" + prompt.replace('\n', '<br>')}
```
srbalakr, line 108: Should return conversatoins. Let me see if I can fix in this PR or put a seperate one since it might impact UX

app/backend/approaches/chatreadretrieveread.py
```diff
@@ -117,3 +114,70 @@ def get_chat_history_as_text(self, history: Sequence[dict[str, str]], include_la
...
+        Keep track of token count for each conversation
+        If token count exceeds limit, break
+        '''
+        for h in reversed(history[:-1]):
+            if h.get("bot"):
+                messages.insert(1, {"role": self.ASSISTANT, "content" : h.get("bot")})
+                token_count += self.num_tokens_from_messages(messages[1], self.chatgpt_model)
+            messages.insert(1, {"role": self.USER, "content" : h.get("user")})
+            token_count += self.num_tokens_from_messages(messages[1], self.chatgpt_model)
+            if token_count > approx_max_tokens*4:
```
chuwik, line 150: Note for future work re: supporting different models, since each model has a different token limit we'll need a way to handle different ones (e.g. 4096 for gpt3.5, 8192 for gpt-4).

app/backend/approaches/chatreadretrieveread.py
```diff
@@ -117,3 +114,70 @@ def get_chat_history_as_text(self, history: Sequence[dict[str, str]], include_la
...
+                break
+        return messages
+    
+    '''
+    Source: https://github.com/openai/openai-cookbook/blob/main/examples/How_to_count_tokens_with_tiktoken.ipynb
+    Adapted: https://learn.microsoft.com/en-us/azure/cognitive-services/openai/how-to/chatgpt?pivots=programming-language-chat-completions#managing-conversations
+
+    Method takes in a single conversation and calculate prompt tokens
```
chuwik, line 158: I found the function comments a bit hard to follow, could you please tweak them so it explains the function's purpose more clearly/simply?

app/backend/approaches/chatreadretrieveread.py
chuwik: +1, please ensure UX still works.

app/backend/approaches/chatreadretrieveread.py
```diff
@@ -117,3 +114,70 @@ def get_chat_history_as_text(self, history: Sequence[dict[str, str]], include_la
...
+            if h.get("bot"):
+                messages.insert(1, {"role": self.ASSISTANT, "content" : h.get("bot")})
+                token_count += self.num_tokens_from_messages(messages[1], self.chatgpt_model)
+            messages.insert(1, {"role": self.USER, "content" : h.get("user")})
+            token_count += self.num_tokens_from_messages(messages[1], self.chatgpt_model)
+            if token_count > approx_max_tokens*4:
+                break
+        return messages
+    
+    '''
```
pamelafox, line 154: Please move the docstring beneath the function signature, per https://peps.python.org/pep-0257/

app/backend/approaches/chatreadretrieveread.py
pamelafox: Perhaps this?


"""
Returns the approximate token count for a single message,
based on token encodings used for the given model.

Original: https://github.com/openai/openai-cookbook/blob/main/examples/How_to_count_tokens_with_tiktoken.ipynb
Adapted: https://learn.microsoft.com/azure/cognitive-services/openai/how-to/chatgpt?pivots=programming-language-chat-completions#managing-conversations

>>> num_tokens_from_message({"role":"assistant", "content":"how can I assist you?"}, "gpt-35-turbo")
127 (Or whatever the count is)
"""

The last two lines follow doctest format (https://docs.python.org/3/library/doctest.html)

app/backend/approaches/chatreadretrieveread.py
```diff
@@ -117,3 +114,70 @@ def get_chat_history_as_text(self, history: Sequence[dict[str, str]], include_la
...
+
+    message = {"role":"assistant", "content":"how can I assist you?"}
+    '''
+    def num_tokens_from_messages(self, message: any, model: str):
+        encoding = tiktoken.encoding_for_model(self.get_oai_chatmodel_tiktok(model))
+        num_tokens = 0
+        num_tokens += 2  # every message follows {role/name}\n{content}\n
+        for key, value in message.items():
+            num_tokens += len(encoding.encode(value))
+            if key == "name":  # if there's a name, the role is omitted
```
pamelafox, line 175: Do we get responses back that include a "name" key? We don't seem to send "name", from what I can tell.

app/backend/approaches/chatreadretrieveread.py
```diff
@@ -117,3 +114,70 @@ def get_chat_history_as_text(self, history: Sequence[dict[str, str]], include_la
...
+                num_tokens += -1  # role is always required and always 1 token
+        return num_tokens
+
+    def get_oai_chatmodel_tiktok(self, aoaimodel: str):
```
pamelafox, line 179: Design note: Both these methods (num_tokens_from_messages and get_oai_chatmodel_tiktok) are written as instance methods that take a self. However, they don't internally refer to self, and they instead accept the model as a parameter. I think they should either refer to self and use self.chatgpt_model, or they should be a @staticmethod. 

This design might naturally change when you change the other approaches (you might put this in the base class or a separate module), so you could wait to change it.

app/backend/approaches/chatreadretrieveread.py
```diff
@@ -117,3 +114,70 @@ def get_chat_history_as_text(self, history: Sequence[dict[str, str]], include_la
...
+
+    Keys role and content are accounted seperately.
+
+    Values of content are encoded by model type and calculated the length.
+    
+    This gives close proximity of token length measurement used in gpt models
+
+    message = {"role":"assistant", "content":"how can I assist you?"}
+    '''
+    def num_tokens_from_messages(self, message: any, model: str):
```
pamelafox, line 169: This seems to take a single message now, so it should be `num_tokens_from_message`

Can message be typed as a `dict`? That's what it appears to be in your docstring.

app/backend/approaches/chatreadretrieveread.py
pamelafox: Can tiktoken give us the max count as well? If so, there could be another function to return that, so that we can avoid the hand-wavey "*4" multiplier here.

app/backend/approaches/chatreadretrieveread.py
srbalakr: This need more changes. I will carry out in subsequent PR's

app/backend/approaches/chatreadretrieveread.py
srbalakr: Updated.

app/backend/approaches/chatreadretrieveread.py
srbalakr: Ack, will refactor them in upcoming PR's

app/backend/approaches/chatreadretrieveread.py
srbalakr: Good catch. Fixed.

app/backend/approaches/chatreadretrieveread.py
srbalakr: Good catch. removed.

chuwik (approved):

app/backend/approaches/chatreadretrieveread.py
```diff
@@ -117,3 +114,66 @@ def get_chat_history_as_text(self, history: Sequence[dict[str, str]], include_la
             if len(history_text) > approx_max_tokens*4:
                 break    
         return history_text
+    
+    # Generate messages needed for chat Completion api
```
pamelafox, line 118: Please move comment inside function, using standard docstring formatting, and capitalization.

"""Generate messages needed for Chat Completion API"""

app/backend/approaches/chatreadretrieveread.py
```diff
@@ -117,3 +114,66 @@ def get_chat_history_as_text(self, history: Sequence[dict[str, str]], include_la
...
+    def num_tokens_from_messages(self, message: dict[str,str], model: str):
+        """
+        Calculate the number of tokens required to encode a message.
+        Args:
+            message (any): The message to encode, represented as a dictionary.
+            model (str): The name of the model to use for encoding.
+        Returns:
+            int: The total number of tokens required to encode the message.
+        Example:
+            message = {'role': 'user', 'name': 'John', 'content': 'Hello, how are you?'}
```
pamelafox, line 163: You removed 'name' from later in the logic, but added to the example. So do we expect to use the 'name' key in these conversations or not? It seems like we won't, and I didnt see a mention of name in the linked learn.com article.

app/backend/approaches/chatreadretrieveread.py
```diff
@@ -117,3 +114,66 @@ def get_chat_history_as_text(self, history: Sequence[dict[str, str]], include_la
...
+            model = 'gpt-3.5-turbo'
+            num_tokens_from_messages(message, model)
+            output: 11
+        """
+        encoding = tiktoken.encoding_for_model(self.get_oai_chatmodel_tiktok(model))
+        num_tokens = 0
+        num_tokens += 2  # every message follows {role/name}\n{content}\n
```
pamelafox, line 170: The 2 is for the keys, right?

If so, this inline comment seems clearer?

```suggestion
        num_tokens += 2  # For "role" and "content" keys
```

If its not for the keys, then I'm confused.

app/backend/approaches/chatreadretrieveread.py
```diff
@@ -117,3 +114,66 @@ def get_chat_history_as_text(self, history: Sequence[dict[str, str]], include_la
...
+            system_message = self.system_message_chat_conversation.format(injected_prompt="", follow_up_questions_prompt=follow_up_questions_prompt)
+        elif prompt_override.startswith(">>>"):
+            system_message = self.system_message_chat_conversation.format(injected_prompt=prompt_override[3:] + "\n", follow_up_questions_prompt=follow_up_questions_prompt)
+        else:
+            system_message = prompt_override.format(follow_up_questions_prompt=follow_up_questions_prompt)
+
+        messages.append({"role":self.SYSTEM, "content": system_message})
+        token_count += self.num_tokens_from_messages(messages[-1], self.chatgpt_model)
+        
+        #latest conversation
```
pamelafox, line 132:
```suggestion
        # latest conversation
```

app/backend/approaches/chatreadretrieveread.py
srbalakr: Right, In our sample we dont. I have removed it.

pamelafox (approved):

pamelafox:
Thanks for addressing my feedback! The code is clearer to follow now.

Issue #388, opened 2023-07-06
 Creating account deployment is not supported by the model 'text-davinci-003'


damilareisaac:
```
ERROR: deployment failed: failing invoking action 'provision', error deploying infrastructure: deploying to subscription:
Deployment Error Details:
InvalidTemplateDeployment: The template deployment 'openai' is not valid according to the validation procedure. The tracking id is '198cc047-82d3-482a-9e67-4083cebb5cbd'. See inner errors for details.
DeploymentModelNotSupported: Creating account deployment is not supported by the model 'text-davinci-003'. This is usually because there are better models available for similar functionality.
```

I got the above error while trying to redeploy my application. I have not changed anything since the last deployment; I only add new documents to the data and want to redeploy. 

I try changing to GPT-4, but I did not get the error; got an error relating to the versioning. 

Is anyone else seeing the error? What is the workaround for this?

pamelafox:
This is a new issue, starting today, likely related to the OpenAI deprecation announcements. I'm testing to see if gpt3-turbo can be used instead.

pamelafox:
If all you changed was the code, then you can just run `azd deploy` to deploy just the code.

You don't actually need to re-deploy after adding docs to the data, since the index is updated instantly, independent of the app code.

pamelafox:
Another workaround for folks who already have the davinci model deployed:

1. Go into infra/main.bicep
2. Delete the first deployment in the deployments property:

```
      {
        name: gptDeploymentName
        model: {
          format: 'OpenAI'
          name: gptModelName
          version: '1'
        }
        sku: {
          name: 'Standard'
          capacity: gptDeploymentCapacity
        }
      }
```

phosseini:
> Another workaround for folks who already have the davinci model deployed:
> 
> 1. Go into infra/main.bicep
> 2. Delete the first deployment in the deployments property:
> 
> ```
>       {
>         name: gptDeploymentName
>         model: {
>           format: 'OpenAI'
>           name: gptModelName
>           version: '1'
>         }
>         sku: {
>           name: 'Standard'
>           capacity: gptDeploymentCapacity
>         }
>       }
> ```

This worked for me

sarahsofia93:
Has anyone succesfully deployed using another model? When I exchange text-davinci-003 for gpt-35-turbo in the main.bicep file, I get an error:

[...]

PR #389, opened 2023-07-06
Change davinci to gpt35-turbo

pamelafox:
## Purpose

Today, Azure disabled new deployments of text-davinci-003, so deployments will fail with an error:

"DeploymentModelNotSupported: Creating account deployment is not supported by the model 'text-davinci-003'. This is usually because there are better models available for the similar functionality."

This change does the minimal change to get deploys to work, swapping text-davinci-003 with gpt-35-turbo. Unfortunately, it really does not do that well on generating search queries, at least with our current prompt, so the chat interface isn't working well. Perhaps we should default to the "Chat" tab until it's worked out.

```
[X] Yes
[ ] No
```

## Pull Request Type

```
[X] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test

* Get the code
* Run azd env new, azd up
* See it deploy successfully

### DIFF
--- a/infra/main.bicep
+++ b/infra/main.bicep
@@ -39,7 +39,7 @@ param formRecognizerSkuName string = 'S0'
 
 param gptDeploymentName string // Set in main.parameters.json
 param gptDeploymentCapacity int = 30
-param gptModelName string = 'text-davinci-003'
+param gptModelName string = 'gpt-35-turbo'
 param chatGptDeploymentName string // Set in main.parameters.json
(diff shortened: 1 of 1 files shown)


pamelafox:
Another approach: Let developers specify if they have an existing davinci model, use it if they have it, and otherwise use the chatgpt model for everything. /CC @vhvb1989

tonybaloney (approved):

tonybaloney:
Fixes #388

PR #390, opened 2023-07-07
Use location picker for openAI

pamelafox:
## Purpose

This PR uses the location picker functionality for *just* the Open AI resource group location.
That allows us to constrain the locations to just the supported ones, so that folks dont find out after a long deploy about a bad location selection.
This also makes it easy for folks to pick a different location for OpenAI than their main location, which came up in a recent issue.

Here's what the picker looks like - it comes up after packaging but before provision:

Screenshot 2023-07-06 at 4 56 00 PM

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[ ] Bugfix
[X] Feature
[ ] Other... Please describe:
```

## How to Test

* Run azd up

### DIFF
--- a/infra/main.bicep
+++ b/infra/main.bicep
@@ -27,7 +27,14 @@ param storageContainerName string = 'content'
 
 param openAiServiceName string = ''
 param openAiResourceGroupName string = ''
-param openAiResourceGroupLocation string = location
+@description('Location for the OpenAI resource group')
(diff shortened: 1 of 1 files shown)


tonybaloney (approved):
This worked for me in a fresh deployment picking the main region as one that doesn't support Azure Open AI (Australia East) and the OAI model in US East which does. Tested the file uploads and the chat interface

screenshot 2023-07-07 at 15 43 58

PR #397, opened 2023-07-08
Frontend basic code splitting

chuwik:
## Purpose

* This PR introduces basic code splitting in the frontend:
  * Updated react-router-dom hash router to 6.4 API
  * Lazy routing to 'q&a' and 'not found' pages. 
    * 'Layout' and 'Chat' aren't lazy loaded because that's our initial page.
  * Configured rollup manual chunking in vite. _This also gets rid of the 'chunks over 500kb' vite warning_.
    * @fluentui/react-icons goes to its own chunk ('fluentui-icons')
    * @fluentui/react goes to its own chunk ('fluentui-react')
    * Other node_modules dependencies go to a 'vendor' chunk

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name
```

* Test the code

```
cd app
start.ps1 or start.sh
```

## What to Check
* All frontend pages load as expected.

### DIFF
--- a/app/frontend/src/index.tsx
+++ b/app/frontend/src/index.tsx
@@ -1,33 +1,38 @@
 import React from "react";
 import ReactDOM from "react-dom/client";
-import { HashRouter, Routes, Route } from "react-router-dom";
+import { createHashRouter, RouterProvider } from "react-router-dom";
 import { initializeIcons } from "@fluentui/react";
 
 import "./index.css";
 
 import Layout from "./pages/layout/Layout";
-import NoPage from "./pages/NoPage";
-import OneShot from "./pages/oneshot/OneShot";
 import Chat from "./pages/chat/Chat";
 
 initializeIcons();
 
-export default function App() {
-    return (
-        <HashRouter>
-            <Routes>
-                <Route path="/" element={<Layout />}>
-                    <Route index element={<Chat />} />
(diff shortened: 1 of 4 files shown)


app/frontend/src/index.tsx
```diff
@@ -1,33 +1,38 @@
...
-                <Route path="/" element={<Layout />}>
-                    <Route index element={<Chat />} />
-                    <Route path="qa" element={<OneShot />} />
-                    <Route path="*" element={<NoPage />} />
-                </Route>
-            </Routes>
-        </HashRouter>
-    );
-}
+const router = createHashRouter([
```
pamelafox, line 13: Is it possible to use BrowserRouter instead? (Using the HTML history / pushState APIs). It seems generally preferable over createHashRouter, but maybe that's not compatible with how this app is setup, I haven't tried.

pamelafox (approved):
I left a question about the BrowserRouter vs HashRouter decision, looks reasonable otherwise.

app/frontend/src/index.tsx
chuwik: Going to leave hash routing, we discussed switching it at the start but decided to keep it.

PR #411, opened 2023-07-11
[UX] Update 'thought process' to include conversations

srbalakr:
## Purpose
Update thought process tab to include system prompt and conversations from chat/completion api

image

```
[ ] Yes
[x ] No
```

## Pull Request Type

```
[ ] Bugfix
[ x] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
```

## What to Check
* ...

### DIFF
--- a/app/backend/approaches/chatreadretrieveread.py
+++ b/app/backend/approaches/chatreadretrieveread.py
@@ -105,7 +105,11 @@ Search query:
         
         chat_content = chat_completion.choices[0].message.content
 
-        return {"data_points": results, "answer": chat_content, "thoughts": f"Searched for:<br>{q}<br><br>Prompt:<br>" + prompt.replace('\n', '<br>')}
+        msg_to_display = ''
+        for msg in messages:
+            msg_to_display += str(msg) + '\n\n'
+
+        return {"data_points": results, "answer": chat_content, "thoughts": f"Searched for:<br>{q}<br><br>Conversations:<br>" + msg_to_display.replace('\n', '<br>')}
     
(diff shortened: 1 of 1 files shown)


valangar (approved):

chuwik (approved):

app/backend/approaches/chatreadretrieveread.py
```diff
@@ -105,7 +105,11 @@ def run(self, history: Sequence[dict[str, str]], overrides: dict[str, Any]) -> A
         
         chat_content = chat_completion.choices[0].message.content
 
-        return {"data_points": results, "answer": chat_content, "thoughts": f"Searched for:<br>{q}<br><br>Prompt:<br>" + prompt.replace('\n', '<br>')}
+        msg_to_display = ''
+        for msg in messages:
+            msg_to_display += str(msg) + '\n\n'
```
pamelafox, line 110: There's a one-liner to do that in Python, if you'd like:

```
msg_to_display = '\n\n'.join(str(message) for message in messages)
```

That uses a tuple comprehension and then joins the tuple.

I see you later replace \n with br, I assume you do that later instead of using br here as you're also trying to replace \n inside the content keys?

pamelafox: Updated above to use a list comprehension, more performant than what I previously had.

srbalakr: Thats right, and updated with list comprehensions

pamelafox (approved):

PR #418, opened 2023-07-12
Handle non-ascii filenames

### LINKED ISSUE #400: Invalid document key: '_________20230101_txt-1'. Keys cannot start with a leading underscore.
If you do "pwsh . \Јscripts\prepdocs.ps1" will change only the Japanese part of the filename in the DATA to underscores.
What is used is a text file.

pamelafox:
## Purpose

Fixes #400 

This fixes issues with filenames that start with non-ASCII characters. This also adds a basic pytest. I didn't put CI/CD for the test in this workflow yet, since @tonybaloney has a PR with that.

I also tested locally with a filename with all japanese characters, and it seems to work well:

Screenshot 2023-07-12 at 10 22 26 AM

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[X] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test

* python3 -m pytest scripts/test_prepdocs.py
* Manually: Add a file in data called ファイル名.pdf, run prepdocs.sh, then answer a question from that data and click the citations.


scripts/prepdocs.py
```diff
@@ -1,57 +1,25 @@
...
 from azure.core.credentials import AzureKeyCredential
-from azure.storage.blob import BlobServiceClient
+from azure.identity import AzureDeveloperCliCredential
+from azure.search.documents import SearchClient
 from azure.search.documents.indexes import SearchIndexClient
 from azure.search.documents.indexes.models import *
-from azure.search.documents import SearchClient
-from azure.ai.formrecognizer import DocumentAnalysisClient
+from azure.storage.blob import BlobServiceClient
+from pypdf import PdfReader, PdfWriter
```
pamelafox, line 17: I sorted the imports with isort since I was adding another import anyway.

scripts/prepdocs.py
```diff
@@ -1,57 +1,25 @@
...
 
 MAX_SECTION_LENGTH = 1000
 SENTENCE_SEARCH_LIMIT = 100
 SECTION_OVERLAP = 100
 
-parser = argparse.ArgumentParser(
```
pamelafox, line 21: I moved all this into if __name__ == "__main__" below, no changes otherwise.

scripts/prepdocs.py
```diff
@@ -220,10 +188,16 @@ def find_page(offset):
     if start + SECTION_OVERLAP < end:
         yield (all_text[start:end], find_page(start))
 
+def filename_to_id(filename):
```
pamelafox, line 191: The new function!

scripts/prepdocs.py
```diff
@@ -291,25 +265,64 @@ def remove_from_index(filename):
...
-        else:
-            if not args.skipblobs:
-                upload_blobs(filename)
-            page_map = get_document_text(filename)
-            sections = create_sections(os.path.basename(filename), page_map)
-            index_sections(os.path.basename(filename), sections)
+
+if __name__ == "__main__":
+
+    parser = argparse.ArgumentParser(
```
pamelafox, line 271: Moved from above with no changes

requirements-dev.txt
```diff
@@ -0,0 +1,3 @@
+-r app/backend/requirements.txt
```
chuwik, line 1: When should requirements-dev.txt be run? Should we update instructions/scripts?

chuwik (approved):

requirements-dev.txt
pamelafox: Only if you're running the tests. As mentioned in the PR, I didn't add a workflow file in this PR since there's one in Anthony's PR, but I'm also happy to do it in this PR as well. Right now these tests are opt-in.

I can also add a README about running tests, but it doesn't feel that useful for the average developer until we have much higher test coverage.

.github/workflows/python-test.yaml
```diff
@@ -0,0 +1,30 @@
...
+        - name: Setup python
+          uses: actions/setup-python@v2
+          with:
+            python-version: ${{ matrix.python_version }}
+            architecture: x64
+        - name: Install dependencies
+          run: |
+            python -m pip install --upgrade pip
+            pip install -r requirements-dev.txt
+        - name: Check formatting with black
```
chuwik, line 29: Please rename, looks like this is running the tests not checking formatting

chuwik (approved):

.github/workflows/python-test.yaml
pamelafox: Renamed, thanks!

chuwik (approved):

PR #419, opened 2023-07-12
Migration Completion api to chat completion api

srbalakr:
## Purpose
This PR address
1. Port completion api to chat completion api
2. Add helper function to track model tokens
3. Remove boilerplate codes. Make code clean
   3.1 Solve model token limit. 
   3.2 Clean up token estimation code

Breaking changes?

```
[x ] Yes
[ ] No
```

## Pull Request Type

```
[ ] Bugfix
[ x] Feature
[x ] Code style update (formatting, local variables)
[ ] Other... Please describe:
```


app/backend/approaches/chatreadretrieveread.py
```diff
@@ -30,26 +30,26 @@ class ChatReadRetrieveReadApproach(Approach):
     Try not to repeat questions that have already been asked.
     Only generate questions and do not generate any text before or after the questions, such as 'Next Questions'"""
 
-    query_prompt_template = """Below is a history of the conversation so far, and a new question asked by the user that needs to be answered by searching in a knowledge base about employee healthcare plans and the employee handbook.
-    Generate a search query based on the conversation and the new question. 
+    query_prompt_template = """Assistant is intelligent bot that helps to generate a search query.
```
srbalakr, line 33: @pamelafox  @chuwik  I tested for sometime, search query generation seems to be in parity with new template and fewer shots. It would be helpful if you play with my branch in local to see if quality is stable.

app/backend/approaches/chatreadretrieveread.py
```diff
@@ -30,26 +30,26 @@ class ChatReadRetrieveReadApproach(Approach):
...
-Search query:
+Search Query:
 """
-
-    def __init__(self, search_client: SearchClient, chatgpt_deployment: str, chatgpt_model: str, gpt_deployment: str, sourcepage_field: str, content_field: str):
+    query_prompt_few_shots = [
+        {'role' : USER, 'content' : 'What is included in my Northwind Health Plus plan that is not in standard?' },
+        {'role' : ASSISTANT, 'content' : 'Northwind Health Plus plan benefits comparison standard' },
+        {'role' : USER, 'content' : 'does my plan cover eye exam' },
+        {'role' : ASSISTANT, 'content' : 'Northwind Health Plus eye exam coverage' }
```
chuwik, line 47: The actual responses we get when asking these for real is more complex, are these 'few shot' examples enough to make the model produce good answers? Have you checked how they compare to using the version without your changes?

app/backend/approaches/chatreadretrieveread.py
```diff
@@ -90,94 +92,58 @@ def run(self, history: Sequence[dict[str, str]], overrides: dict[str, Any]) -> A
...
-        Keep track of token count for each conversation
-        If token count exceeds limit, break
-        '''
+    def get_messages_from_history(self, system_prompt: str, modelid: str, history: Sequence[dict[str, str]], user_conv: str, few_shots = []) -> []:
+        message_builder = MessageBuilder(system_prompt, modelid)
+
+        #add shots
+        for shot in few_shots:
+            message_builder.append_message(shot.get('role'), shot.get('content'))
+
```
chuwik, line 135: nit: it'd be good to pass the code through some python styling/linting tool to standardize whitespace, blank lines, etc. In typescript we use prettier/eslint.

app/backend/approaches/chatreadretrieveread.py
```diff
@@ -90,94 +92,58 @@ def run(self, history: Sequence[dict[str, str]], overrides: dict[str, Any]) -> A
...
-        Enqueue in reverse order
-        if limit exceeds truncate old messages 
-        leaving system message behind
-        Keep track of token count for each conversation
-        If token count exceeds limit, break
-        '''
+    def get_messages_from_history(self, system_prompt: str, modelid: str, history: Sequence[dict[str, str]], user_conv: str, few_shots = []) -> []:
+        message_builder = MessageBuilder(system_prompt, modelid)
+
+        #add shots
```
chuwik, line 132: nit: expand the comment a tiny bit to explain what/why this is done.

app/backend/approaches/retrievethenread.py
```diff
@@ -13,39 +14,29 @@ class RetrieveThenReadApproach(Approach):
...
 "Answer the following question using only the data provided in the sources below. " + \
 "For tabular information return it as an html table. Do not return markdown format. "  + \
 "Each source has a name followed by colon and the actual information, always include the source name for each fact you use in the response. " + \
-"If you cannot answer using the sources below, say you don't know. " + \
-"""
+"If you cannot answer using the sources below, say you don't know. Use below example to answer"
 
-###
-Question: 'What is the deductible for the employee plan for a visit to Overlake in Bellevue?'
+    #shots/sample conversation
```
chuwik, line 25: nit: please expand this comment so it provides a tiny bit of context/explanation, or remove.

app/backend/app.py
```diff
@@ -12,6 +12,7 @@
 from approaches.readdecomposeask import ReadDecomposeAsk
 from approaches.chatreadretrieveread import ChatReadRetrieveReadApproach
 from azure.storage.blob import BlobServiceClient
+from core.chatgptproxy import ChatGptProxy
```
pamelafox, line 15:
```suggestion

from core.chatgptproxy import ChatGptProxy
```

There should be a line between third-party imports and local imports, to make it clear what comes from where. (The isort extension/package will do this for you)

app/backend/core/ichatgptproxy.py
```diff
@@ -0,0 +1,32 @@
+from abc import abstractmethod
+
+class IChatGptProxy:
```
pamelafox, line 3: Whats the reason for this abstract class? I see only one subclass of it currently.

app/backend/core/messagebuilder.py
```diff
@@ -0,0 +1,31 @@
...
+    self.message = [{'role' : 'system', 'content': system_content}]
+    self.model = chatgpt_model
+    self.token_count = num_tokens_from_messages(self.message[-1], self.model)
+
+  def append_message(self, role:str, content: str, index: int = 1):
+    self.message.insert(index, {'role' : role, 'content': content})
+    self.token_count += num_tokens_from_messages(self.message[index], self.model)
+    return
+
+  def token_length(self) -> int:
```
pamelafox, line 26: Why not just access token_count directly? Why the wrapper method?

app/backend/core/messagebuilder.py
```diff
@@ -0,0 +1,31 @@
...
+    """
+  def __init__(self, system_content: str, chatgpt_model: str):
+    self.message = [{'role' : 'system', 'content': system_content}]
+    self.model = chatgpt_model
+    self.token_count = num_tokens_from_messages(self.message[-1], self.model)
+
+  def append_message(self, role:str, content: str, index: int = 1):
+    self.message.insert(index, {'role' : role, 'content': content})
+    self.token_count += num_tokens_from_messages(self.message[index], self.model)
+    return
```
pamelafox, line 24: Unnecessary return, return None is default

app/backend/core/messagebuilder.py
```diff
@@ -0,0 +1,31 @@
...
+        model (str): The name of the ChatGPT model.
+        token_count (int): The total number of tokens in the conversation.
+    Methods:
+        __init__(self, system_content: str, chatgpt_model: str): Initializes the MessageBuilder instance.
+        append_message(self, role: str, content: str, index: int = 1): Appends a new message to the conversation.
+        token_length(self) -> int: Returns the total token count in the conversation.
+        to_messages(self) -> list: Returns the list of messages in the conversation.
+    """
+  def __init__(self, system_content: str, chatgpt_model: str):
+    self.message = [{'role' : 'system', 'content': system_content}]
```
pamelafox, line 17: Suggest calling it messages since its a list

app/backend/core/messagebuilder.py
```diff
@@ -0,0 +1,31 @@
...
+
+  def append_message(self, role:str, content: str, index: int = 1):
+    self.message.insert(index, {'role' : role, 'content': content})
+    self.token_count += num_tokens_from_messages(self.message[index], self.model)
+    return
+
+  def token_length(self) -> int:
+    return self.token_count
+
+  def to_messages(self):
```
pamelafox, line 29: Same question, wrapper method doesnt seem necessary

app/backend/core/modelhelper.py
```diff
@@ -0,0 +1,46 @@
...
+  "gpt-4" : 8100,
+  "gpt-4-32k" : 32000
+}
+
+AOAI_2_OAI = {
+  "gpt-35-turbo" : "gpt-3.5-turbo",
+  "gpt-35-turbo-16k" : "gpt-3.5-turbo-16k"
+}
+
+def get_token_limit(modelid: str) -> int:
```
pamelafox, line 17: modelid -> model_id for readability

app/backend/app.py
```diff
@@ -43,6 +44,9 @@
 openai_token = azure_credential.get_token("https://cognitiveservices.azure.com/.default")
 openai.api_key = openai_token.token
 
+# Create chatGptProxy
+chatgpt_proxy = ChatGptProxy(openai, AZURE_OPENAI_CHATGPT_DEPLOYMENT, AZURE_OPENAI_CHATGPT_MODEL, AZURE_OPENAI_GPT_DEPLOYMENT)
```
pamelafox, line 48: Hm, It seems that you're passing the entire openai module here, but then also importing openai in the chatgptproxy file. I need to research more, but I'm not sure whether that's a good idea, it may lead to some strangeness.

Ideally, the OpenAI SDK would actually allow the construction of an OpenAI() object with the credentials, instead of us having to set the credentials as globals on the module. With the current OpenAI SDK design, we have to be very careful about where we set credentials and the way  we import openai.

app/backend/approaches/chatreadretrieveread.py
```diff
@@ -30,26 +30,26 @@ class ChatReadRetrieveReadApproach(Approach):
...
-{question}
-
-Search query:
+Search Query:
 """
-
-    def __init__(self, search_client: SearchClient, chatgpt_deployment: str, chatgpt_model: str, gpt_deployment: str, sourcepage_field: str, content_field: str):
+    query_prompt_few_shots = [
+        {'role' : USER, 'content' : 'What is included in my Northwind Health Plus plan that is not in standard?' },
+        {'role' : ASSISTANT, 'content' : 'Northwind Health Plus plan benefits comparison standard' },
```
pamelafox, line 45: If you're using the UI's sample questions for the few-shot, then you should change the sample question in the UI (as it seems a bit too easy for the bot!). Alternatively, you could come up with different example questions.

app/backend/approaches/chatreadretrieveread.py
srbalakr: This was bit tricky. For the very first question, it adds '+' irrespective of system prompt 6/10 times. In order to mimic the response we want, i need to add a actual response in the shots. It helps for the 1st question and for the rest of the conversation, it follow previous complex responses as it is (similar to previous approach).

app/backend/core/ichatgptproxy.py
srbalakr: The idea was to allow developers to create their own proxy based on open ai models and different model versions. Also decouple approaches from model knowledge. Turns out it adds more complexity for other approaches. I have removed the changes and kept it simple as before.

app/backend/core/messagebuilder.py
srbalakr: Comes from my previous experience with strong type languages. removed them :)

app/backend/app.py
srbalakr: Thanks Pamlea. I shared my thoughts for the changes in above comments. I have removed the proxy and kept implementation simple as it was.

app/backend/approaches/chatreadretrieveread.py
srbalakr: Thanks for the catch. I have revised the shots .

app/backend/core/messagebuilder.py
pamelafox: No worries, that's what I figured. Python generally doesn't use getters/setters, since attributes are public by default. If you really want to protect an attribute, you can put a double underscore in front, and Python will make it harder (but not impossible) to access from outside the object. You can then use @property to make a method that looks like an attribute to access a dynamically calculated property.

In this case, I think it's fine to just have them be publicly accessible attributes (no underscores).

app/backend/approaches/chatreadretrieveread.py
```diff
@@ -30,26 +30,26 @@ class ChatReadRetrieveReadApproach(Approach):
     Try not to repeat questions that have already been asked.
     Only generate questions and do not generate any text before or after the questions, such as 'Next Questions'"""
 
-    query_prompt_template = """Below is a history of the conversation so far, and a new question asked by the user that needs to be answered by searching in a knowledge base about employee healthcare plans and the employee handbook.
-    Generate a search query based on the conversation and the new question. 
+    query_prompt_template = """Assistant is intelligent bot that helps to generate a search query.
+    Below is a history of the conversation so far, and a new question asked by the user that needs to be answered by searching in a knowledge base about employee healthcare plans and the employee handbook.
+    Generate a SEARCH QUERY based on the previous chat history and the new question.
```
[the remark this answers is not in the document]
srbalakr: I came across few prompts where they insisted to have upper case instructions. But turns out in our case was the trick to add a trigger line before user question. Reverted prompt.

pamelafox (commented):
The overall code organization looks good now. I do have a concern that these two approaches are now *only* using the ChatGPT deployment variables. However, some customers may still have Davinci, so this code won't enable them to use Davinci. Plus, in the future, we're hoping to have the gpt35-instruct model. So I *think* we want to keep the distinction between chatgpt_deployment and gpt_deployment for now. Feel free to push back if you think we should consolidate on one model.

pamelafox:
@srbalakr Okay, got it, so I assume you'll be removing the extra deployment from the Bicep files in a subsequent PR.

chuwik (approved):

PR #219, opened 2023-05-18
Add support for vector search for retrieval

pablocastro:
## Purpose
This change introduces support for using vector search during retrieval, using Azure Cognitive Search vector search and Azure OpenAI embeddings. This includes changes to the data preparation and ingestion process (to vectorize data using the OpenAI embeddings API) and during response generation where we now offer the option to do hybrid (vector + text), pure vector, or pure text retrieval.

## Does this introduce a breaking change?
```
[x] Yes
[ ] No
```

*Breaking change in that existing deployment won't work without either manually adding the embedding field to the index and populating it, or deleting the search index and re-running the ingestion process (prepdocs.py or prepdocs.ps1).

## Pull Request Type
What kind of change does this Pull Request introduce?

```
[ ] Bugfix
[x ] Feature
[ ] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[ ] Documentation content changes
[ ] Other... Please describe:
```

## How to Test
* Deploy this updated version, re-creating the search index as appropriate
* Issue a chat or question request, the default is already set to "hybrid" and will use vectors

PR #424, opened 2023-07-14
Vector search

pablocastro:
## Purpose
Using the new vector search capabilities in Azure Cognitive Search, this change introduces options for pure vector search, hybrid search, and the existing keyword search for retrieval. It also supports combining vector/hybrid search with L2 reranking for even better relevance.

There's a new dropdown in the developer options that allows the user to choose between vectors, hybrid and text-based retrieval, and can be changed at any time.

```
[x] Yes
[ ] No
```

It's breaking because the indexes need to be updated to include a vector field with the embeddings of the content. The new version of prepdocs.py creates indexes that include a vector field and uses the OpenAI embeddings API to compute embeddings during ingestion.

## Pull Request Type
```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```


vikramhn:
Are there any adversely affects like performance or timeouts when trying to upload large volumes of docs into the index - as the below  introduces extra calls outside the indexing platform as up until now indexing was self contained. Hope it is usable for higher volume for example about 10k Docs 
_**because the indexes need to be updated to include a vector field with the embeddings of the content. The new version of prepdocs.py creates indexes that include a vector field and uses the OpenAI embeddings API to compute embeddings during ingestion**_

itmilos:
@vikramhn 

Uploading over 10,000 documents would necessitate a distinct procedure and substantial enhancements for the prepdocs scripts. I would advise against using this demo with such a large number of documents.

itmilos:
@pablocastro I think we might have to add missing params for prepdocs.sh - as stated here https://github.com/Azure-Samples/azure-search-openai-demo/pull/219#issuecomment-1555363280

itmilos:
bug - If following settings 
image

result

image

itmilos:
I believe we might encounter rate limit issues, particularly with certain files reaching the limit rapidly. Therefore, it would be beneficial to implement a throttling mechanism to regulate the flow. @pablocastro

pablocastro:
Good catch @itmilos, fixed.

Re; rate limiting, I've been running the loading code a few times and haven't seen rate limiting. I'll start as-is for simplicity (prepdocs has gotten complicated already), but we certainly consider adding rate-limiting code if this becomes an issue.

vikramhn:
> Good catch @itmilos, fixed.
> 
> Re; rate limiting, I've been running the loading code a few times and haven't seen rate limiting. I'll start as-is for simplicity (prepdocs has gotten complicated already), but we certainly consider adding rate-limiting code if this becomes an issue.

May be add a reference to this article 
https://learn.microsoft.com/en-us/samples/azure-samples/azure-search-power-skills/azure-open-ai-embeddings-generator/ 
if it makes sense so folks can consider this for large workloads/production use cases.

srbalakr (approved):

app/backend/approaches/chatreadretrieveread.py
```diff
@@ -107,7 +130,7 @@ def run(self, history: Sequence[dict[str, str]], overrides: dict[str, Any]) -> A
 
         msg_to_display = '\n\n'.join([str(message) for message in messages])
 
-        return {"data_points": results, "answer": chat_content, "thoughts": f"Searched for:<br>{q}<br><br>Conversations:<br>" + msg_to_display.replace('\n', '<br>')}
+        return {"data_points": results, "answer": chat_content, "thoughts": f"Searched for:<br>{query_text}<br><br>Conversations:<br>" + msg_to_display.replace('\n', '<br>')}
```
srbalakr, line 133: do we want to show the vectors generated in UX as part of thoughts ?

app/backend/app.py
```diff
@@ -31,7 +32,7 @@
 # just use 'az login' locally, and managed identity when deployed on Azure). If you need to use keys, use separate AzureKeyCredential instances with the 
 # keys for each service
 # If you encounter a blocking error during a DefaultAzureCredntial resolution, you can exclude the problematic credential by using a parameter (ex. exclude_shared_token_cache_credential=True)
-azure_credential = DefaultAzureCredential()
+azure_credential = DefaultAzureCredential(exclude_shared_token_cache_credential = True)
```
chuwik, line 35: Do you want to commit 'exclude_shared_token_cache_credential = True'?

chuwik (approved):

app/backend/approaches/chatreadretrieveread.py
```diff
@@ -107,17 +131,17 @@ def run(self, history: Sequence[dict[str, str]], overrides: dict[str, Any]) -> A
...
     
     def get_chat_history_as_text(self, history: Sequence[dict[str, str]], include_last_turn: bool=True, approx_max_tokens: int=1000) -> str:
         history_text = ""
         for h in reversed(history if include_last_turn else history[:-1]):
-            history_text = """<|im_start|>user""" + "\n" + h["user"] + "\n" + """<|im_end|>""" + "\n" + """<|im_start|>assistant""" + "\n" + (h.get("bot", "") + """<|im_end|>""" if h.get("bot") else "") + "\n" + history_text
+            history_text = """user:""" + "\n" + h["user"] + "\n\n" + ("""assistant:""" + "\n" + h.get("bot", "") if h.get("bot") else "") + "\n" + history_text
```
pamelafox, line 139: This has now changed in main, so I think you'll need to merge/rebase and possibly resolve conflicts.

app/backend/approaches/readretrieveread.py
```diff
@@ -44,29 +44,51 @@ class ReadRetrieveReadApproach(Approach):
...
+        self.embedding_deployment = embedding_deployment
         self.sourcepage_field = sourcepage_field
         self.content_field = content_field
 
-    def retrieve(self, q: str, overrides: dict[str, Any]) -> Any:
-        use_semantic_captions = True if overrides.get("semantic_captions") else False
+    def retrieve(self, query_text: str, overrides: dict[str, Any]) -> Any:
+        has_text = overrides.get("retrieval_mode") in ["text", "hybrid", None]
+        has_vector = overrides.get("retrieval_mode") in ["vectors", "hybrid", None]
+        use_semantic_captions = True if overrides.get("semantic_captions") and has_text else False
```
pamelafox, line 57: I notice a lot of repeated code across the approaches now. Would it make sense to move any of it into a method on the Approach class? (These 3 lines, for examples)

Might help developers to see what's actually different across the approaches.

scripts/prepdocs.py
```diff
@@ -6,6 +6,7 @@
 import os
 import re
 import time
+import openai
```
pamelafox, line 9: nit: external non-stdlib imports belong in the next set of imports. (If you run isort, it'll fix it up).

pamelafox:
High-level comment: Does the app/UI default to what you believe to be the most successful approach? I think many developers will go with the default and we want to set them up for success. Are there documents or videos we can point to to help understand more about the embeddings approach? I know you made some great videos for Microsoft Mechanics (which I watched in order to better understand this repo).

app/backend/app.py
pablocastro: I noticed sometimes systems get into a state where the legacy cache gets in the way. To save folks from a bunch of troubleshooting I'm leaving this set to true. If there's a more proper troubleshooting approach to this, e.g. how to locate and clear caches, etc., happy to update this.

app/backend/approaches/readretrieveread.py
pablocastro: The original goal of this demo was to be small, and each "approach" had a lot of redundancy because I thought people could take it, try the various approaches, and then pick one and delete the rest.

Clearly this codebase keeps growing, so yes, I some point we may want to consider refactoring to extract common parts. I'll keep this PR focused on introducing vector search, we can debate separately when we do a bit of refactoring.

pablocastro:
> High-level comment: Does the app/UI default to what you believe to be the most successful approach? I think many developers will go with the default and we want to set them up for success. Are there documents or videos we can point to to help understand more about the embeddings approach? I know you made some great videos for Microsoft Mechanics (which I watched in order to better understand this repo).

Good question @pamelafox. Yes, I think that the defaults for chat (using hybrid retrieval + L2 reranking) will yield the best results in general. Of course specific scenarios will vary, but this feels like a reasonable starting point. The reasoning is that hybrid fuses the best of text search and the best of vector search (we have lots of data supporting this), and that the diverse set of candidates produced by the 2 retrieval paths feed the L2 reranking with a rich input to rerank.

pamelafox (approved):

itmilos:
🚀

hanwesh:
@pablocastro my understanding is , if I have an existing app, i need to delete the index and run azd up?

hanwesh:
> @pablocastro my understanding is , if I have an existing app, i need to delete the index and run azd up?

The answer is yes as I just test it.

PR #447, opened 2023-07-21
Added a rigid filtering approach

tloren:
Added a rigid filtering approach for finding specific PHIS from the index 

We can turn this feature off from the FE with a checkbox


tloren:
Selected the wrong branch, apologies

Issue #272, opened 2023-06-05
Website dies randomly when asking questions


kikaragyozov:
I've deployed the project to azure as instructed by using `azd up`. I did not use any previous azure resources. Everything was made by the supplied scripts in the repository.

What's going on? Inspecting the Live stream logs of the backend application, I don't see anything special. But basically, the azure website stops responding at certain points in time, specifically after asking a question in chat. I'm not sure what's going on behind the scenes.

kikaragyozov:
> Hmm. Have you tried opening the Network console in the browser to see if an HTTP request is going through to the backend? A successful request looks like: <img alt="Screenshot 2023-06-05 at 3 03 20 PM" width="1095" src="https://user-images.githubusercontent.com/297042/243486690-6e95d1ba-090d-4e4c-9e71-bb182fae789e.png">

After inspecting the HTTP Server Errors, I noticed a few 504 Gateway Timeout errors and 500, but I can't find any information as to why they were produced.

Yes, many of the requests to the /chat endpoint end up with a HTTP Status code of 200 (OK). But there comes a time where either: 

* The website crashes down after asking a question.
* The response takes upto 3-5 minutes for a specific `/chat` request. While that's happening, it seems the whole site becomes unresponsive, as if python has a single thread serving the whole thing...

Could anyone provide more insights as to what's happening? Or any possible workarounds to this?

mbrenigjones:
Hi @kikaragyozov, After I deployed the template I saw somewhat similar issues: simultaneous requests took a long time to return, and then returned in sequence as if there was only one thread serving them. I didn't see the 500 or 504 errors you mention.

I configured diagnostic settings on the App-Service to send logs to a Log Analytics workspace [[1]](https://learn.microsoft.com/en-gb/azure/azure-monitor/essentials/diagnostic-settings?WT.mc_id=Portal-Microsoft_Azure_Monitoring&tabs=portal). In those logs I found that when the App-Service starts up gunicorn (the WSGI server that App-Service uses to serve Python apps) is using synchronous workers and only starting one worker thread:

![image](https://github.com/Azure-Samples/azure-search-openai-demo/assets/88146971/72dda855-f0c7-4401-9354-a77f4085070a)

[...]

PR #461, opened 2023-07-26
Tell Oryx to configure gunicorn workers dynamically

pamelafox:
## Purpose

This PR enables PYTHON_ENABLE_GUNICORN_MULTIWORKERS on App Service. When Oryx sees that, it will auto configure gunicorn to use a worker count based off the formula of CPU_COUNT * 2 + 1. This should hopefully resolve issues like #272.

Here's Oryx docs about the config option:
https://github.com/microsoft/Oryx/blob/5239b0416cdddcf5bf0aef6703eb339efa865d67/doc/runtimes/python.md?plain=1#L101

The other approach is for us to specify our own startup script that does the same thing (as I describe in https://github.com/Azure-Samples/azure-search-openai-demo/issues/272#issuecomment-1589303445), but for now, this is the quickest fix.

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[ ] Bugfix
[X] Feature
[ ] Other... Please describe:
```

## How to Test

* azd up
* Check Advanced Tools > Download as zip > default_docker.log, see that it spawned 3 workers
* Also SSH into App Service and check out /opt/startup/startup.sh, you'll see workers=3 in the command.

### DIFF
--- a/infra/core/host/appservice.bicep
+++ b/infra/core/host/appservice.bicep
@@ -70,6 +70,7 @@ resource appService 'Microsoft.Web/sites@2022-03-01' = {
         SCM_DO_BUILD_DURING_DEPLOYMENT: string(scmDoBuildDuringDeployment)
         ENABLE_ORYX_BUILD: string(enableOryxBuild)
       },
+      runtimeName == 'python' ? { PYTHON_ENABLE_GUNICORN_MULTIWORKERS: 'true'} : {},
       !empty(applicationInsightsName) ? { APPLICATIONINSIGHTS_CONNECTION_STRING: applicationInsights.properties.ConnectionString } : {},
       !empty(keyVaultName) ? { AZURE_KEY_VAULT_ENDPOINT: keyVault.properties.vaultUri } : {})
   }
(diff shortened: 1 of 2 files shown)


pamelafox:
cc @jongio We may want to put this setting by default in core/host/appservice.bicep - the only reason its not default on Oryx is backwards compatibility issues.

chip-davis:
Just chiming in that making this change fixed the issue I was having as mentioned in #272

chuwik:
> cc @jongio We may want to put this setting by default in core/host/appservice.bicep - the only reason its not default on Oryx is backwards compatibility issues.

+1 to making the change in core/host/appservice.bicep, feels like this should be specific to the app service config rather than in the main bicep file.

chuwik (approved):

vrajroutu:
@pamelafox  Can you also look in to this? Adjust the number of workers based on the number of CPU cores  

`app/start.sh`


`# Install gunicorn  
./backend_env/bin/pip install gunicorn  
  
# Adjust the number of workers based on the number of CPU cores  
workers=$(( $(nproc) * 2 ))  
  
# Start the backend using Gunicorn with multiple workers  
./backend_env/bin/gunicorn -w $workers -b 127.0.0.1:8000 app:app  
  
if [ $? -ne 0 ]; then  
    echo "Failed to start backend"  
    exit $?  
fi`

PR #469, opened 2023-07-27
Respect ENV values for resource group and index name

Patrick-Davis-MSFT:
## Purpose
This allows the bicep to respect the values ( AZURE_RESOURCE_GROUP & AZURE_SEARCH_SERVICE) manually entered in the .azure/.env file should they be required.

```
[ ] Yes
[X ] No
```

## Pull Request Type

```
[X ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

1. Run `azd up` on a new environment
1. Update the values AZURE_RESOURCE_GROUP and AZURE_SEARCH_SERVICE in the `.azure/.env` file. 
```
```

## What to Check
Verify that the following are valid.
* Verify that the full product was created appropriately after the first step
* Verify the resource group and new webapp were created appropriately. Verify the new index with the new name was created. 
* Verify a search index was created with the name entered in the value was respected

### DIFF
--- a/infra/abbreviations.json
+++ b/infra/abbreviations.json
@@ -131,6 +131,5 @@
   "webSitesAppService": "app-",
   "webSitesAppServiceEnvironment": "ase-",
   "webSitesFunctions": "func-",
-  "webStaticSites": "stapp-",
-  "searchIndexName":"gptkbindex"
+  "webStaticSites": "stapp-"
 }
--- a/infra/abbreviations.json
+++ b/infra/abbreviations.json
@@ -131,5 +131,6 @@
   "webSitesAppService": "app-",
   "webSitesAppServiceEnvironment": "ase-",
   "webSitesFunctions": "func-",
-  "webStaticSites": "stapp-"
+  "webStaticSites": "stapp-",
+  "searchIndexName":"gptkbindex"
 }
--- a/infra/main.bicep
+++ b/infra/main.bicep
@@ -18,7 +18,7 @@ param searchServiceResourceGroupName string = ''
 param searchServiceResourceGroupLocation string = location
 
 param searchServiceSkuName string = 'standard'
-param searchIndexName string = 'gptkbindex'
+param searchIndexName string = ''
 
 param storageAccountName string = ''
 param storageResourceGroupName string = ''
@@ -117,7 +117,7 @@ module backend 'core/host/appservice.bicep' = {
       AZURE_STORAGE_ACCOUNT: storage.outputs.name
       AZURE_STORAGE_CONTAINER: storageContainerName
       AZURE_OPENAI_SERVICE: openAi.outputs.name
-      AZURE_SEARCH_INDEX: searchIndexName
+      AZURE_SEARCH_INDEX: !empty(searchIndexName) ? searchIndexName : '${abbrs.searchIndexName}'
       AZURE_SEARCH_SERVICE: searchService.outputs.name
       AZURE_OPENAI_GPT_DEPLOYMENT: gptDeploymentName
       AZURE_OPENAI_CHATGPT_DEPLOYMENT: chatGptDeploymentName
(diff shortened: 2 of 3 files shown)


Patrick-Davis-MSFT:
@microsoft-github-policy-service agree [company="Microsoft"]

Patrick-Davis-MSFT:
@microsoft-github-policy-service agree company="Microsoft"

infra/main.bicep
```diff
@@ -18,7 +18,7 @@ param searchServiceResourceGroupName string = ''
 param searchServiceResourceGroupLocation string = location
 
 param searchServiceSkuName string = 'standard'
-param searchIndexName string = 'gptkbindex'
+param searchIndexName string = ''
```
pamelafox, line 21: I think actually the right approach here is what we did for the deployment names - don't set this to any value, like:

```
param gptDeploymentName string // Set in main.parameters.json
```

Then in main.parameters.json, do:

```
   "searchIndexName": {
      "value": "${AZURE_SEARCH_INDEX=gptkbindex}"
    },
```

I don't recall why right now, but I think this works better in more situations.

pamelafox (changes requested):
This is a helpful change, thanks! I know many folks want to reuse existing indices. I have a few suggestions, however.

infra/main.bicep
Patrick-Davis-MSFT: Updated to the requested pattern

infra/abbreviations.json
```diff
@@ -131,5 +131,6 @@
   "webSitesAppService": "app-",
   "webSitesAppServiceEnvironment": "ase-",
   "webSitesFunctions": "func-",
-  "webStaticSites": "stapp-"
+  "webStaticSites": "stapp-",
+  "searchIndexName":"gptkbindex"
```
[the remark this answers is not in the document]
Patrick-Davis-MSFT: The Abbreviations file seemed to be the place where default values were pulled from. This matches the other items (example Main.bicep line 92, the resource token is not needed here because the index is not required to be globally unique)

infra/main.bicep
```diff
@@ -117,7 +117,7 @@ module backend 'core/host/appservice.bicep' = {
       AZURE_STORAGE_ACCOUNT: storage.outputs.name
       AZURE_STORAGE_CONTAINER: storageContainerName
       AZURE_OPENAI_SERVICE: openAi.outputs.name
-      AZURE_SEARCH_INDEX: searchIndexName
+      AZURE_SEARCH_INDEX: !empty(searchIndexName) ? searchIndexName : '${abbrs.searchIndexName}'
```
pamelafox, line 120:
```suggestion
      AZURE_SEARCH_INDEX: searchIndexName
```

infra/main.parameters.json
```diff
@@ -29,6 +32,9 @@
     "formRecognizerSkuName": {
       "value": "S0"
     },
+    "searchIndexName": {
+      "value": "${AZURE_SEARCH_INDEX}"
```
pamelafox, line 36:
```suggestion
      "value": "${AZURE_SEARCH_INDEX=gptkbindex}"
```

infra/abbreviations.json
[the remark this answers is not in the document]
pamelafox: This abbreviations files came from the azd repository (https://github.com/Azure/azure-dev/blob/main/templates/common/infra/bicep/abbreviations.json) and stores default prefixes for created Azure resources, so that all azd templates create resources with similar naming scheme. The index name is more akin to the model deployment name, like "davinci", which aren't stored in here.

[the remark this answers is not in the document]
Patrick-Davis-MSFT: Understood. I am running a test on this now to test before I resolve the comment

Patrick-Davis-MSFT:
Thank you for your help. I have made the updates and adjustments suggested. Please let me know if you have any other comments.

pamelafox:
Looks good, have you tried deploying with the latest change?

Patrick-Davis-MSFT:
Yes I did... All seemed to work

pamelafox (approved):

pamelafox:
Great, merging. Thanks for the PR!

PR #486, opened 2023-08-01
Sample Locust loadtest

pamelafox:
## Purpose

WIP, loadtest for Locust framework.

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[ ] Bugfix
[X] Feature
[ ] Other... Please describe:
```

## How to Test

* pip install locust
* locust

### DIFF
--- a/locustfile.py
+++ b/locustfile.py
@@ -1,18 +1,25 @@
-from locust import HttpUser, task, between
+import time
+import random
+
+from locust import HttpUser, task, between, constant_throughput, constant_pacing
 
 
 class ChatUser(HttpUser):
-    wait_time = between(60,61)
-
-    def on_start(self):
-        self.client.get("/")
+    wait_time = between(5, 20)
 
     @task
     def ask_question(self):
+        self.client.get("/")
+        time.sleep(5)
         self.client.post("/chat", json={
             "history": [
                 {
-                "user": "What is included in my Northwind Health Plus plan that is not in standard?"
+                "user": random.choice([
+                    "What is included in my Northwind Health Plus plan that is not in standard?",
+                    "What does a Product Manager do?",
+                    "What happens in a performance review?",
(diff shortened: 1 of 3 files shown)


locustfile.py
```diff
@@ -0,0 +1,26 @@
...
+
+    def on_start(self):
+        self.client.get("/")
+
+    @task
+    def ask_question(self):
+        self.client.post("/chat", json={
+            "history": [
+                {
+                "user": "What is included in my Northwind Health Plus plan that is not in standard?"
```
tonybaloney, line 15: I'd suggest having a list of questions and using random.choice instead of the same question each time.

locustfile.py
```diff
@@ -0,0 +1,26 @@
+from locust import HttpUser, task, between
+
+
+class ChatUser(HttpUser):
+    wait_time = between(60,61)
+
+    def on_start(self):
```
tonybaloney, line 7: This should just be a task, otherwise there won't be a wait period between this `.get()` and the subsequence `.post()` in the next function

locustfile.py
```diff
@@ -0,0 +1,26 @@
...
+            ],
+            "approach": "rrr",
+            "overrides": {
+                "retrieval_mode": "hybrid",
+                "semantic_ranker": True,
+                "semantic_captions": False,
+                "top": 3,
+                "suggest_followup_questions": False
+            }
+            })
```
tonybaloney, line 26: Does the client make any requests after this?

tonybaloney (changes requested):

locustfile.py
pamelafox: I've added a follow-up request with more history. I also moved everything into a single task. Seemed like better behavior modeling. In future, one could also have an /ask task, but just checking /chat reveals a lot of issues.

tonybaloney (approved):

PR #495, opened 2023-08-02
Use an app factory pattern to enable app testing

pamelafox:
## Purpose

This PR changes app.py to use an app factory pattern (create_app) along with a Flask Blueprint to organize the code. This makes testing much easier and is a general best practice.

This PR is based off the previous PR that ruff-linted the files, so that needs to be merged first to isolate the changes. Then I'll turn the draft into an actual PR.

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test

* python3 -m pytest
* Run start.sh in app
* Run azd up for prod

### DIFF
--- a/.github/workflows/python-test.yaml
+++ b/.github/workflows/python-test.yaml
@@ -26,5 +26,7 @@ jobs:
           run: |
             python -m pip install --upgrade pip
             pip install -r requirements-dev.txt
+        - name: Lint with ruff
+          run: ruff .
         - name: Run Python tests
           run: python3 -m pytest
--- a/.github/workflows/python-test.yaml
+++ b/.github/workflows/python-test.yaml
@@ -14,7 +14,7 @@ jobs:
       fail-fast: false
       matrix:
         os: ["ubuntu-20.04"]
-        python_version: ["3.8", "3.9", "3.10", "3.11"]
+        python_version: ["3.9", "3.10", "3.11"]
     steps:
         - uses: actions/checkout@v3
         - name: Setup python
--- a/.github/workflows/python-test.yaml
+++ b/.github/workflows/python-test.yaml
@@ -22,6 +22,15 @@ jobs:
           with:
             python-version: ${{ matrix.python_version }}
             architecture: x64
+        - name: Setup node
+          uses: actions/setup-node@v2
+          with:
+            node-version: 18
+        - name: Build frontend
+          run: |
+            cd ./app/frontend
+            npm install
+            npm run build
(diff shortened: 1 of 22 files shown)


.github/workflows/python-test.yaml
```diff
@@ -14,14 +14,23 @@ jobs:
       fail-fast: false
       matrix:
         os: ["ubuntu-20.04"]
-        python_version: ["3.8", "3.9", "3.10", "3.11"]
+        python_version: ["3.9", "3.10", "3.11"]
```
pamelafox, line 17: Removing 3.8 since we don't currently support it (due to our typing syntax). We can add it back if its an issue for customers.

.github/workflows/python-test.yaml
```diff
@@ -14,14 +14,23 @@ jobs:
...
             architecture: x64
+        - name: Setup node
+          uses: actions/setup-node@v2
+          with:
+            node-version: 18
+        - name: Build frontend
+          run: |
+            cd ./app/frontend
+            npm install
+            npm run build
```
pamelafox, line 33: This must be built since the Flask tests try to hit up index.html

app/backend/app.py
```diff
@@ -5,85 +5,67 @@
...
+def index():
+    return bp.send_static_file("index.html")
+
+@bp.route("/favicon.ico")
+def favicon():
+    return bp.send_static_file("favicon.ico")
+
+@bp.route("/assets/<path:path>")
+def assets(path):
+    return send_from_directory("static/assets", path)
```
pamelafox, line 61: This is a security upgrade over the previous route that used send_static_file for arbitrary paths.

app/backend/app.py
```diff
@@ -95,13 +77,13 @@ def content_file(path):
     blob_file.seek(0)
     return send_file(blob_file, mimetype=mime_type, as_attachment=False, download_name=path)
 
-@app.route("/ask", methods=["POST"])
+@bp.route("/ask", methods=["POST"])
 def ask():
-    if not request.json:
-        return jsonify({"error": "request must be json"}), 400
+    if not request.is_json:
```
pamelafox, line 82: This was a bug, caught by the new tests. Calling request.json causes an immediate 415 HTTP error if theres no JSON passed in, thus it's now request.is_json.

tonybaloney:
Great improvement, this will cause a few merge conflicts with existing PRs, but that can't be avoided

PR #503, opened 2023-08-06
Port to Quart

pamelafox:
## Purpose

This PR ports the app to Quart, a Python ASGI web framework. 
Flask is a WSGI framework, which is synchronous. That means that when a request comes in, the server will wait for the request to be processed before it can handle another request. We have added support for multiple gunicorn workers/threads, but they can still easily be blocked by a long-running request. Each /chat request makes 4 API calls, which take from 0.2-7 seconds each.

Quart is designed with the same API as Flask, with a few changes needed for asynchronous support. Most of my changes are the addition of await/async. For the Azure SDK calls, I used the async versions of the clients (imported from `.aio`). For the OpenAI API calls, I used the async versions of the functions (`acreate` instead of `create`).

For deployment on App Service, I use gunicorn with the uvicorn worker.

```
[X] Yes
[ ] No
```

## Pull Request Type

```
[ ] Bugfix
[X] Feature
[ ] Other... Please describe:
```

## How to Test

* `app/start.sh` should start the app
* `azd up` should deploy the app
* `python3 -m pytest` should run the tests

### DIFF
--- a/app/backend/approaches/chatreadretrieveread.py
+++ b/app/backend/approaches/chatreadretrieveread.py
@@ -105,7 +105,7 @@ If you cannot generate a search query, return just the number 0.
         if overrides.get("semantic_ranker") and has_text:
             r = await self.search_client.search(query_text,
                                           filter=filter,
-                                          query_type=QueryType.SIMPLE,
+                                          query_type=QueryType.SEMANTIC,
                                           query_language="en-us",
                                           query_speller="lexicon",
                                           semantic_configuration_name="default",
--- a/app/backend/approaches/chatreadretrieveread.py
+++ b/app/backend/approaches/chatreadretrieveread.py
@@ -1,16 +1,16 @@
-from typing import Any, Sequence
+from typing import Any
 
 import openai
 from azure.search.documents.aio import SearchClient
 from azure.search.documents.models import QueryType
 
-from approaches.approach import Approach
+from approaches.approach import ChatApproach
 from core.messagebuilder import MessageBuilder
 from core.modelhelper import get_token_limit
 from text import nonewlines
 
 
-class ChatReadRetrieveReadApproach(Approach):
+class ChatReadRetrieveReadApproach(ChatApproach):
     # Chat roles
     SYSTEM = "system"
     USER = "user"
@@ -57,7 +57,7 @@ If you cannot generate a search query, return just the number 0.
         self.content_field = content_field
         self.chatgpt_token_limit = get_token_limit(chatgpt_model)
 
-    async def run(self, history: Sequence[dict[str, str]], overrides: dict[str, Any]) -> Any:
+    async def run(self, history: list[dict[str, str]], overrides: dict[str, Any]) -> Any:
         has_text = overrides.get("retrieval_mode") in ["text", "hybrid", None]
         has_vector = overrides.get("retrieval_mode") in ["vectors", "hybrid", None]
         use_semantic_captions = True if overrides.get("semantic_captions") and has_text else False
@@ -161,7 +161,7 @@ If you cannot generate a search query, return just the number 0.
 
         return {"data_points": results, "answer": chat_content, "thoughts": f"Searched for:<br>{query_text}<br><br>Conversations:<br>" + msg_to_display.replace('\n', '<br>')}
 
-    def get_messages_from_history(self, system_prompt: str, model_id: str, history: Sequence[dict[str, str]], user_conv: str, few_shots = [], max_tokens: int = 4096) -> []:
+    def get_messages_from_history(self, system_prompt: str, model_id: str, history: list[dict[str, str]], user_conv: str, few_shots = [], max_tokens: int = 4096) -> list:
         message_builder = MessageBuilder(system_prompt, model_id)
 
         # Add examples to show the chat what responses we want. It will try to mimic any responses and make sure they match the rules laid out in the system message.
@@ -174,9 +174,10 @@ If you cannot generate a search query, return just the number 0.
         message_builder.append_message(self.USER, user_content, index=append_index)
 
         for h in reversed(history[:-1]):
-            if h.get("bot"):
-                message_builder.append_message(self.ASSISTANT, h.get('bot'), index=append_index)
-            message_builder.append_message(self.USER, h.get('user'), index=append_index)
+            if bot_msg := h.get("bot"):
+                message_builder.append_message(self.ASSISTANT, bot_msg, index=append_index)
+            if user_msg := h.get("user"):
+                message_builder.append_message(self.USER, user_msg, index=append_index)
             if message_builder.token_length > max_tokens:
                 break
 
--- a/app/backend/approaches/chatreadretrieveread.py
+++ b/app/backend/approaches/chatreadretrieveread.py
@@ -1,7 +1,7 @@
 from typing import Any, Sequence
 
 import openai
-from azure.search.documents import SearchClient
+from azure.search.documents.aio import SearchClient
 from azure.search.documents.models import QueryType
 
 from approaches.approach import Approach
@@ -57,7 +57,7 @@ If you cannot generate a search query, return just the number 0.
         self.content_field = content_field
         self.chatgpt_token_limit = get_token_limit(chatgpt_model)
 
-    def run(self, history: Sequence[dict[str, str]], overrides: dict[str, Any]) -> Any:
+    async def run(self, history: Sequence[dict[str, str]], overrides: dict[str, Any]) -> Any:
         has_text = overrides.get("retrieval_mode") in ["text", "hybrid", None]
         has_vector = overrides.get("retrieval_mode") in ["vectors", "hybrid", None]
         use_semantic_captions = True if overrides.get("semantic_captions") and has_text else False
@@ -77,7 +77,7 @@ If you cannot generate a search query, return just the number 0.
             self.chatgpt_token_limit - len(user_q)
             )
 
-        chat_completion = openai.ChatCompletion.create(
+        chat_completion = await openai.ChatCompletion.acreate(
             deployment_id=self.chatgpt_deployment,
             model=self.chatgpt_model,
             messages=messages,
@@ -93,7 +93,7 @@ If you cannot generate a search query, return just the number 0.
 
         # If retrieval mode includes vectors, compute an embedding for the query
         if has_vector:
-            query_vector = openai.Embedding.create(engine=self.embedding_deployment, input=query_text)["data"][0]["embedding"]
+            query_vector = (await openai.Embedding.acreate(engine=self.embedding_deployment, input=query_text))["data"][0]["embedding"]
         else:
             query_vector = None
 
@@ -103,9 +103,9 @@ If you cannot generate a search query, return just the number 0.
 
         # Use semantic L2 reranker if requested and if retrieval mode is text or hybrid (vectors + text)
         if overrides.get("semantic_ranker") and has_text:
-            r = self.search_client.search(query_text,
+            r = await self.search_client.search(query_text,
                                           filter=filter,
-                                          query_type=QueryType.SEMANTIC,
+                                          query_type=QueryType.SIMPLE,
                                           query_language="en-us",
                                           query_speller="lexicon",
                                           semantic_configuration_name="default",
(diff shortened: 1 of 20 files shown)


tonybaloney (changes requested):

app/backend/approaches/retrievethenread.py
```diff
@@ -63,9 +63,9 @@ def run(self, q: str, overrides: dict[str, Any]) -> Any:
 
         # Use semantic ranker if requested and if retrieval mode is text or hybrid (vectors + text)
         if overrides.get("semantic_ranker") and has_text:
-            r = self.search_client.search(query_text,
+            r = await self.search_client.search(query_text,
                                           filter=filter,
-                                          query_type=QueryType.SEMANTIC,
+                                          query_type=QueryType.SIMPLE,
```
srbalakr, line 68: Why this change ?

pamelafox: Good catch, that was unintentionally committed, thats a change I make locally due to running out of semantic search credits this month. Reverted.

app/backend/approaches/chatreadretrieveread.py
```diff
@@ -57,7 +57,7 @@ def __init__(self, search_client: SearchClient, chatgpt_deployment: str, chatgpt
         self.content_field = content_field
         self.chatgpt_token_limit = get_token_limit(chatgpt_model)
 
-    def run(self, history: Sequence[dict[str, str]], overrides: dict[str, Any]) -> Any:
+    async def run(self, history: Sequence[dict[str, str]], overrides: dict[str, Any]) -> Any:
```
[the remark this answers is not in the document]
pamelafox: Fixed, also fixed other issues brought up by mypy (except that model/client issue that seems like a bug somewhere, langchain or pydantic)

app/backend/approaches/readretrieveread.py
```diff
@@ -142,8 +147,9 @@ def __init__(self, employee_name: str, callbacks: Callbacks = None):
                          name="Employee",
                          description="useful for answering questions about the employee, their benefits and other personal information",
                          callbacks=callbacks)
-        self.func = self.employee_info
+        self.func = lambda _: 'Not implemented'
+        self.coroutine = self.employee_info
         self.employee_name = employee_name
 
-    def employee_info(self, name: str) -> str:
+    async def employee_info(self, name: str) -> str:
```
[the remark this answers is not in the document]
pamelafox: Yes, due to the agent being called with arun. Otherwise this is the error:

```
Traceback (most recent call last):
  File "/workspaces/azure-search-openai-demo/app/backend/app.py", line 90, in ask
    r = await impl.run(request_json["question"], request_json.get("overrides") or {})
  File "/workspaces/azure-search-openai-demo/app/backend/approaches/readretrieveread.py", line 134, in run
    result = await agent_exec.arun(q)
  File "/workspaces/azure-search-openai-demo/app/backend/backend_env/lib/python3.10/site-packages/langchain/chains/base.py", line 263, in arun
    return (await self.acall(args[0], callbacks=callbacks))[self.output_keys[0]]
  File "/workspaces/azure-search-openai-demo/app/backend/backend_env/lib/python3.10/site-packages/langchain/chains/base.py", line 178, in acall
    raise e
  File "/workspaces/azure-search-openai-demo/app/backend/backend_env/lib/python3.10/site-packages/langchain/chains/base.py", line 172, in acall
    await self._acall(inputs, run_manager=run_manager)
  File "/workspaces/azure-search-openai-demo/app/backend/backend_env/lib/python3.10/site-packages/langchain/agents/agent.py", line 1002, in _acall
    next_step_output = await self._atake_next_step(
  File "/workspaces/azure-search-openai-demo/app/backend/backend_env/lib/python3.10/site-packages/langchain/agents/agent.py", line 928, in _atake_next_step
    result = await asyncio.gather(
  File "/workspaces/azure-search-openai-demo/app/backend/backend_env/lib/python3.10/site-packages/langchain/agents/agent.py", line 909, in _aperform_agent_action
    observation = await tool.arun(
  File "/workspaces/azure-search-openai-demo/app/backend/backend_env/lib/python3.10/site-packages/langchain/tools/base.py", line 358, in arun
    raise e
  File "/workspaces/azure-search-openai-demo/app/backend/backend_env/lib/python3.10/site-packages/langchain/tools/base.py", line 330, in arun
    await self._arun(*tool_args, run_manager=run_manager, **tool_kwargs)
  File "/workspaces/azure-search-openai-demo/app/backend/backend_env/lib/python3.10/site-packages/langchain/tools/base.py", line 438, in _arun
    raise NotImplementedError("Tool does not support async")
NotImplementedError: Tool does not support async
```

pamelafox:
Btw, the creator of Quart took a look at this PR and says the Quart usage looks good to him:
https://fosstodon.org/@pgjones/110842739565496118

tonybaloney (approved):

app/backend/approaches/__init__.py
srbalakr, line 1: can this be removed ?

app/backend/approaches/chatreadretrieveread.py
```diff
@@ -115,16 +115,16 @@ def run(self, history: Sequence[dict[str, str]], overrides: dict[str, Any]) -> A
...
                                           top=top,
                                           vector=query_vector,
                                           top_k=50 if query_vector else None,
                                           vector_fields="embedding" if query_vector else None)
         if use_semantic_captions:
-            results = [doc[self.sourcepage_field] + ": " + nonewlines(" . ".join([c.text for c in doc['@search.captions']])) for doc in r]
+            results = [doc[self.sourcepage_field] + ": " + nonewlines(" . ".join([c.text for c in doc['@search.captions']])) async for doc in r]
         else:
-            results = [doc[self.sourcepage_field] + ": " + nonewlines(doc[self.content_field]) for doc in r]
+            results = [doc[self.sourcepage_field] + ": " + nonewlines(doc[self.content_field]) async for doc in r]
```
srbalakr, line 127: Curious, async in iterations, does it yield faster results ? Our list size does not exceeed 10 or 20 at max !

app/backend/approaches/__init__.py
pamelafox: This was something I added for mypy, it was confused without it. Generally Python modules should have an __init__.py 
See https://docs.python.org/3/tutorial/modules.html#packages
Let me know if you think there's a reason to not have one.

app/backend/approaches/chatreadretrieveread.py
pamelafox: I assume it's not a big change in performance, but I had to make the change for API compatibility reasons once I started using the aio client. I've done a little loadtesting so far but not deep profiling, I need to try @Anthony's Opentelemetry integration.

app/backend/approaches/__init__.py
srbalakr: got it. My impression was that it would be implicitly created on the runtime. Also app worked previously without it. Lets keep it if thats the good practice.

srbalakr (approved):

chuwik (approved):

vrajroutu:
@pamelafox could you also merge the streaming feature to the quart branch? Streaming is much more faster and starts generating the responses right away, instead of waiting for the complete response to be generated.

BitcoinCode:
@pamelafox thank you , this is really useful, could you please shed some light on the following as this is important for me to understand clearly:

What are the 4 APIs that get called when user submits a question ?
How many approaches the application uses, to the best of my understanding it uses the chatreadretrieveread.py file ?
Is there any tech documentation for the applications that would help understand how the application works end to end ?

appreciate your help on this

BitcoinCode:
I've implemented the Quart changes by @pamelafox but the performance is still the same  if two users using the application it answers some questions then just stuck on generate answer... then the json error, the only thing I could think off is that the azure subscription is in EU but the repo deployed to east US, would this contribute to the issue ?  

I have also upgraded the app service from B1 to B3 and scaled out the instances manually to 3 instances as the auto scale is greyed out

is there anything else I can do to improve the performance with the assumption that  30 users to be accessing the app around the same time ?

**Failed to forward request to http://000.000.000.00:8000. 
Encountered a System.Threading.Tasks.TaskCanceledException 
exception after 300221.453ms with message: 
The request was canceled due to the configured 
HttpClient.Timeout of 300 seconds elapsing.. 
Check application logs to verify the application is properly handling HTTP traffic.**

Appreciate your help.

Request time out

json error

pamelafox:
@BitcoinCode What region is your OpenAI instance deployed in? My loadtests have been in South Central US currently, and I have been recommended to try one of these regions:
Australia East
Canada East
East United States 2
Japan East
United Kingdom South

I'll re-provision to one of those and re-run loadtests on this branch.

vrajroutu:
@pamelafox  I have also experimented with this branch. It performed satisfactorily for me, demonstrating noticeable improvements in performance. However, I am utilizing content streaming, which could pose challenges when using Quart and asynchronous requests.

BitcoinCode:
> @BitcoinCode What region is your OpenAI instance deployed in? My loadtests have been in South Central US currently, and I have been recommended to try one of these regions: Australia East Canada East East United States 2 Japan East United Kingdom South
> 
> I'll re-provision to one of those and re-run loadtests on this branch.

My Open AI instance is in East US, this is interesting as I have another Open AI instance on West Europe and this instance doesn't seem to have this issue even with Flask, I will implement Quart on it just in case.

Thank you.

BitcoinCode:
> @pamelafox I have also experimented with this branch. It performed satisfactorily for me, demonstrating noticeable improvements in performance. However, I am streaming the responses, which could pose challenges when using Quart and asynchronous requests.

would you please share the code for streaming the responses?

pamelafox:
This is now merged. If you want to incorporate these changes in an existing branch, make sure you run `azd up` since this requires re-provisioning to update the appCommandLine. To avoid re-indexing docs, you can delete the postprovision hooks section from azure.yaml.

pamelafox:
@BitcoinCode I will create a PR when the streaming functionality is ready. You can watch this repo to be notified.

ShantanuNair:
> I've been doing additional load testing on this branch, and though I think it's an improvement, there are still some hurdles to getting 0% errors.
> 
> First I ran into the issue described here: [openai/openai-python#371](https://github.com/openai/openai-python/issues/371)
> 
> There were two suggested workarounds there, setting the session yourself and using tenacity to retry. I implemented the second suggestion and re-ran the loadtests. I no longer encountered the "Connection reset" error, but I still had errors with responses taking too long and timing out.
> 
> I realized that App Service has a hard limit of 230 seconds, which corresponded with when errors occurred: https://learn.microsoft.com/en-us/troubleshoot/azure/app-service/web-apps-performance-faqs#why-does-my-request-time-out-after-230-seconds
> 
> I then tried specifying a `request_timeout` parameter for the OpenAI API request, but that caused every request to timeout, due to this issue: [openai/openai-python#387](https://github.com/openai/openai-python/pull/387)
> 
> My proposal for next steps is to implement retry on the _client_ in JavaScript, but I am still discussing this with other Azure folks for ideas on optimal configuration.

This was super helpful @pamelafox I'm running into these issues via langchain, which uses the openai-python client. Can I ask - why didn't you go with the first solution of setting the context? And has this issue been solved in openai-python([I ask because of this PR](https://github.com/openai/openai-python/pull/390/files))?

pamelafox:
@ShantanuNair I did actually end up using the first approach of setting the session:
https://github.com/Azure-Samples/azure-search-openai-demo/blob/e91496076f2b061b0d9f898d3bdc3b9b87633323/app/backend/app.py#L101
I didn't add it to the stream endpoint yet as it doesn't seem needed based on my loadtests so far, but I will keep testing.
I do not believe there has been a resolution in openai-python, based off what I've read of their changelog.

ShantanuNair:
@pamelafox Appreciate your help! I thought the PR I linked, allows you to pass in a ClientSession, similar to what you're doing? You are setting a new session on every request right?

```
I then tried specifying a request_timeout parameter for the OpenAI API request, but that caused every request to timeout, due to this issue: https://github.com/openai/openai-python/pull/387
```

It makes me wonder how people are using the library in long running chains in production. Also, do you ever notice a tonne of 502 bad_gateway errors when you send multiple requests with a large prompt token count? I do with 11-12k token inputs on 3.5k-turbo-16k openai, and I assumed it's because of langchain's tenacity retries failing due to the openai issue#371.

PR #507, opened 2023-08-08
Adding support for non-Azure openai instances

carlotta94c:
## Purpose
This PR adds OpenAI API support for users who don't have access to Azure OpenAI. Additions:

- Calling OpenAI endpoint if env variable OPENAI_API_TYPE is set to openai (default models used: text-davinci-003, gpt-3.5-turbo and text-embedding-ada-002)
- Making bicep configuration conditional (Azure OpenAI resource is created only if OPENAI_API_TYPE env variable is set to 'azure')
- Adding instruction to switch from an Azure OpenAI endpoint (default) to a non-azure OpenAI instance

```
[ ] Yes
[x ] No
```

## Pull Request Type

```
[ ] Bugfix
[ x] Feature
[x ] Documentation content changes
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code
Configure an existing non-Azure OpenAI instance and deploy the app:
1. Run `azd env set OPENAI_API_TYPE openai`
2. Run `azd env set OPENAI_ORGANIZATION {Your OpenAI organization}`
3. Run `azd env set OPENAI_API_KEY {Your OpenAI API key}`
4. Run `azd up`
5. Test the app by running it locally (Run Task -> Start App in VSCode) and in the Cloud (click on the link printed in the console after deployment)

## What to Check
* Running 'azd up' without changing env variables settings described above will ensure the same behavior of provisioning and deployment as before this PR (using Azure OpenAI resource is the default).


bgadrian:
Is there an extra config step is being missed out from the readme? openai API responds with an error "that the model is not specified", when the embeddings are done.

carlotta94c:
> Is there an extra config step is being missed out from the readme? openai API responds with an error "that the model is not specified", when the embeddings are done.

@bgadrian thanks for the heads up! Yes I did miss the configuration of the openai embedding model name. I added it into the bicep file.

app/backend/approaches/chatreadretrieveread.py
```diff
@@ -77,13 +79,22 @@ def run(self, history: Sequence[dict[str, str]], overrides: dict[str, Any]) -> A
...
-        chat_completion = openai.ChatCompletion.create(
-            deployment_id=self.chatgpt_deployment,
-            model=self.chatgpt_model,
-            messages=messages,
-            temperature=0.0,
-            max_tokens=32,
-            n=1)
+        if self.openai_type == "azure":
+            chat_completion = openai.ChatCompletion.create(
+                deployment_id=self.chatgpt_deployment,
```
pamelafox, line 84: @carlotta94c I'm wondering if we can avoid all the if/else with the OpenAI calls. Is it possible to just pass None or an empty string here, or does that completely confuse the OpenAI SDK?

pamelafox: Suggestion from Krista:

kwargs = {"model": self.chatgpt_model} if api_type == "openai" \
    else {"deployment_id": self.chatgpt_deployment}

Can be used like:

response = openai.ChatCompletion.create(messages=messages, stream=True, **kwargs)

Can you try that style? That should reduce if/else

carlotta94c: I applied Krista's suggestion and it works well. Can you please check it @pamelafox ?

pamelafox:
@carlotta94c I've merged main, made some style changes, verified that Azure OpenAI still works, and added unit tests (which resulted in snapshot updates due to the paramaterization). Please verify that non-Azure OpenAi deployment still works for you.

infra/main.bicep
```diff
@@ -54,7 +54,7 @@ param formRecognizerSkuName string = 'S0'
 
 param chatGptDeploymentName string // Set in main.parameters.json
 param chatGptDeploymentCapacity int = 30
-param chatGptModelName string = 'gpt-35-turbo'
+param chatGptModelName string = (openAiHost == 'azure') ? 'gpt-35-turbo' : 'gpt-3.5-turbo'
```
pamelafox, line 57: Ohhh forgot about that!

carlotta94c:
> @carlotta94c I've merged main, made some style changes, verified that Azure OpenAI still works, and added unit tests (which resulted in snapshot updates due to the paramaterization). Please verify that non-Azure OpenAi deployment still works for you.

@pamelafox I've add some fixes and tested again deploy and run with non-Azure openai endpoint. It's working well now!

pamelafox (approved):
Okay, there's still a non-optimal user experience for non-Azure OpenAI users where they get prompted for the location, but everything seems to work. Thanks for all the work on this Carlotta.

PR #515, opened 2023-08-08
Add application insights for request and backend tracing of performance data

tonybaloney:
## Application Insights

This PR adds Application Insights with tracing and logging in the app using Open Telemetry

## End-to-End tracing of chat requests

Admins can see the trace of any HTTP request including all backend calls to Cognitive Search and Open AI APIs.

screenshot 2023-08-09 at 09 21 36

## Exceptions captured and logged

Server-side exceptions are now logged and available in the Failures page

screenshot 2023-08-09 at 09 20 38


app/backend/app.py
```diff
@@ -117,12 +120,20 @@ def ensure_openai_token():
 
 
 def create_app():
+    if APPLICATIONINSIGHTS_CONNECTION_STRING:
```
pamelafox, line 123: Do you want to be more specific and say is not None, or do you intentionally want to account for falsy values like ""?

app/backend/requirements.txt
```diff
@@ -5,3 +5,8 @@ openai[datalib]==0.27.8
 tiktoken==0.4.0
 azure-search-documents==11.4.0b6
 azure-storage-blob==12.14.1
+
+azure-monitor-opentelemetry --pre
```
pamelafox, line 9: No version numbers?

app/backend/app.py
```diff
@@ -117,12 +120,20 @@ def ensure_openai_token():
...
+        configure_azure_monitor()
+
     app = Flask(__name__)
+    FlaskInstrumentor().instrument_app(app)
+
+    # For local debugging, requires opentelemetry-exporter-richconsole package
```
pamelafox, line 129:
```suggestion
    # For local debugging, add opentelemetry-exporter-richconsole to requirements.txt and un-comment:
```

infra/core/monitor/applicationinsights.bicep
```diff
@@ -0,0 +1,17 @@
+param name string
+param location string = resourceGroup().location
+param tags object = {}
```
pamelafox, line 3: Are these straight copies from azure-dev?

app/backend/app.py
pamelafox: Ah based on seeing Bicep, I think you're intentionally accounting for "" since thats whats get set by default.

infra/main.bicep
```diff
@@ -123,6 +137,7 @@ module backend 'core/host/appservice.bicep' = {
       AZURE_OPENAI_GPT_DEPLOYMENT: gptDeploymentName
       AZURE_OPENAI_CHATGPT_DEPLOYMENT: chatGptDeploymentName
       AZURE_OPENAI_EMB_DEPLOYMENT: embeddingDeploymentName
+      APPLICATIONINSIGHTS_CONNECTION_STRING: monitoring.outputs.applicationInsightsConnectionString
```
pamelafox, line 140: You can also do a fancy union thing where you dont set it at all if its empty, but this is probably fine.

pamelafox (changes requested):
I think we need versions pinned?

app/backend/app.py
tonybaloney: I want a False-y statement, if its an empty string it'll fail to boot

infra/core/monitor/applicationinsights.bicep
tonybaloney: yes

app/backend/requirements.txt
tonybaloney: The version dependencies for Otel are tricky right now until there's a V1. It needs to be flexible. I could set an upper-limit incase there are more breaking changes

pamelafox: Okay, thats fine, just add a comment about that. Upper limit is good if possible.

pamelafox (approved):

pamelafox:
Please update to ASGI goodness

pamelafox:
@chuwik suggests making it opt in, like a USE_OPENTELEMETRY or USE_APPINSIGHTS parameter perhaps?

infra/main.parameters.json
```diff
@@ -55,6 +55,9 @@
     },
     "gptDeploymentName": {
       "value": "${AZURE_OPENAI_GPT_DEPLOYMENT=davinci}"
+    },
+    "useApplicationInsights": {
+      "value": "${AZURE_USE_APPLICATION_INSIGHTS=false}"
```
pamelafox, line 60: I notice the TODO templates also output the similar USE_APIM in main.bicep - am not sure when/if thats needed for this scenario: https://github.com/pamelafox/azure-dev/blob/37f88d5196a3aed4899f2928a67f43714dbff967/templates/todo/projects/csharp-sql/.repo/bicep/infra/main.bicep#L200

If you're not sure either, maybe @vhvb1989 can chime in.

app/backend/app.py
```diff
@@ -186,6 +189,10 @@ async def setup_clients():
 
 
 def create_app():
+    if APPLICATIONINSIGHTS_CONNECTION_STRING:
+        configure_azure_monitor()
```
pamelafox, line 193: does this default to 100% sampling rate or smaller? Is the default appropriate for production or should we make it obvious how to change the rate?

pamelafox (approved):
A couple questions but no clear blockers, very excited to try this out!

app/backend/app.py
tonybaloney: Yes, 100%. It shouldn't interfere with performance because it uses batch span exporter (waits until idle to send events and is non-blocking for most things)

PR #520, opened 2023-08-09
Update main.bicep with more allowed regions

pamelafox:
## Purpose

This PR updates @allowed per the regions in:
https://learn.microsoft.com/en-us/azure/ai-services/openai/concepts/models#model-summary-table-and-region-availability

This will allow devs to pick regions that aren't as overloaded as the originally allowed regions, like southcentralus.

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[X] Other... Please describe:
```

## How to Test
*  azd up, see picker display all the options, choose one and deploy

### DIFF
--- a/infra/main.bicep
+++ b/infra/main.bicep
@@ -113,7 +113,7 @@ module backend 'core/host/appservice.bicep' = {
     appServicePlanId: appServicePlan.outputs.id
     runtimeName: 'python'
     runtimeVersion: '3.10'
-    appCommandLine: 'python3 -m gunicorn main:app'
+    appCommandLine: 'python3 -m gunicorn "app:create_app()"'
     scmDoBuildDuringDeployment: true
     managedIdentity: true
     appSettings: {
--- a/infra/main.bicep
+++ b/infra/main.bicep
@@ -28,7 +28,7 @@ param storageContainerName string = 'content'
 param openAiServiceName string = ''
 param openAiResourceGroupName string = ''
 @description('Location for the OpenAI resource group')
-@allowed(['australiaeast', 'canadaeast', 'eastus', 'eastus2', 'francecentral', 'japaneast', 'northcentralus', 'uksouth'])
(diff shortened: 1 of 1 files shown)


pamelafox:
Not yet ready for review

infra/main.bicep
```diff
@@ -28,7 +28,7 @@ param storageContainerName string = 'content'
 param openAiServiceName string = ''
 param openAiResourceGroupName string = ''
 @description('Location for the OpenAI resource group')
-@allowed(['eastus', 'southcentralus', 'westeurope'])
+@allowed(['canadaeast', 'eastus', 'francecentral', 'japaneast', 'northcentralus'])
```
pamelafox, line 31: This is the intersection of the embedding and gpt-35 regions. Unfortunately they're not available in the same regions, so we're providing less options than if we had multiple parameters.

infra/main.bicep
```diff
@@ -47,9 +47,11 @@ param formRecognizerSkuName string = 'S0'
 param gptDeploymentName string // Set in main.parameters.json
 param gptDeploymentCapacity int = 30
 param gptModelName string = 'gpt-35-turbo'
+param gptModelVersion string = '0613'
```
pamelafox, line 50: I think @jongio wanted this, keeping the version numbers as parameters next to the model names, since they're so related. Also, I had to update the version from 0310 to 0613, as it wouldn't let me deploy new 0301s. I'm still trying to figure out differences between the two, i.e. if 0613 is always better.

pamelafox: @pablocastro Was there a reason to use 0310 over 0613 for our use case?

tonybaloney (approved):

pamelafox:
I'm going to merge due to how many devs are running into region-related issues. Let me know if you have concerns about the version change.

PR #529, opened 2023-08-11
Address recent bugs

srbalakr:
## Purpose
This PR address some of the recent bugs
1. Users experience 401 during index build.
   - Refresh token every 5 minutes for open ai calls
2. Suggestion prompt does not work
   - Gpt turbo do not handle lengthy system prompts. Moved document sources to user conversation
   
```
[ ] Yes
[x ] No
```

## Pull Request Type

```
[ x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
```

## What to Check
* ...

### DIFF
--- a/app/backend/approaches/chatreadretrieveread.py
+++ b/app/backend/approaches/chatreadretrieveread.py
@@ -144,7 +144,7 @@ If you cannot generate a search query, return just the number 0.
             system_message,
             self.chatgpt_model,
             history,
-            history[-1]["user"]+ "\n\nSources:\n" + content,
+            history[-1]["user"]+ "\n\nSources:\n" + content, # Model does not handle lengthy system messages well. Moving sources to latest user conversation to solve follow up questions prompt.
             max_tokens=self.chatgpt_token_limit)
 
         chat_completion = await openai.ChatCompletion.acreate(
--- a/app/backend/approaches/chatreadretrieveread.py
+++ b/app/backend/approaches/chatreadretrieveread.py
@@ -141,10 +141,10 @@ If you cannot generate a search query, return just the number 0.
             system_message = prompt_override.format(follow_up_questions_prompt=follow_up_questions_prompt)
 
         messages = self.get_messages_from_history(
-            system_message + "\n\nSources:\n" + content,
+            system_message,
             self.chatgpt_model,
             history,
-            history[-1]["user"],
+            history[-1]["user"]+ "\n\nSources:\n" + content,
             max_tokens=self.chatgpt_token_limit)
 
         chat_completion = await openai.ChatCompletion.acreate(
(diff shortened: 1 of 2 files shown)


app/backend/approaches/chatreadretrieveread.py
```diff
@@ -141,10 +141,10 @@ async def run(self, history: list[dict[str, str]], overrides: dict[str, Any]) ->
             system_message = prompt_override.format(follow_up_questions_prompt=follow_up_questions_prompt)
 
         messages = self.get_messages_from_history(
-            system_message + "\n\nSources:\n" + content,
+            system_message,
             self.chatgpt_model,
             history,
-            history[-1]["user"],
+            history[-1]["user"]+ "\n\nSources:\n" + content,
```
pamelafox, line 147: It seems un-intuitive to me that sources would be part of a user message, so it'd help to add a comment here.

scripts/prepdocs.py
```diff
@@ -35,6 +35,11 @@
 SENTENCE_SEARCH_LIMIT = 100
 SECTION_OVERLAP = 100
 
+OPEN_AI_TOKEN_CACHE = {}
```
pamelafox, line 38: The capitalization makes this seem like a constant, but its a mutable dict, so I think it should just be lowercase. 
https://pep8.org/#constants

scripts/prepdocs.py
```diff
@@ -301,6 +307,12 @@ def remove_from_index(filename):
         # It can take a few seconds for search results to reflect changes, so wait a bit
         time.sleep(2)
 
+# refresh open ai token every 5 minutes
+def refresh_openai_token():
+    if OPEN_AI_TOKEN_CACHE[CACHE_KEY_TOKEN_TYPE] == 'azure_ad' and OPEN_AI_TOKEN_CACHE[CACHE_KEY_CREATED_TIME] + 300 < time.time():
```
pamelafox, line 312: I'm bad with time math, but are you sure this should be less than? I keep thinking it should be >

pamelafox (changes requested):
In the future, I think it'd be nice to send unrelated bug fixes in two different PRs. That makes it easy to patch or revert single fixes as needed. Thanks!

app/backend/approaches/chatreadretrieveread.py
srbalakr: done

scripts/prepdocs.py
srbalakr: yes.
assume token created time 00:00
Upto 00:05, we dont refresh token

We refresh token once created time lapses 5 minutes.

pamelafox (approved):
Okay, I stepped through the time logic with my partner, it does work - I was flipping the times in my head.

chuwik (approved):

PR #532, opened 2023-08-12
Adds optional streaming support for chat requests

pamelafox:
## Purpose

This PR adds optional streaming support to the Chat API. The backend [sends an "event/stream"](https://quart.palletsprojects.com/en/latest/how_to_guides/streaming_response.html) of [newline-delimited JSON](http://ndjson.org/) and the frontend parses it as a [ReadableStream](https://developer.mozilla.org/en-US/docs/Web/API/Streams_API/Using_readable_streams). This is the same approach used in https://github.com/Azure-Samples/chatgpt-quickstart and https://github.com/microsoft/sample-app-aoai-chatGPT/blob/main/README_azd.md

I did not add it to the /ask endpoint as it doesn't seem as obviously needed there. I made it optional as some devs have expressed difficulty with using streams over proxies like APIM.

It could be breaking for some if we default it to true (as it is in UI currently), if they deploy to a setup where streaming doesn't work. But I think they'll notice that.

```
[X] Yes
[ ] No
```

## Pull Request Type

```
[ ] Bugfix
[X] Feature
[ ] Other... Please describe:
```

## How to Test

* app/start.sh, chat, see streaming used for response
* disable stream in settings, chat, see no streaming
* deploy and repeat
* run tests


vrajroutu:
I tested this just now, and it's working very smoothly. Thanks for the PR.

app/backend/app.py
```diff
@@ -103,8 +111,15 @@ async def chat():
         impl = current_app.config[CONFIG_CHAT_APPROACHES].get(approach)
         if not impl:
             return jsonify({"error": "unknown approach"}), 400
-        r = await impl.run(request_json["history"], request_json.get("overrides") or {})
-        return jsonify(r)
+        should_stream = request_json.get("should_stream", False)
+        response_generator = impl.run(request_json["history"], request_json.get("overrides", {}), should_stream)
```
tonybaloney, line 115: `impl.run` is async, so this should be `await`

app/backend/app.py
```diff
@@ -103,8 +111,15 @@ async def chat():
...
+        if should_stream:
+            response = await make_response(format_as_ndjson(response_generator))
+            response.timeout = None
+            return response
+        else:
+            # anext() is nicer but only available in Python 3.10
+            return jsonify(await response_generator.__anext__())
```
tonybaloney, line 122: this will only call `__anext__` once though, is that what we need? Doesn't it need to exhaust the iterator?

app/backend/approaches/chatreadretrieveread.py
```diff
@@ -146,20 +146,27 @@ async def run(self, history: list[dict[str, str]], overrides: dict[str, Any]) ->
...
-            max_tokens=1024,
-            n=1)
-
-        chat_content = chat_completion.choices[0].message.content
-
         msg_to_display = '\n\n'.join([str(message) for message in messages])
 
-        return {"data_points": results, "answer": chat_content, "thoughts": f"Searched for:<br>{query_text}<br><br>Conversations:<br>" + msg_to_display.replace('\n', '<br>')}
+        extra_info = {"data_points": results, "thoughts": f"Searched for:<br>{query_text}<br><br>Conversations:<br>" + msg_to_display.replace('\n', '<br>')}
+        chat_coroutine = openai.ChatCompletion.acreate(
```
tonybaloney, line 152: Missing an await statement

tonybaloney (dismissed):

app/backend/approaches/chatreadretrieveread.py
pamelafox: I await it later, since the order depends on streaming or not

srbalakr:
Hi @pamelafox , on high level I am thinking if we could have a new api in backend to support streaming and new method in chat approach implementation to support streaming response. This helps for better maintenance and adding new support and features. If streaming fails or json implementation fails, it will be easy for us to track down.

@chuwik thoughts ?

pamelafox:
@srbalakr So, to clarify, a new HTTP API like:

/chat_stream ?

And then a wrapper method in ChatReadRetrieve that returns it as a stream? 
I can try that - the awkward part is wanting to get the thoughts/data_points yielded immediately, so that we don't have to send them over in every stream chunk - but I'll play around with that approach.

srbalakr:
> @srbalakr So, to clarify, a new HTTP API like:
> 
> /chat_stream ?
> 
> And then a wrapper method in ChatReadRetrieve that returns it as a stream? I can try that - the awkward part is wanting to get the thoughts/data_points yielded immediately, so that we don't have to send them over in every stream chunk - but I'll play around with that approach.

Yes we could start with a simpler version

srbalakr (approved):

app/backend/app.py
```diff
@@ -97,12 +100,36 @@ async def chat():
...
+        return jsonify({"error": "request must be json"}), 415
+    request_json = await request.get_json()
+    approach = request_json["approach"]
+    try:
+        impl = current_app.config[CONFIG_CHAT_APPROACHES].get(approach)
+        if not impl:
+            return jsonify({"error": "unknown approach"}), 400
+        response_generator = impl.run_with_streaming(request_json["history"], request_json.get("overrides", {}))
+        response = await make_response(format_as_ndjson(response_generator))
+        response.timeout = None
```
tonybaloney, line 126: It looks like `timeout` is a late-initialized attributed on Quart's response object, but not Werkzeug's response object. Since `make_response()` says it returns a Union of either type, this is giving me a type warning. Either `make_response()` return type shouldn't have the werkzeug type as a potential reply, or this will crash at runtime

tonybaloney (commented):
Apart from the comment on a missing attribute, LGTM

app/backend/app.py
pamelafox: Discussed with Anthony and Quart creator, just going to ignore for now, it doesn't appear to introduce a runtime error. https://github.com/pallets/quart/discussions/269

PR #534, opened 2023-08-14
Add optional AAD auth

pamelafox:
## Purpose

This PR adds optional AAD authentication to the App Service app, using the same approach I used in https://github.com/microsoft/sample-app-aoai-chatGPT/blob/main/README_azd.md

This is opt-in, developers must enable it using ```azd env set AZURE_USE_AUTHENTICATION true```
That's similar to how @tonybaloney made App Insights opt-in.

This does not currently work on GitHub actions due to the service principal lacking the permissions.

In the future, I would prefer to set it up completely in Bicep (vs Bicep + 2 scripts), but am eagerly awaiting https://github.com/Azure/bicep/issues/7724 for that to be possible.

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[ ] Bugfix
[X] Feature
[ ] Other... Please describe:
```

## How to Test

* azd up, confirm that it works with no auth
* azd env set AZURE_USE_AUTHENTICATION true
* azd up, confirm auth is required


scripts/auth_init.sh
```diff
@@ -0,0 +1,5 @@
+ #!/bin/sh
+
+. ./scripts/loadenv.sh
```
pamelafox, line 3: FYI: Calling loadenv.sh in every script (auth_init, auth_update, prepdocs) means that it pip installs requirements every time. I could just call it in the first script, but then other hooks won't work if a dev deletes the first script hook for some reason. 
I believe I also tried calling load_env as its own hook, but then the other hooks executed in a separated environment and didnt have access to the exported variables.

Another approach is to dump the env variables into a .env file and load them that way, but that'd be a larger change. That is what I do in https://github.com/pamelafox/chatgpt-quickstart/blob/main/azure.yaml

scripts/auth_init.ps1
```diff
@@ -0,0 +1,9 @@
+. ./scripts/loadenv.ps1
```
zaffaste, line 1: | The term './scripts/loadenv.ps1' is not recognized as a name of a cmdlet, function, script file, or executable program. Check the spelling of the name, or if a path was included, verify that 
| the path is correct and try again.

zaffaste: ok, as for loadenv.sh I cut the first 35 lines of prepdocs.ps1 and I put them in a new loadenv.ps1 file, then I added a new line in prepdocs.ps1
`. ./scripts/loadenv.ps1`

scripts/auth_init.ps1
```diff
@@ -0,0 +1,9 @@
...
+
+$venvPythonPath = "./.venv/scripts/python.exe"
```
zaffaste, line 3: this path give me an error I replaced it with `"./scripts/.venv/scripts/python.exe"` and `"./scripts/.venv/bin/python"` in line 6

pamelafox:
@zaffaste I just pushed loadenv.ps1 and a related change to prepdocs.ps1, they both work for me now inside pwsh, can you try?

zaffaste:
> @zaffaste I just pushed loadenv.ps1 and a related change to prepdocs.ps1, they both work for me now inside pwsh, can you try?

@pamelafox 
If I don't change `$venvPythonPath = "./.venv/scripts/python.exe"` to `$venvPythonPath = "./scripts/.venv/scripts/python.exe"` in auth_init.ps1 and auth_update.ps1 files the scripts execution returns me an error 
`Line |
      9 |  Start-Process -FilePath $venvPythonPath -ArgumentList "./scripts/auth …
         |  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
         | This command cannot be run due to the error: The system cannot find the file specified.

ERROR: failed running post hooks: 'postprovision' hook failed with exit code: '1', Path: 'xxx\Local\Temp\azd-postprovision-3370852331.ps1'. : exit code: 1`

The script in loadenv.ps1 and the updates to prepdocs.ps1 works good for me, thank you!

scripts/prepdocs.ps1
zaffaste, line 1: @pamelafox same problem in this file: if the script create a python virtual environment in "./scripts/.venv" why we have to change Path to "./.venv/..." instead of "./scirpts/.venv/..." in line 3 and 6?

pamelafox: Ahh, sorry, I always create a .venv in root for myself, so I didn't notice that issue when testing. I'll change that to ./scripts/.venv. Thanks!

pamelafox:
Moving this to a draft as the new doc ACL requires slightly different setup (client/server) and @mattmsft is investigating scripting that (based on these scripts).

pamelafox:
Closing in favor of Matt's new PR.

PR #540, opened 2023-08-17
Remove unneeded deployment

pamelafox:
## Purpose

This PR removes the GPT deployment that we aren't currently using it, since it is redundant with the ChatGPT deployment (same model, gpt-35-turbo). We might use a different deployment in future if instruct becomes available and works well for query generation, but in the meantime, the additional deployment is confusing folks.

Also, while testing this, I discovered the two optional Ask approaches do *not* work with the Chat Completions API - they only work with the completions API, but 0613 doesn't support completions and is the only version that can be deployed newly now. I'll file that as an issue. 

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test

* azd up
* start server locally


srbalakr (approved):

srbalakr:
I was working on to make a PR. Thank you.

PR #550, opened 2023-08-22
Remove defaults for getenv

pamelafox:
## Purpose

This PR is inspired by debugging session with @connorwehrum - he was getting an unauthorized error on cognitive search, but the real issue is that he had never created a cognitive search and had no relevant environment variables set. I think it would be better if the code failed immediately when these env variables are not set. This should only affect users who are using something outside of the typical azd flow.

It may break people who didn't have these values set but were successfully using the defaults?

```
[X] Yes
[ ] No
```

## Pull Request Type

```
[X] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test

* Run locally, works
* Deploy, still works
* python3 -m pytest, all pass

### DIFF
--- a/app/backend/app.py
+++ b/app/backend/app.py
@@ -28,27 +28,11 @@ from approaches.readdecomposeask import ReadDecomposeAsk
 from approaches.readretrieveread import ReadRetrieveReadApproach
 from approaches.retrievethenread import RetrieveThenReadApproach
 
-# Replace these with your own values, either in environment variables or directly here
-AZURE_STORAGE_ACCOUNT = os.getenv("AZURE_STORAGE_ACCOUNT", "mystorageaccount")
-AZURE_STORAGE_CONTAINER = os.getenv("AZURE_STORAGE_CONTAINER", "content")
-AZURE_SEARCH_SERVICE = os.getenv("AZURE_SEARCH_SERVICE", "gptkb")
-AZURE_SEARCH_INDEX = os.getenv("AZURE_SEARCH_INDEX", "gptkbindex")
-AZURE_OPENAI_SERVICE = os.getenv("AZURE_OPENAI_SERVICE", "myopenai")
-AZURE_OPENAI_CHATGPT_DEPLOYMENT = os.getenv("AZURE_OPENAI_CHATGPT_DEPLOYMENT", "chat")
-AZURE_OPENAI_CHATGPT_MODEL = os.getenv("AZURE_OPENAI_CHATGPT_MODEL", "gpt-35-turbo")
-AZURE_OPENAI_EMB_DEPLOYMENT = os.getenv("AZURE_OPENAI_EMB_DEPLOYMENT", "embedding")
-
-KB_FIELDS_CONTENT = os.getenv("KB_FIELDS_CONTENT", "content")
-KB_FIELDS_CATEGORY = os.getenv("KB_FIELDS_CATEGORY", "category")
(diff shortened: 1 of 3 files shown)


app/backend/app.py
```diff
@@ -28,27 +28,11 @@
 from approaches.readretrieveread import ReadRetrieveReadApproach
 from approaches.retrievethenread import RetrieveThenReadApproach
 
-# Replace these with your own values, either in environment variables or directly here
-AZURE_STORAGE_ACCOUNT = os.getenv("AZURE_STORAGE_ACCOUNT", "mystorageaccount")
```
pamelafox, line 32: For ease of mocking the environment in tests, I moved these into the function below. This also means less globals, yay!

app/backend/app.py
```diff
@@ -69,8 +53,8 @@ async def assets(path):
 # can access all the files. This is also slow and memory hungry.
 @bp.route("/content/<path>")
 async def content_file(path):
-    blob_container = current_app.config[CONFIG_BLOB_CLIENT].get_container_client(AZURE_STORAGE_CONTAINER)
```
pamelafox, line 72: I now store the specific container client in the config, since we seem to always use that anyway (versus the more general storage client).

infra/main.bicep
```diff
@@ -135,6 +135,7 @@ module backend 'core/host/appservice.bicep' = {
       AZURE_SEARCH_INDEX: searchIndexName
       AZURE_SEARCH_SERVICE: searchService.outputs.name
       AZURE_OPENAI_CHATGPT_DEPLOYMENT: chatGptDeploymentName
+      AZURE_OPENAI_CHATGPT_MODEL: chatGptModelName
```
pamelafox, line 138: We were relying on the default value here, eek!

app/backend/app.py
```diff
@@ -130,20 +114,35 @@ async def ensure_openai_token():
...
+    KB_FIELDS_SOURCEPAGE = os.getenv("KB_FIELDS_SOURCEPAGE", "sourcepage")
+
     # Use the current user identity to authenticate with Azure OpenAI, Cognitive Search and Blob Storage (no secrets needed,
     # just use 'az login' locally, and managed identity when deployed on Azure). If you need to use keys, use separate AzureKeyCredential instances with the
     # keys for each service
     # If you encounter a blocking error during a DefaultAzureCredential resolution, you can exclude the problematic credential by using a parameter (ex. exclude_shared_token_cache_credential=True)
     azure_credential = DefaultAzureCredential(exclude_shared_token_cache_credential = True)
 
     # Set up clients for Cognitive Search and Storage
+    print("AZURE_SEARCH_INDEX", AZURE_SEARCH_INDEX)
```
chuwik, line 137: Leftover from debugging?

chuwik (changes requested):

app/backend/app.py
pamelafox: Good eye, removed!

infra/main.bicep
```diff
@@ -343,6 +344,7 @@ output AZURE_RESOURCE_GROUP string = resourceGroup.name
 output AZURE_OPENAI_SERVICE string = openAi.outputs.name
 output AZURE_OPENAI_RESOURCE_GROUP string = openAiResourceGroup.name
 output AZURE_OPENAI_CHATGPT_DEPLOYMENT string = chatGptDeploymentName
+output AZURE_OPENAI_CHATGPT_MODEL string = chatGptModelName
```
pamelafox, line 347: Also needs to be outputted for local dev to work

srbalakr (approved):

PR #569, opened 2023-08-31
Allow for subdirectory structure within data directory

PfisterAn:
- recursively read the data directory
  - @TODO/SUGGEST - may be worth to think about using the sub directory information for enriching the index and/or use it for access control (e.g., subdirectory names may be named like the Azure AD Group for access control)

## Purpose

* Ability to bring subfolder structure into data directory for better arranging the data, at least on the directory level for now

```
[ ] Yes
[ x] No
```

## Pull Request Type

```
[ ] Bugfix
[ x] Feature
[ ] Other... Please describe:
```

## How to Test
*  Run prepdocs on the data directory where content is arranged within subdirectories

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
```

## What to Check
* ...

### DIFF
--- a/scripts/prepdocs.py
+++ b/scripts/prepdocs.py
@@ -314,12 +314,12 @@ def refresh_openai_token():
         openai.api_key = token_cred.get_token("https://cognitiveservices.azure.com/.default").token
         open_ai_token_cache[CACHE_KEY_CREATED_TIME] = time.time()
 
-"""Readfile function
-
-recursively read directory structure under `pathPattern` and execute indexing for the individual files
-"""
-def read_files(pathPattern, use_vectors):
-    for filename in glob.glob(pathPattern):
+def read_files(path_pattern: str, use_vectors: bool):
+    """
+    Recursively read directory structure under `path_pattern`
+    and execute indexing for the individual files
+    """
+    for filename in glob.glob(path_pattern):
         if args.verbose: print(f"Processing '{filename}'")
         if args.remove:
             remove_blobs(filename)
--- a/scripts/prepdocs.py
+++ b/scripts/prepdocs.py
@@ -147,7 +147,7 @@ def get_document_text(filename):
 
     return page_map
 
-def split_text(page_map):
+def split_text(page_map, filename):
     SENTENCE_ENDINGS = [".", "!", "?"]
     WORDS_BREAKS = [",", ";", ":", " ", "(", ")", "[", "]", "{", "}", "\t", "\n"]
     if args.verbose: print(f"Splitting '{filename}' into sections")
@@ -214,7 +214,7 @@ def filename_to_id(filename):
 
 def create_sections(filename, page_map, use_vectors):
     file_id = filename_to_id(filename)
-    for i, (content, pagenum) in enumerate(split_text(page_map)):
+    for i, (content, pagenum) in enumerate(split_text(page_map, filename)):
         section = {
             "id": f"{file_id}-page-{i}",
             "content": content,
--- a/scripts/prepdocs.py
+++ b/scripts/prepdocs.py
@@ -314,9 +314,12 @@ def refresh_openai_token():
         openai.api_key = token_cred.get_token("https://cognitiveservices.azure.com/.default").token
         open_ai_token_cache[CACHE_KEY_CREATED_TIME] = time.time()
 
-# recursively read the directory to allow for a subdirectory structure
-def read_files(location, use_vectors):
-    for filename in glob.glob(location):
+"""Readfile function
+
+recursively read directory structure under `pathPattern` and execute indexing for the individual files
+"""
+def read_files(pathPattern, use_vectors):
+    for filename in glob.glob(pathPattern):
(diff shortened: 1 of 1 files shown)


scripts/prepdocs.py
```diff
@@ -314,6 +314,29 @@ def refresh_openai_token():
...
         open_ai_token_cache[CACHE_KEY_CREATED_TIME] = time.time()
 
+# recursively read the directory to allow for a subdirectory structure
+def read_files(location, use_vectors):
+    for filename in glob.glob(location):
+        if args.verbose: print(f"Processing '{filename}'")
+        if args.remove:
+            remove_blobs(filename)
+            remove_from_index(filename)
+        elif args.removeall:
```
pamelafox, line 324: This was already an issue, but should this if here be outside the function?

scripts/prepdocs.py
```diff
@@ -314,6 +314,29 @@ def refresh_openai_token():
         openai.api_key = token_cred.get_token("https://cognitiveservices.azure.com/.default").token
         open_ai_token_cache[CACHE_KEY_CREATED_TIME] = time.time()
 
+# recursively read the directory to allow for a subdirectory structure
+def read_files(location, use_vectors):
```
pamelafox, line 318: Could you type this to make it clear whether location is a string or a path-like object?

pamelafox:
Thanks, this seems useful!

scripts/prepdocs.py
PfisterAn: Agree, thought the wrong way here, at least the removeall needs to be outside the function, which already is, so its redundant actually and/or never executed anyways, will be removed!

scripts/prepdocs.py
PfisterAn: Should be fine now now, renamed from `location` to `pathPattern`

scripts/prepdocs.py
```diff
@@ -314,6 +314,29 @@ def refresh_openai_token():
         openai.api_key = token_cred.get_token("https://cognitiveservices.azure.com/.default").token
         open_ai_token_cache[CACHE_KEY_CREATED_TIME] = time.time()
 
+# recursively read the directory to allow for a subdirectory structure
```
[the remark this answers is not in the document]
PfisterAn: Done!

scripts/prepdocs.py
PfisterAn: This got fixed!

scripts/prepdocs.py
```diff
@@ -314,6 +314,29 @@ def refresh_openai_token():
         openai.api_key = token_cred.get_token("https://cognitiveservices.azure.com/.default").token
         open_ai_token_cache[CACHE_KEY_CREATED_TIME] = time.time()
 
+"""Readfile function
+
+recursively read directory structure under `pathPattern` and execute indexing for the individual files
+"""
+def read_files(pathPattern, use_vectors):
```
pamelafox, line 321: pathPattern should be path_pattern, per naming conventions detailed at pep8.org

Please also add a type annotation- learn more here: https://realpython.com/lessons/annotations/

scripts/prepdocs.py
```diff
@@ -314,6 +314,29 @@ def refresh_openai_token():
         openai.api_key = token_cred.get_token("https://cognitiveservices.azure.com/.default").token
         open_ai_token_cache[CACHE_KEY_CREATED_TIME] = time.time()
 
+"""Readfile function
```
pamelafox, line 317: Please move inside the function and remove top line, ie.

def read_files(path_pattern, use_vectors):
  """
  Recursively read directory structure under `path_pattern`
  and execute indexing for the individual files
  """

scripts/prepdocs.py
```diff
@@ -379,17 +402,4 @@ def refresh_openai_token():
             create_search_index()
 
         print("Processing files...")
-        for filename in glob.glob(args.files):
-            if args.verbose: print(f"Processing '{filename}'")
-            if args.remove:
-                remove_blobs(filename)
-                remove_from_index(filename)
-            elif args.removeall:
```
pamelafox, line 387: Did we lose removeall entirely? I expected it to move outside the function.

PfisterAn: Its in the __main__ body, it was already there before, thats why you dont see it in the Review.

https://github.com/Azure-Samples/azure-search-openai-demo/blob/36c14b6248b4a5c82e80c6682535990651b25bc6/scripts/prepdocs.py#L397-L405

scripts/prepdocs.py
PfisterAn: done

scripts/prepdocs.py
PfisterAn: done

PfisterAn:
Missed to pass the filename towards the split_text function. split_text was setup to utilise the global variable filename, which doesnt exist anymore after introducing the read_files function.

pamelafox (approved):

PR #618, opened 2023-09-10
Smooth streaming responses and ensure visibility

SteveJSteiner:
## Purpose

Allow the streamed output to become visible at a smoother cadence.  
Comparing the left (current streaming) and the right (smooth streaming) a few things are different:
For the first call the response packets arrive at relatively the same rate but ReAct is buffering and displaying them in chunks on the left and displaying them at a fixed rate on the right. Ideally, we want streaming to provide the response at a fixed rate that is faster than typical reading speed, but no faster.  (Zooming into the gif can help see the differences.)

This change includes a fix for streaming responses scrolling past the end of the screen.

SmoothScrolling

```
[ ] Yes
[ X] No
```

## Pull Request Type

```
[X] Bugfix
[X] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
azd up
```

* Test the code

```
Ask sufficient questions to ensure the chat scrolls off the screen.
```

## What to Check
* The text of responses streams smoothly, and the answers remain visible.  Test swapping streaming off and on


app/frontend/src/components/Answer/AnswerStreaming.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+    const sanitizedAnswerHtml = DOMPurify.sanitize(parsedAnswer.answerHtml);
+
+    return (
+        <Stack className={`${styles.answerContainer}`} verticalAlign="space-between">
+            <Stack.Item>
+                <Stack horizontal horizontalAlign="space-between">
+                    <AnswerIcon />
+                    <div>
+                        <IconButton
```
pamelafox, line 28: Feels like these buttons could just be an IconButtonToolbar component that could be enabled or disabled? That'd be less code repetition.

pamelafox:
I like the effect! I'm a fan of less code repetition, which would mean moving some answer-related UI parts into reusable components for both Answer.tsx and AnswerStreaming.tsx, but there's also something to be said for making it easy for folks to just delete what they're not using. So I am going to defer to @chuwik as I know he also has more recent React experience than me.

chuwik:
> I like the effect! I'm a fan of less code repetition, which would mean moving some answer-related UI parts into reusable components for both Answer.tsx and AnswerStreaming.tsx, but there's also something to be said for making it easy for folks to just delete what they're not using. So I am going to defer to @chuwik as I know he also has more recent React experience than me.

+1 to @pamelafox's thoughts. @SteveJSteiner could you extract the common UI functionality, to avoid duplication?

The effect looks really cool :)

SteveJSteiner:
Happy to do what makes sense implementation wise.

I’m no react expert so don’t hesitate to provide more pointed feedback if my updates are off the beaten path.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

chuwik (approved):

pamelafox:
@SteveJSteiner I checked out the branch locally and noticed one UI issue: now the frontend renders citations before they're fully done (i.e. before the ending bracket), so you'll see a [BenefitOptions flask briefly before turning into a footnote. Did you look into the complexity of avoiding those flashes? A naive approach would be the avoid rendering an open square bracket before seeing its close square bracket, but it seems possible there are some answers there that are using square brackets for other reasons.

SteveJSteiner:
@pamelafox - Thanks! I've pushed some code that gets rid of the flashing on partial citations.
  
I was confused for a bit as I had thought the earlier approval was actually a notification it had merged.

srbalakr:
@SteveJSteiner  I see lot of changes w.r.t formatting and adding Open ai support. was this intended ? Could you please remove them from your PR if planning to get the streaming changes merged ?

SteveJSteiner:
No not intended.  I had thought this was already merged, and had already pulled changes.  Apparently into the relevant branch.

Perhaps a new PR would be simpler?

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

SteveJSteiner:
I've opened PR #659 which is omits the accidentally merged files from main, and has the updates from the feedback so far in this PR.  .. Absent objection I'll close this one and we can continue there.  @pamelafox, @chuwik, @srbalakr

Issue #541, opened 2023-08-17
Langchain based ask approaches not compatible with 0613 (Chat Completions)


pamelafox:
### This issue is for a: (mark with an `x`)
```
- [X] bug report -> please search issues before submitting
- [ ] feature request
- [ ] documentation issue or request
- [ ] regression (a behavior that used to work and stopped in a new release)
```

### Minimal steps to reproduce

We recently changed to version from 0301 to 0613, since Azure isn't allowing new 0301 deployments. Unfortunately, 0613 only supports the new Chat Completions API, not the old Completions API, and the LangChain agents all assume use of the Completions API.

I have a branch that attempts to update the LangChain code to use Chat Completions, but am still QAing it.

aparnasharmav:
Hi @pamelafox : Will be waiting for an update on this, we are trying to deploy it in a corporate scenario, I managed to build an ARM template from the bicep template you have shared and removed role permissions required, I am using 2 models 1. Chat Gpt 35 turbo with version 0301 and chat gpt 35 turbo 16k with 0613 to deploy this but fail to deploy the template with an error saying 'standard' is not part of the 0613 version, if I remove the scale standard and deploy it fails the validation.

pamelafox:
To clarify:

The sample includes 4 different RAG (Retrieval-Augmented Generation) approaches: ChatReadRetrieveRead, ReadDecomposeAsk, ReadRetrieveRead, RetrieveThenRead. The two default approaches are ChatReadRetrieveRead and RetrieveThenRead, and they are both working very well with the Chat Completion APIs. The other two approaches use Langchain and the current code only works with the older Completion API (0301). Those approaches can be deleted from the code/UI, and the app would still work.

Is the problem that you definitely need to use those other two approaches for your particular use case, or is the problem that you can't deploy 0613? Do you have the latest main.bicep and cognitiveservices.bicep? The method of specifying capacity changed a few months ago.

aparnasharmav:
Hi Pamela, Thank you for getting back really quickly. Our org currently doesn't allow a bicep file download (.exe file download), so I had to convert the bicep file into an ARM template, remove permissions and roles, and then deploy. I am attaching the ARM template. Regarding your note on the 4 RAG approaches I might just need the first two, how do I modify the code, the app backend code in Python that I need to modify? Where can I get more information about this as the readme doesn't detail this? Lastly, deploy to Azure functionality was super useful on other examples can we expect something like that for this?
[...]

PR #624, opened 2023-09-13
Add support for an optional login and document level access control system.

mattmsft:
## Purpose

* Add an optional login and document level access control system
* Support for access control based on the User Object ID and Group ID
* Add optional login UI to the frontend components
* Add optional authentication tokens and authentication understanding capabilities to the backend
* Use security filters in Azure Cognitive Search to implement the access control system.

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[ ] Bugfix
[X] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

* Test the code
1. Follow the manual setup instructions in `LoginAndAclSetup.md`.
1. After running `azd up` or running locally, a login button will appear at the top right. Use this to log into your account in your Azure AD tenant.
1. Enable the optional `oids` or `groups` security filters in the developer settings. Sample access control can be setup by setting `AZURE_ADLS_GEN2_STORAGE_ACCOUNT` to a valid data lake storage account prior to running `azd up` or running `prepdocs.py`. Access control values can also be manually managed using the `manageacl.ps1` script.


LoginAndAclSetup.md
```diff
@@ -0,0 +1,160 @@
...
+  1. Select one of the available key durations.
+  1. The generated key value will be displayed after you select **Add**.
+  1. Copy the generated ke value and run the following `azd` command to save this ID: `azd env set AZURE_SERVER_APP_SECRET <generated key value>`.
+1. Select **API Permissions** in the left hand menu. By default, the [delegated `User.Read`](https://learn.microsoft.com/graph/permissions-reference#user-permissions) permission should be present. This permission is required to read the signed-in user's profile to get the security information used for document level access control. If this permission is not present, it needs to be added to the application.
+  1. Select **Add a permission**, and then **Microsoft Graph**.
+  1. Select **Delegated permissions**.
+  1. Search for and and select `User.Read`.
+  1. Select **Add permissions**.
+1. Select **Expose an API** in the left hand menu. The server app works by using the [https://learn.microsoft.com/azure/active-directory/develop/v2-oauth2-on-behalf-of-flow#protocol-diagram](On-Behalf-Of Flow), which requires the server app to expose at least 1 API.
```
mattmsft, line 56: this link does not render properly. Fix using github markdown syntax

LoginAndAclSetup.md
```diff
@@ -0,0 +1,160 @@
...
+  1. In the **Name** section, enter a meaningful application name. This name will be displayed to users of the app, for example `Azure Search OpenAI Demo Web App`.
+  1. Under **Supported account types**, select **Accounts in this organizational directory only**.
+  1. Under `Redirect URI (optional)` section, select `Single-page application (SPA)` in the combo-box and enter the following redirect URI:
+    1. If you are running the sample locally, use `http://localhost:50505/redirect`.
+    1. If you are running the sample, use the endpoint provided by `azd up`: `https://<your-endpoint>.azurewebsites.net/redirect`.
+1. Select **Register** to create the application
+1. In the app's registration screen, find the **Application (client) ID**.
+  1. Run the following `azd` command to save this ID: `azd env set AZURE_CLIENT_APP_ID <Application (client) ID>`.
+1. In the left hand menu, select **API permissions**. You will add permission to access the **access_as_user** API on the server app. This permission is required for the [https://learn.microsoft.com/azure/active-directory/develop/v2-oauth2-on-behalf-of-flow#protocol-diagram](On-Behalf-Of Flow) to work.
```
mattmsft, line 84: this link does not render properly. Fix using github markdown syntax

app/backend/app.py
```diff
@@ -43,6 +45,10 @@
 async def index():
     return await bp.send_static_file("index.html")
 
+@bp.route("/redirect")
+async def redirect():
```
mattmsft, line 49: comment - empty page required for redirect to work correctly.

app/backend/app.py
```diff
@@ -116,19 +124,24 @@ async def chat_stream():
...
+        response_generator = impl.run_with_streaming(request_json["history"], request_json.get("overrides", {}), auth_claims)
         response = await make_response(format_as_ndjson(response_generator))
         response.timeout = None # type: ignore
         return response
     except Exception as e:
         logging.exception("Exception in /chat")
         return jsonify({"error": str(e)}), 500
 
+@bp.route("/auth_setup", methods=["GET"])
+def auth_setup():
```
mattmsft, line 142: comment - MSAL.js settings for the client application.

app/backend/core/authentication.py
```diff
@@ -0,0 +1,164 @@
...
+
+from quart import request
+from abc import ABC
+import msal, os
+from msgraph import GraphServiceClient
+from azure.core.credentials import TokenCredential, AccessToken
+import logging
+from typing import Any
+
+class AuthError(Exception):
```
mattmsft, line 11: comment - exception thrown when authentication can't be parsed

README.md
```diff
@@ -180,6 +180,10 @@ By default, the deployed Azure web app will have no authentication or access res
 
 To then limit access to a specific set of users or groups, you can follow the steps from [Restrict your Azure AD app to a set of users](https://learn.microsoft.com/azure/active-directory/develop/howto-restrict-your-app-to-a-set-of-users) by changing "Assignment Required?" option under the Enterprise Application, and then assigning users/groups access.  Users not granted explicit access will receive the error message -AADSTS50105: Your administrator has configured the application <app_name> to block users unless they are specifically granted ('assigned') access to the application.-
 
+### Enabling optional login and document level access control
+
+By default, the deployed Azure web app allows users to chat with all your indexed data. You can enable an optional login system using Azure Active Directory to restrict access to indexed data based on the logged in user. Enable the optional login and document level access control system by following [this guide](./LoginAndAclSetup.md).
```
mattmsft, line 185: Since this repository was created there are now tests. See if you can add some tests for this

app/backend/core/authentication.py
```diff
@@ -0,0 +1,164 @@
+# Refactored from https://github.com/Azure-Samples/ms-identity-python-on-behalf-of
+
+from quart import request
+from abc import ABC
+import msal, os
+from msgraph import GraphServiceClient
+from azure.core.credentials import TokenCredential, AccessToken
+import logging
+from typing import Any
```
pamelafox, line 9: Can you run ruff and black on your code, per CONTRIBUTING.md?

app/backend/requirements.txt
```diff
@@ -11,3 +11,5 @@ azure-monitor-opentelemetry==1.0.0b15
 opentelemetry-instrumentation-asgi==0.40b0
 opentelemetry-instrumentation-requests==0.40b0
 opentelemetry-instrumentation-aiohttp-client==0.40b0
+msal==1.23.0
+msgraph-sdk==1.0.0a14
```
pamelafox, line 15: Their pypi says not to use the SDK in production environments, has that changed? Thats why my PR uses HTTP calls instead. https://pypi.org/project/msgraph-sdk/

app/frontend/src/components/LoginButton/LoginButton.tsx
```diff
@@ -0,0 +1,40 @@
...
+       * For more information, please follow this link: https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-browser/docs/login-user.md#redirecturi-considerations
+       */
+      instance
+          .loginPopup({
+              ...loginRequest,
+              redirectUri: '/redirect',
+          })
+          .catch((error) => console.log(error));
+  };
+  const handleLogoutPopup = () => {
```
mattmsft, line 24: is there a way to wait until the user logs out using the popup? if i X out it "logs out", but when i log in it doesn't prompt me again to log in, it just logs in using the last account automatically.

app/backend/core/authentication.py
```diff
@@ -0,0 +1,164 @@
...
+        return all(var_name in os.environ for var_name in var_names)
+
+    @staticmethod
+    def get_auth_setup_for_client() -> dict[str, Any]:
+        # returns MSAL.js settings used by the client app
+        return {
+            "useLogin": AuthenticationHelper.use_authentication(), # Whether or not login elements are enabled on the UI
+            "msalConfig": {
+                "auth": {
+                    "clientId": os.getenv("AZURE_CLIENT_APP_ID"), # Client app id used for login
```
pamelafox, line 49: This code references a lot of env variables, which makes it harder to test. Could all configuration be passed in via class constructors or function args instead?

LoginAndAclSetup.md
```diff
@@ -0,0 +1,160 @@
+# Setting up optional login and document level access control
+
+## Table of Contents
+
+- [Requirements](#requirements)
+- [Setting up Azure AD Apps](#setting-up-azure-ad-apps)
+  - [Manual Setup](#manual-setup)
```
mattmsft, line 7: Unfortunately missing from this PR - automatic setup. It's quite the can of worms because it involves calling the graph API, but we are now pulling in the graph sdk which should make this simpler. Can automatic setup wait until a follow up PR? @pamelafox  has a similar script in her PR #534

app/backend/requirements.txt
mattmsft: this is good feedback - i can remove the graph SDK and page through the results

app/backend/core/authentication.py
```diff
@@ -0,0 +1,164 @@
...
+        elif len(parts) > 2:
+            raise AuthError({"code": "invalid_header",
+                            "description": "Authorization header must be Bearer token"}, 401)
+
+        token = parts[1]
+        return token
+
+    @staticmethod
+    def get_confidential_client():
+        # This example only uses the default memory token cache and should not be used for production
```
pamelafox, line 95: It says this shouldnt be used for production?

app/backend/core/authentication.py
```diff
@@ -0,0 +1,164 @@
+# Refactored from https://github.com/Azure-Samples/ms-identity-python-on-behalf-of
```
pamelafox, line 1: High-level observation: there's a lot going on in this code. Does any of this belong in a different SDK? And did you check if https://github.com/rayluo/identity does any of this for you? It's an msal abstraction for login.

app/backend/requirements.txt
```diff
@@ -11,3 +11,5 @@ azure-monitor-opentelemetry==1.0.0b15
 opentelemetry-instrumentation-asgi==0.40b0
```
mattmsft, line 11: somehow i didn't update the blob SDK in this file - update this

app/frontend/src/components/LoginButton/LoginButton.module.css
```diff
@@ -0,0 +1,7 @@
+.loginButton {
+    border-radius: 5px;
+    padding: 30px 30px;
+    font-weight: 100;
+    /*font-size: 16px;*/
```
pamelafox, line 5: Rm commented code

app/frontend/src/components/LoginButton/LoginButton.module.css
```diff
@@ -0,0 +1,7 @@
...
+    /*box-shadow: 0px 2px 4px rgba(0, 0, 0, 0.2);*/
+}
```
pamelafox, line 7: Run the pre-commit hooks to take care of whitespace/EOF issues https://github.com/Azure-Samples/azure-search-openai-demo/blob/main/.pre-commit-config.yaml

app/frontend/src/authConfig.ts
```diff
@@ -0,0 +1,70 @@
+import { AuthenticationResult, IPublicClientApplication } from "@azure/msal-browser";
+
+interface AuthSetup {
```
pamelafox, line 3: Was this pasted from somehere? (Given all the comments) It's fine if so, but may be useful to note.

app/start.ps1
```diff
@@ -65,7 +65,7 @@ Write-Host "Starting backend"
 Write-Host ""
 Set-Location ../backend
 
-Start-Process -FilePath $venvPythonPath -ArgumentList "-m quart --app main:app run --port 50505 --reload" -Wait -NoNewWindow
+Start-Process -FilePath $venvPythonPath -ArgumentList "-m quart --app main:app run --port 50505 --host localhost --reload" -Wait -NoNewWindow
```
pamelafox, line 68: Will this work alright in codespaces?

scripts/adlsgen2setup.ps1
```diff
@@ -0,0 +1,109 @@
+## Set the preference to stop on the first error
+$ErrorActionPreference = "Stop"
```
pamelafox, line 2: Can you write a sh equivalent? Or alternatively write most of this using the Azure Python SDK so that the sh/ps1 files are just small wrappers?

scripts/manageacl.py
```diff
@@ -0,0 +1,90 @@
...
+        if not any(field.name == 'groups' for field in index_definition.fields):
+            index_definition.fields.append(SimpleField(name="groups", type=SearchFieldDataType.Collection(SearchFieldDataType.String), filterable=True))
+        
+        index_client.create_or_update_index(index_definition)
+    except Exception as e:
+        print(f"\tGot an error while updating index {index} -> {e} --> unable to enable acls")
+
+def update_acl(credential: AzureKeyCredential|TokenCredential, search_service: str, index: str, acl_type: str, acl_action: str, document: str, acl: str = None):
+    search_client = SearchClient(
+        endpoint=f"https://{args.search_service}.search.windows.net/",
```
pamelafox, line 29: Please avoid using args in the functions as it makes code much harder to test (prepdocs does this a bunch now, and its made it hard to write tests). Please pass as args or create a class which takes in the configuration if there are too many for args.

LoginAndAclSetup.md
```diff
@@ -0,0 +1,160 @@
...
+
+## Setting up Azure AD Apps
+
+Two Azure AD apps must be registered in order to make the optional login and document level access control system work correctly.
+
+### Manual Setup
+
+The following instructions explain how to setup the two apps using the Azure Portal.
+
+#### Setting up the Server App
```
pamelafox, line 34: I think my AAD Auth PR automates a lot of this. I prefer automation when possible as its a pain to setup app registrations and even more of a pain to debug the process (when an env variable isn't quite right). Let's join forces!

app/backend/core/authentication.py
mattmsft: this also has the "DISCLAIMER: The code in this repo is not officially supported by Microsoft and is not intended for production use. " in the description. I'm happy to look at it, but i used an existing sample for this code

app/frontend/src/authConfig.ts
mattmsft: yes, it was pasted from an existing sample, it's possible the attribution got wiped here

app/start.ps1
mattmsft: if you don't specify localhost the redirect uri will not work. So perhaps we should add a disclaimer this won't work in codespaces at all.

scripts/adlsgen2setup.ps1
mattmsft: sure, i can take a look at doing this

scripts/manageacl.py
mattmsft: Do you have an example of how this should be done? i'm a bit confused as how i should write this, there's an argparse at the beginning of the script and then smaller functions.

app/backend/core/authentication.py
pamelafox: Oh, interesting, we changed the learn.com tutorials to use that repo. I'll ask Ray the status of that.

LoginAndAclSetup.md
mattmsft: We should do automation, agreed. The problem is setting up the API permissions on the apps, we need to fill in some of the body to the API calls with a bunch of manifest settings. and then we need to go back after the app is provisioned and setup the redirect uris correctly

app/frontend/src/components/LoginButton/LoginButton.module.css
mattmsft: ran black and ruff on all the code

app/backend/core/authentication.py
mattmsft: this was pasted in from the sample - i can use a token cache somewhere else

app/backend/core/authentication.py
mattmsft: i took a brief look at this and did not see the functionality that i needed (claims parsing, on behalf of flow), but we can look again if required.

LoginAndAclSetup.md
pamelafox: I don't think my branch does the manifest setting, but it does include a postprovision hook to setup the redirect URIs. I can look into the manifest setting automation.

pamelafox:
Very exciting, one of the top requests for this repo! Actually, two of the top requests! 🎉

app/backend/requirements.txt
mattmsft: sdk removed, replaced with urllib3 usage similar to your PR

app/backend/core/authentication.py
mattmsft: found one in the msal-extensions library

mattmsft: switched to a persisted encrypted token cache

app/frontend/src/authConfig.ts
mattmsft: at this point it's been heavily refactored (auth setup shared from the server), but the original sample is really useful so i added a lnk

app/start.ps1
mattmsft: @pamelafox is there an env var i can check to avoid specifying this in codespaces?

scripts/manageacl.py
mattmsft: making these adjustments

LoginAndAclSetup.md
mattmsft: we spoke offline - let's get this in a follow up pr

scripts/adlsgen2setup.ps1
mattmsft: writing in python, writing sh file

app/frontend/src/components/LoginButton/LoginButton.tsx
mattmsft: fyi @chuwik , @srbalakr is it OK if i file a followup issue, the login button can be a bit flaky with the popup. it mostly works though

app/start.ps1
mattmsft: testing the fix for this

mattmsft: investigated - this is actually running the app so localhost is ok. Manual test of codespaces to see if hosting still works

app/frontend/src/components/LoginButton/LoginButton.tsx
srbalakr: Let me add a internal work item to track.

LoginAndAclSetup.md
mattmsft: follow up pr

app/frontend/vite.config.ts
```diff
@@ -20,7 +20,8 @@ export default defineConfig({
                     }
                 }
             }
-        }
+        },
+        target: "esnext"
```
pamelafox, line 24: I'm curious why this change was needed? (Using more recent syntax?)

docs/applogincomponents.png
pamelafox, line 1: I find the current image and this one misleading since the repo doesnt support databases yet, what do you think of just showing the files?

scripts/adlsgen2setup.ps1
```diff
@@ -0,0 +1,47 @@
...
+
+Write-Host ""
+Write-Host "Loading azd .env file from current environment"
+Write-Host ""
+
+$output = azd env get-values
+
+foreach ($line in $output) {
```
pamelafox, line 10: As youre now repeating this code in a few files, can you do what I did in the AAD PR and move into a loadenv.sh and loadenv.ps1 script?

scripts/adlsgen2setup.py
```diff
@@ -0,0 +1,166 @@
...
+        self.filesystem_name = filesystem_name
+        self.credentials = credentials
+        self.security_enabled_groups = security_enabled_groups
+        self.verbose = verbose
+
+    async def run(self):
+        token_result = await self.credentials.get_token("https://graph.microsoft.com/.default")
+        self.graph_headers = {"Authorization": f"Bearer {token_result.token}"}
+        if self.verbose:
+            print("Ensuring gptkbcontainer exists...")
```
pamelafox, line 45: I think it'd be nicer to move away from print() statements and use logging module instead, these could be logging.info() or logging.debug(). Then you can set overall logging level to decide when to see them.  (And if you ran this on a server, it'd be easier to pipe logs somewhere)

scripts/adlsgen2setup.py
```diff
@@ -0,0 +1,166 @@
...
+                await filesystem_client.create_file_system()
+
+            if self.verbose:
+                print("Creating groups...")
+            admin_id = await self.create_or_get_group("GPTKB_AdminTest")
+            hr_id = await self.create_or_get_group("GPTKB_HRTest")
+            employee_id = await self.create_or_get_group("GPTKB_EmployeeTest")
+
+            if self.verbose:
+                print("Ensuring benefitinfo and employeeinfo directories exist...")
```
pamelafox, line 57: Are these hardcoded for this example? How would someone customize?

scripts/adlsgen2setup.py
```diff
@@ -0,0 +1,166 @@
...
+            async with await filesystem_client.create_directory(
+                "benefitinfo"
+            ) as benefit_info, await filesystem_client.create_directory("employeeinfo") as employee_info:
+                if self.verbose:
+                    print("Uploading PDFs...")
+                await self.upload_file(
```
pamelafox, line 63: Oh yeah this is definitely hardcoded - people change up these files a lot so I would like something more generic if possible.

scripts/manageacl.py
```diff
@@ -0,0 +1,158 @@
...
+from azure.identity import AzureDeveloperCliCredential
+from azure.search.documents.aio import SearchClient
+from azure.search.documents.indexes.aio import SearchIndexClient
+from azure.search.documents.indexes.models import SearchFieldDataType, SimpleField
+
+class ManageAcl:
+    async def __aenter__(self):
+        endpoint = f"https://{self.service_name}.search.windows.net"
+        self.search_client = SearchClient(endpoint=endpoint, credential=self.credentials, index_name=self.index_name)
+        await self.search_client.__aenter__()
```
pamelafox, line 16: it might look nicer to use "async with search_client" here? We usually avoid direct calling dunders when possible, prefer syntactic sugar :)

scripts/manageacl.py
```diff
@@ -0,0 +1,158 @@
...
+        self.search_index_client = SearchIndexClient(endpoint=endpoint, credential=self.credentials)
+        await self.search_index_client.__aexit__()
+        return self
+
+    async def __aexit__(self, *args, **kwargs):
+        await self.search_client.__aexit__(*args, **kwargs)
+        await self.search_index_client.__aexit__(*args, **kwargs)
```
pamelafox, line 23: why does it exit search_index_client twice? (here and above?)

scripts/manageacl.py
```diff
@@ -0,0 +1,158 @@
...
+from enum import Enum
+
+from azure.core.credentials import AzureKeyCredential
+from azure.core.credentials_async import AsyncTokenCredential
+from azure.identity import AzureDeveloperCliCredential
+from azure.search.documents.aio import SearchClient
+from azure.search.documents.indexes.aio import SearchIndexClient
+from azure.search.documents.indexes.models import SearchFieldDataType, SimpleField
+
+class ManageAcl:
```
pamelafox, line 12: Can you add module or class docstrings to explain what this does at a high-level?

scripts/manageacl.py
```diff
@@ -0,0 +1,158 @@
...
+            elif self.acl_action == "add":
+                new_acls = document[self.acl_type]
+                if not any(acl_value == self.acl for acl_value in new_acls):
+                    new_acls.append(self.acl)
+            else:
+                new_acls = []
+            documents_to_merge.append({"id": document["id"], self.acl_type: new_acls})
+
+        await self.search_client.merge_documents(documents=documents_to_merge)
+        if self.verbose:
```
pamelafox, line 93: Ditto my comment re logging. Found nice example of how azure sdk handles it for scripts: https://github.com/Azure/azure-sdk-for-python/blob/6f168b61101bbf851a30c085747d51470db4be0c/tools/azure-sdk-tools/packaging_tools/generate_package.py#L46

pamelafox (changes requested):
Just a high level ask: will you or the team be writing blog posts or making videos about that? The vector search blog post was super helpful when that feature was added. Also curious why datalake is the chosen storage mechanism vs other options.

pamelafox:
Another ask: could you fill out PR description? Just expanding a bit on the PR title to describe the front and backend features added.

pamelafox:
Oh and looks like ruff is failing on CI, some errors.

app/frontend/vite.config.ts
mattmsft: yes, top level await. I actually believe the intent was to use esnext because it's present in the node files

scripts/adlsgen2setup.ps1
mattmsft: Great idea

scripts/adlsgen2setup.py
mattmsft: The problem with the access control lists are that they are specific to the user. What I can do is add a helper function that takes a list of files / their acls rather than hardcoding it this way, and take some info as parameters. The hardcoded behavior is useful for demonstrating a complete flow with the example, let me know if you disagree.

scripts/manageacl.py
mattmsft: this was a mistake - but i'll remove this per your other feedback

mattmsft:
> Just a high level ask: will you or the team be writing blog posts or making videos about that? The vector search blog post was super helpful when that feature was added. Also curious why datalake is the chosen storage mechanism vs other options.

yes, I'll be writing a blog about this one

scripts/adlsgen2setup.py
pamelafox: Ah okay, could you make it so that the class/function accepts a data type (dict or list) and then for the demo, we pass in the data either via shell script args, acls.csv or acl.json file?

app/backend/core/authentication.py
```diff
@@ -0,0 +1,202 @@
...
+                groups.append(group["id"])
+            nextLink = resp.json().get("@odata.nextLink")
+            if nextLink:
+                resp = urllib3.request("GET", nextLink, headers=headers, timeout=urllib3.Timeout(connect=10, read=10))
+            else:
+                break
+
+        return groups
+
+    async def get_auth_claims_if_enabled(self) -> dict[str:Any]:
```
pamelafox, line 165: Just curious, is this code similar to https://github.com/snok/django-auth-adfs/blob/cc700730a4130365dbd6094612f1cf57ad2a3d6f/django_auth_adfs/backend.py#L196 ?
Might be useful to compare your approach if so, I know you mentioned there was some trickiness to group claim checking.

mattmsft: my approach is pretty much the same. the only difference is they are using the transitive member of, which I can switch to

scripts/adlsgen2setup.ps1
mattmsft: copied from your PR

app/backend/core/authentication.py
mattmsft: i'll keep it member of for now

mattmsft: made the change

LoginAndAclSetup.md
```diff
@@ -0,0 +1,178 @@
...
+
+This guide demonstrates how to add an optional login and document level access control system to the sample. This system can be used to restrict access to indexed data to specific users based on what [Azure Active Directory (Azure AD) groups](https://learn.microsoft.com/azure/active-directory/fundamentals/how-to-manage-groups) they are a part of, or their [user object id](https://learn.microsoft.com/partner-center/find-ids-and-domain-names#find-the-user-object-id).
+
+![AppLoginArchitecture](./docs/applogincomponents.png)
+
+## Requirements
+
+**IMPORTANT:** In order to add optional login and document level access control, you'll need the following in addition to the normal sample requirements
+
+* **Azure account permissions**: Your Azure account must have [permission to manage applications in Azure AD](https://learn.microsoft.com/azure/active-directory/develop/quickstart-register-app#prerequisites).
```
pamelafox, line 24: Would it be better to directly link to https://learn.microsoft.com/en-us/azure/active-directory/roles/permissions-reference#cloud-application-administrator ? That's what we do for the other roles in the main README

LoginAndAclSetup.md
```diff
@@ -0,0 +1,178 @@
...
+
+## Setting up Azure AD Apps
+
+Two Azure AD apps must be registered in order to make the optional login and document level access control system work correctly.
```
pamelafox, line 28: Is there a short explanation of why 2 apps are needed? I only needed one for mine.

LoginAndAclSetup.md
```diff
@@ -0,0 +1,178 @@
...
+* Sign in to the [Azure portal](https://portal.azure.com/).
+* Select the Azure AD Service.
+* In the left hand menu, select **Application Registrations**.
+* Select **New Registration**.
+  * In the **Name** section, enter a meaningful application name. This name will be displayed to users of the app, for example `Azure Search OpenAI Demo API`.
+  * Under **Supported account types**, select **Accounts in this organizational directory only**.
+* Select **Register** to create the application
+* In the app's registration screen, find the **Application (client) ID**.
+  * Run the following `azd` command to save this ID: `azd env set AZURE_SERVER_APP_ID <Application (client) ID>`.
+* Select **Certificates & secrets** in the left hand menu.
```
pamelafox, line 45: It will only be possible in Bicep to use certificates, and the docs all discourage the use of secrets, so I wonder if we should encourage the (admittingly more difficult) route of using a certificate here. Otherwise it'll be hard for folks to upgrade.

https://learn.microsoft.com/en-us/graph/applications-how-to-add-certificate

"Azure Active Directory (Azure AD) supports three types of credentials to authenticate apps and service principals: passwords (app secrets), certificates, and federated identity credentials. If you can't use federated identity credentials for your app, we strongly recommend that you use certificates instead of secrets."

The easiest way to get started locally with certificates is self-signed certificates, though those are also discouraged for production.

mattmsft: So this makes it more challenging - I'll see if there's an easy way to add key vault support here. Are you OK leaving in the password for now and creating a follow up issue?

LoginAndAclSetup.md
```diff
@@ -0,0 +1,178 @@
...
+
+* Sign in to the [Azure portal](https://portal.azure.com/).
+* Select the Azure AD Service.
+* In the left hand menu, select **Application Registrations**.
+* Select **New Registration**.
+  * In the **Name** section, enter a meaningful application name. This name will be displayed to users of the app, for example `Azure Search OpenAI Demo Web App`.
+  * Under **Supported account types**, select **Accounts in this organizational directory only**.
+  * Under `Redirect URI (optional)` section, select `Single-page application (SPA)` in the combo-box and enter the following redirect URI:
+    * If you are running the sample locally, use `http://localhost:50505/redirect`.
+    * If you are running the sample, use the endpoint provided by `azd up`: `https://<your-endpoint>.azurewebsites.net/redirect`.
```
pamelafox, line 86:
```suggestion
    * If you are running the sample on Azure, use the endpoint provided by `azd up`: `https://<your-endpoint>.azurewebsites.net/redirect`.
```

LoginAndAclSetup.md
```diff
@@ -0,0 +1,178 @@
...
+    * If you are running the sample from codespaces, use the codespaces endpoint: `https://<your-codespace>-50505.app.github.dev/`
```
pamelafox, line 87:
```suggestion
    * If you are running the sample from GitHub Codespaces, use the Codespaces endpoint: `https://<your-codespace>-50505.app.github.dev/`
```

LoginAndAclSetup.md
mattmsft: I am planning on going into more detail in the blog post - but the reason there's only 1 app needed for yours is that you are putting the login in front of the web app.

There's 2 apps here because the client UI is a Single Page Application and supports the login. This needs its own app. The API needs its own app to call the graph API

LoginAndAclSetup.md
```diff
@@ -0,0 +1,178 @@
...
+* In the left hand menu, select **API permissions**. You will add permission to access the **access_as_user** API on the server app. This permission is required for the [On Behalf Of Flow](https://learn.microsoft.com/azure/active-directory/develop/v2-oauth2-on-behalf-of-flow#protocol-diagram) to work.
+  * Select **Add a permission**, and then **My APIs**.
+  * In the list of applications, select your server application **Azure Search OpenAI Demo API**
+  * Ensure **Delegated permissions** is selected.
+  * In the **Select permissions** section, select the **access_as_user** permission
+  * Select **Add permissions**.
+
+#### Configure Server App Known Client Applications
+
+Consent from the user must be obtained for use of the client and server app. The client app can prompt the user for consent through a dialog when they log in. The server app has no ability to show a dialog for consent. Client apps can be [added to the list of known clients](https://learn.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-on-behalf-of-flow#gaining-consent-for-the-middle-tier-application) to access the server app, so a consent dialog is shown for the server app.
```
pamelafox, line 103:
```suggestion
Consent from the user must be obtained for use of the client and server app. The client app can prompt the user for consent through a dialog when they log in. The server app has no ability to show a dialog for consent. Client apps can be [added to the list of known clients](https://learn.microsoft.com/azure/active-directory/develop/v2-oauth2-on-behalf-of-flow#gaining-consent-for-the-middle-tier-application) to access the server app, so a consent dialog is shown for the server app.
```

LoginAndAclSetup.md
```diff
@@ -0,0 +1,178 @@
...
+* Navigate to the server app registration
+* In the left hand menu, select **Manifest**
+* Replace `"knownClientApplications": []` with `"knownClientApplications": ["<client application id>"]`
+* Select **Save**
+
+#### Testing
+
+Ensure you run `azd env set AZURE_USE_AUTHENTICATION` to enable the login UI once you have setup the two Azure AD apps before you deploy or run the application. The login UI will not appear unless all [required environment variables](#environment-variables-reference) have been setup.
+
+In both the chat and ask a question modes, an optional **Use oid security filter** and **Use groups security filter** will appear. The oid (User ID) filter maps to the `oids` field and the groups (Group ID) filter maps to the `groups` field in the search index. Use the optional scripts included in the sample to manage values for these fields.
```
pamelafox, line 114:
```suggestion
In both the chat and ask a question modes, an optional **Use oid security filter** and **Use groups security filter** will appear. The oid (User ID) filter maps to the `oids` field in the search index and the groups (Group ID) filter maps to the `groups` field. Use the optional scripts included in the sample to manage values for these fields.
```

LoginAndAclSetup.md
```diff
@@ -0,0 +1,178 @@
...
+
+It's possible that your tenant admin has placed a restriction on consent to apps with [unverified publishers](https://learn.microsoft.com/azure/active-directory/develop/publisher-verification-overview). In this case, only admins may consent to the client and server apps, and normal user accounts are unable to use the login system until the admin consents on behalf of the entire organization.
+
+## Optional scripts
+
+Two optional scripts are provided that allow easier setup of sample data with document level access control.
+
+### Azure Data Lake Storage Gen2 Setup
+
+[Azure Data Lake Storage Gen2](https://learn.microsoft.com/en-us/azure/storage/blobs/data-lake-storage-introduction) implements an [access control model](https://learn.microsoft.com/en-us/azure/storage/blobs/data-lake-storage-access-control) that can be used for document level access control. The [adlsgen2setup.ps1](./scripts/adlsgen2setup.ps1) script uploads the sample data included in the [data](./data) folder to a Data Lake Storage Gen2 storage account.
```
pamelafox, line 124:
```suggestion
[Azure Data Lake Storage Gen2](https://learn.microsoft.com/azure/storage/blobs/data-lake-storage-introduction) implements an [access control model](https://learn.microsoft.com/azure/storage/blobs/data-lake-storage-access-control) that can be used for document level access control. The [adlsgen2setup.ps1](./scripts/adlsgen2setup.ps1) script uploads the sample data included in the [data](./data) folder to a Data Lake Storage Gen2 storage account.
```

LoginAndAclSetup.md
```diff
@@ -0,0 +1,178 @@
...
+
+In order to use this script, an existing Data Lake Storage Gen2 storage account. Run `azd env set AZURE_ADLS_GEN2_STORAGE_ACCOUNT <your-storage-account>` prior to running the script.
```
pamelafox, line 126: First sentence is missing a verb somewhere?

pamelafox, line 126: Do you need to set up roles/permissions to be able to access the account locally?

LoginAndAclSetup.md
```diff
@@ -0,0 +1,178 @@
...
+* Creates two directories, `benefitinfo` and `employeeinfo` in the `gptkbcontainer` filesystem / container.
+* Uploads the sample PDFs into both directories.
+* [Recursively sets Access Control Lists (ACLs)](https://learn.microsoft.com/azure/storage/blobs/data-lake-storage-acl-cli) on the `benefitinfo` and `employeeinfo` for the following groups:
+  * `GPTKB_AdminTest`: Can read all files in `gptkbcontainer`.
+  * `GPTKB_EmployeeTest`: Can only read files in `employeeinfo`.
+  * `GPTKB_HRTest`: Can read files in both `employeeinfo` and `benefitinfo`.
+
+In order to use the sample access control, you need to join these groups in your Azure AD tenant.
+
+Note that this optional scripts may not work in codespaces if your administrator has applied a [Conditional Access policy](https://learn.microsoft.com/azure/active-directory/conditional-access/overview) to your tenant.
```
pamelafox, line 140:
```suggestion
Note that this optional scripts may not work in Codespaces if your administrator has applied a [Conditional Access policy](https://learn.microsoft.com/azure/active-directory/conditional-access/overview) to your tenant.
```

LoginAndAclSetup.md
```diff
@@ -0,0 +1,178 @@
...
+
+### Azure Data Lake Storage Gen2 Prep Docs
+
+Once a Data Lake Storage Gen2 storage account has been setup with sample data and access control lists, the [prepdocs.ps1](./scripts/prepdocs.ps1) can be used to automatically process PDFs in the storage account and store them with their [access control lists in the search index](https://learn.microsoft.com/azure/storage/blobs/data-lake-storage-access-control).
```
pamelafox, line 144:
```suggestion
Once a Data Lake Storage Gen2 storage account has been setup with sample data and access control lists, [prepdocs.py](./scripts/prepdocs.py) can be used to automatically process PDFs in the storage account and store them with their [access control lists in the search index](https://learn.microsoft.com/azure/storage/blobs/data-lake-storage-access-control).
```

LoginAndAclSetup.md
```diff
@@ -0,0 +1,178 @@
...
+
+* `AZURE_ADLS_GEN2_STORAGE_ACCOUNT`: Name of existing [Data Lake Storage Gen2 storage account](https://learn.microsoft.com/azure/storage/blobs/data-lake-storage-introduction).
+* (Optional) `AZURE_ADLS_GEN2_FILESYSTEM`: Name of existing Data Lake Storage Gen2 filesystem / container in the storage account. If empty, `gptkbcontainer` is used.
+* (Optional) `AZURE_ADLS_GEN2_FILESYSTEM_PATH`: Specific path in the Data Lake Storage Gen2 filesystem / container to process. Only PDFs contained in this path will be processed.
+
+Once the environment variables are set, run the script using the following command: `./scripts/prepdocs.ps1` or `./scripts/prepdocs.sh`.
+
+### Manually managing Document Level Access Control
+
+Manually enable document level access control on a search index and manually set access control values using the [manageacl.ps1](./scripts/manageacl.ps1) script.
```
pamelafox, line 156: Is this an alternative to the data lake approach? It might be worth mentioning that at the beginning, just because it's easy to lose track of that in this long readme.

LoginAndAclSetup.md
```diff
@@ -0,0 +1,178 @@
...
+
+Run `azd up` or manually set the `AZURE_SEARCH_SERVICE` and `AZURE_SEARCH_INDEX` environment variables prior to running the script.
```
pamelafox, line 158: Assuming these need to be set as azd env vars -

```suggestion
Run `azd up` or use `azd env set` to manually set `AZURE_SEARCH_SERVICE` and `AZURE_SEARCH_INDEX` prior to running the script.
```

LoginAndAclSetup.md
```diff
@@ -0,0 +1,178 @@
...
+
+To run the script, the following parameters are used:
```
pamelafox, line 160: Is this meant to be a parameters reference, or stepbystep of which commands to use? A reference is helpful, but it might also be useful to show an example usage that enables ACLs and adds ACLs for a document, and to clarify that this section is a reference.

LoginAndAclSetup.md
```diff
@@ -0,0 +1,178 @@
...
+* `./scripts/manageacls.ps1 --document [name-of-pdf.pdf] --acl-type [oids or groups]--acl-action remove --acl [ID of group or user]`: Removes an access control value associated with either User IDs or Group IDs for a specific document. Example to remove a specific User ID from the Benefit_Options PDF: `./scripts/manageacls.ps1 --document Benefit_Options.pdf --acl-type oids --acl-action remove --acl xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx`.
+
+## Environment Variables Reference
+
+The following environment variables are used to setup the optional login and document level access control:
+
+* `AZURE_USE_AUTHENTICATION`: Enables Azure AD based optional login and document level access control. Set to true before running `azd up`.
+* `AZURE_SERVER_APP_ID`: (Required) Application ID of the Azure AD app for the API server.
+* `AZURE_SERVER_APP_SECRET`: [Client secret](https://learn.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-client-creds-grant-flow) used by the API server to authenticate using the Azure AD API server app.
```
pamelafox, line 173: As I mentioned above, we've been encouraged to use certificates instead.

README.md
```diff
@@ -215,6 +215,10 @@ By default, the deployed Azure web app will have no authentication or access res
 
 To then limit access to a specific set of users or groups, you can follow the steps from [Restrict your Azure AD app to a set of users](https://learn.microsoft.com/azure/active-directory/develop/howto-restrict-your-app-to-a-set-of-users) by changing "Assignment Required?" option under the Enterprise Application, and then assigning users/groups access.  Users not granted explicit access will receive the error message -AADSTS50105: Your administrator has configured the application <app_name> to block users unless they are specifically granted ('assigned') access to the application.-
 
+### Enabling optional login and document level access control
```
pamelafox, line 218: Please add a link to the TOC

app/backend/app.py
pamelafox: I'm intrigued as to why this is empty, as I've typically had the redirect go to somewhere in previous samples, like 
https://github.com/Azure-Samples/ms-identity-python-webapp/blob/14a6be5ad1deeb38d89707e260620c0d5a4c9494/app.py#L38

How come it's different in this case? Does the user not land up on the /redirect themselves?

app/backend/approaches/chatreadretrieveread.py
```diff
@@ -69,14 +70,25 @@ def __init__(
...
+        auth_claims: dict[str, Any],
+        should_stream: bool = False,
     ) -> tuple:
         has_text = overrides.get("retrieval_mode") in ["text", "hybrid", None]
         has_vector = overrides.get("retrieval_mode") in ["vectors", "hybrid", None]
         use_semantic_captions = True if overrides.get("semantic_captions") and has_text else False
-        top = overrides.get("top") or 3
         exclude_category = overrides.get("exclude_category") or None
-        filter = "category ne '{}'".format(exclude_category.replace("'", "''")) if exclude_category else None
+        top = overrides.get("top") or 3
```
pamelafox, line 83: Just fixing this since it's changed in here-

```suggestion
        top = overrides.get("top", 3)
```

app/backend/approaches/chatreadretrieveread.py
```diff
@@ -69,14 +70,25 @@ def __init__(
...
+
+        security_filter = AuthenticationHelper.build_security_filters(overrides, auth_claims)
+        filters = []
+        if exclude_category:
+            filters.append("category ne '{}'".format(exclude_category.replace("'", "''")))
+        if security_filter:
+            filters.append(security_filter)
+        filter = None if len(filters) == 0 else " and ".join(filters)
```
pamelafox, line 91: Can you add tests for this logic? There's pretty good coverage for chatreadretrieveread right now, so it should be possible to test it. You could spy on the search method and make sure it gets passed the correct filters. You could also move this into a build_search_filters function if that made sense.

app/backend/approaches/readdecomposeask.py
```diff
@@ -37,13 +38,22 @@ def __init__(
...
+    async def search(
+        self, query_text: str, overrides: dict[str, Any], auth_claims: dict[str, Any]
+    ) -> tuple[list[str], str]:
         has_text = overrides.get("retrieval_mode") in ["text", "hybrid", None]
         has_vector = overrides.get("retrieval_mode") in ["vectors", "hybrid", None]
         use_semantic_captions = True if overrides.get("semantic_captions") and has_text else False
-        top = overrides.get("top") or 3
         exclude_category = overrides.get("exclude_category") or None
-        filter = "category ne '{}'".format(exclude_category.replace("'", "''")) if exclude_category else None
+        top = overrides.get("top") or 3
```
pamelafox, line 48:
```suggestion
        top = overrides.get("top", 3)
```

app/backend/approaches/readdecomposeask.py
```diff
@@ -37,13 +38,22 @@ def __init__(
...
+
+        security_filter = AuthenticationHelper.build_security_filters(overrides, auth_claims)
+        filters = []
+        if exclude_category:
+            filters.append("category ne '{}'".format(exclude_category.replace("'", "''")))
+        if security_filter:
+            filters.append(security_filter)
+        filter = None if len(filters) == 0 else " and ".join(filters)
```
pamelafox, line 56: Ditto my comment in other file about testing this. Code looks the same too, so this could live in approach.py, but I know Enrique/Pablo also likes the idea of having some redundant code to make it easier to tweak, so I don't know. I'm happy as long as code is tested in both spots.

app/backend/approaches/readretrieveread.py
```diff
@@ -68,13 +69,20 @@ def __init__(
...
         exclude_category = overrides.get("exclude_category") or None
-        filter = "category ne '{}'".format(exclude_category.replace("'", "''")) if exclude_category else None
+
+        security_filter = AuthenticationHelper.build_security_filters(overrides, auth_claims)
+        filters = []
+        if exclude_category:
+            filters.append("category ne '{}'".format(exclude_category.replace("'", "''")))
+        if security_filter:
+            filters.append(security_filter)
+        filter = None if len(filters) == 0 else " and ".join(filters)
```
pamelafox, line 85: Code spotting #3! I assume this approach still doesn't work, right? Due to #541

app/backend/core/authentication.py
```diff
@@ -0,0 +1,206 @@
...
+                resp_json = await resp.json()
+                resp_status = resp.status
+                if resp_status != 200:
+                    raise AuthError(error=resp_json, status_code=resp_status)
+
+            while resp_status == 200:
+                value = resp_json["value"]
+                for group in value:
+                    groups.append(group["id"])
+                nextLink = resp_json.get("@odata.nextLink")
```
pamelafox, line 157: nextLink should use Python casing, next_link

Fun traversal code here!

app/frontend/src/components/LoginButton/LoginButton.module.css
pamelafox: Did you also run the precommits? You might need to --run all-files as this file doesn't have a newline at end yet.

scripts/adlsgen2setup.py
```diff
@@ -0,0 +1,177 @@
...
+        """
+        self.data_directory = data_directory
+        self.storage_account_name = storage_account_name
+        self.filesystem_name = filesystem_name
+        self.credentials = credentials
+        self.security_enabled_groups = security_enabled_groups
+        self.data_access_control_format = data_access_control_format
+
+    async def run(self):
+        async with DataLakeServiceClient(
```
pamelafox, line 54: I notice you use "async with" for both the client and credentials, and I think in other scripts we just create the clients and pass them around. I'm curious if the "async with" context manager approach is beneficial in some way? Should we be using it elsewhere?

mattmsft: if you pass the normal clients around they make blocking http calls. Using async takes advantage of aiohttp non blocking behavior.

app/backend/approaches/readretrieveread.py
mattmsft: i'm a bit confused - this approach does work. The issue you linked was related to chat completions, but this filter is passed into the search query which doesn't have anything to do with chat completions.

scripts/manageacl.py
```diff
@@ -0,0 +1,176 @@
...
+        self.acl_action = acl_action
+        self.acl_type = acl_type
+        self.acl = acl
+
+    async def run(self):
+        endpoint = f"https://{self.service_name}.search.windows.net"
+        async with SearchClient(
+            endpoint=endpoint, credential=self.credentials, index_name=self.index_name
+        ) as search_client, SearchIndexClient(endpoint=endpoint, credential=self.credentials) as search_index_client:
+            if self.acl_action == "enable_acls":
```
pamelafox, line 61: I think it'd be nice design wise for this class to have enable_acls(), view_acls(), , remove_all_acls(), remove_acl() and add_acl(), and then run() can call those as needed. That way, if someone wanted to use this class via some other mechanism, like a UI or API, they could call the individual methods. It's also easier to test, I think.

app/backend/approaches/readdecomposeask.py
mattmsft: or i can just add the build_search_filters method

scripts/prepdocs.py
```diff
@@ -501,12 +516,96 @@ def read_files(
                 print(f"\tGot an error while reading {filename} -> {e} --> skipping file")
 
 
+def read_adls_gen2_files(
```
pamelafox, line 519: Please test this function (especially the acl listing logic in the middle)

scripts/prepdocs.py
```diff
@@ -501,12 +516,96 @@ def read_files(
...
+                    if use_vectors and vectors_batch_support:
+                        sections = update_embeddings_in_batch(sections)
+                    index_sections(os.path.basename(path.name), sections, acls)
+                except Exception as e:
+                    print(f"\tGot an error while reading {path.name} -> {e} --> skipping file")
+                finally:
+                    try:
+                        os.remove(temp_file_path)
+                    except Exception as e:
+                        print(f"\tGot an error while delete {temp_file_path} -> {e}")
```
pamelafox, line 580:
```suggestion
                        print(f"\tGot an error while deleting {temp_file_path} -> {e}")
```

app/backend/approaches/readretrieveread.py
pamelafox: I meant the approach as a whole - I imagine your internal search query works.

scripts/adlsgen2setup.py
pamelafox: Huh, even if you pass around a client from the aio module? I think that means our current code isn't set up well. Was there documentation around that? (I believe you, just trying to understand better myself)

LoginAndAclSetup.md
pamelafox: Yeah I'm fine with that. I'm very curious how we can empower developers to use certificates, based on my difficulties setting them up.

LoginAndAclSetup.md
mattmsft: updated the readme

LoginAndAclSetup.md
mattmsft: Discussed offline - there is an opportunity to enhance this using Azure Key Vault + self-signed certs once we have bicep support.

mattmsft: Added a note stating that certificates are considered more secure and linked to the certificate creation docs

LoginAndAclSetup.md
mattmsft: good catch!

LoginAndAclSetup.md
mattmsft: Correct - adding this to the doc

LoginAndAclSetup.md
mattmsft: This is right below on 161

mattmsft: changed the formatting to make this more clear

LoginAndAclSetup.md
mattmsft: Agreed - will follow up one bicep support is available

scripts/adlsgen2setup.py
mattmsft: so the problem is you'll randomly see "client session unclosed" errors. A cheap way to reproduce this is to force close the app when it's serving a streaming chat session


Here's the GH PR / has some linked issues: https://github.com/Azure/azure-sdk-for-python/pull/9090

LoginAndAclSetup.md
mattmsft: Yes will do that

app/backend/approaches/chatreadretrieveread.py
mattmsft: adding tests for build_search_filters

scripts/prepdocs.ps1
```diff
@@ -34,6 +9,21 @@ if (Test-Path -Path "/usr") {
...
+if ($env:AZURE_ADLS_GEN2_STORAGE_ACCOUNT) {
+  $adlsGen2FilesystemPathArg = ""
+  if ($env:AZURE_ADLS_GEN2_FILESYSTEM_PATH) {
+    $adlsGen2FilesystemPathArg = "--datalakefilesystempath $env:ADLS_GEN2_FILESYSTEM_PATH"
+  }
+  $adlsGen2FilesystemArg = ""
+  if ($env:AZURE_ADLS_GEN2_FILESYSTEM) {
+    $adlsGen2FilesystemArg = "--datalakefilesystem $env:ADLS_GEN2_FILESYSTEM"
+  }
+  Start-Process -FilePath $venvPythonPath -ArgumentList "./scripts/prepdocs.py --datalakestorageaccount $env:AZURE_ADLS_GEN2_STORAGE_ACCOUNT $adlsGen2FilesystemArg $adlsGen2FilesystemPathArg --useacls --storageaccount $env:AZURE_STORAGE_ACCOUNT --container $env:AZURE_STORAGE_CONTAINER --searchservice $env:AZURE_SEARCH_SERVICE --openaihost $env:OPENAI_HOST --openaiservice $env:AZURE_OPENAI_SERVICE --openaikey `"$env:OPENAI_API_KEY`" --openaiorg `"$env:OPENAI_ORGANIZATION`" --openaideployment $env:AZURE_OPENAI_EMB_DEPLOYMENT --openaimodelname $env:AZURE_OPENAI_EMB_MODEL_NAME --index $env:AZURE_SEARCH_INDEX --formrecognizerservice $env:AZURE_FORMRECOGNIZER_SERVICE --tenantid $env:AZURE_TENANT_ID -v" -Wait -NoNewWindow
```
pamelafox, line 26: Does this have to be an if/else? I'm worried the repeated code will get out of date with each other, and it's already hard for people to follow the command.
It'd also be good to split the command over multiple lines at this point, it's hard to scroll in editors and find the right arg to modify. I also had customers that didn't realize there were options because they were off screen.

tests/conftest.py
```diff
@@ -137,3 +180,332 @@ async def client(monkeypatch, mock_openai_chatcompletion, mock_openai_embedding,
...
+
+    monkeypatch.setattr(azure.storage.filedatalake.FileSystemClient, "__init__", mock_init)
+    monkeypatch.setattr(azure.storage.filedatalake.FileSystemClient, "get_file_client", mock_get_file_client)
+    monkeypatch.setattr(azure.storage.filedatalake.FileSystemClient, "get_paths", mock_get_paths)
+
+    monkeypatch.setattr(azure.storage.filedatalake.aio.FileSystemClient, "__init__", mock_init_filesystem_aio)
+    monkeypatch.setattr(azure.storage.filedatalake.aio.FileSystemClient, "__aenter__", mock_aenter)
+    monkeypatch.setattr(azure.storage.filedatalake.aio.FileSystemClient, "__aexit__", mock_aexit)
+    monkeypatch.setattr(azure.storage.filedatalake.aio.FileSystemClient, "get_paths", mock_get_paths)
+    monkeypatch.setattr(azure.storage.filedatalake.aio.FileSystemClient, "exists", mock_exists_aio)
```
pamelafox, line 420: Are these monkeypatch's being done in the global namespace? (As opposed to being in a fixture?) I think we always want to do it in a fixture, to have more predictability about when things will be monkeypatched.

mattmsft: ah shoot i need to add some tabs. good catch! should have been in this fixture

@pytest.fixture
def mock_data_lake_service_client

mattmsft: it actually does look like there are tabs in there, i think it was just not looking like it how github rendered. But it's under a fixture.

scripts/prepdocs.ps1
mattmsft: fixed!

494206:
Login is also possible using App Service "easy auth"
https://learn.microsoft.com/en-us/azure/app-service/overview-authentication-authorization

mattmsft:
> Login is also possible using App Service "easy auth" https://learn.microsoft.com/en-us/azure/app-service/overview-authentication-authorization

This is true - but you can't use that for document level access control

494206:
Yes, that's true -- that's why I only suggested 'easy auth' as a possible option for "Logins".  

"Easy auth" can be implemented without making changes in the web app code, but it won't go that extra mile to provide granular document access controls.  

It could be an intermediate solution for some use cases, though.

mattmsft:
> Yes, that's true -- that's why I only suggested 'easy auth' as a possible option for "Logins".
> 
> "Easy auth" can be implemented without making changes in the web app code, but it won't go that extra mile to provide granular document access controls.
> 
> It could be an intermediate solution for some use cases, though.

Agreed. We already have a PR to implement this easy-auth solution. We are planning on automating setup of the access control, and perhaps we can add a "login-only" option that uses the easy-auth you are suggesting. Thanks for your feedback!

PR #627, opened 2023-09-13
Reword top label

pamelafox:
## Purpose

I've had two questions today about "Retrieve this many documents from search:" where devs were worried it wouldn't be searching all the documents. I've attempted a reword.
(Also, "documents" seems misused here, since all chunks could come from same PDF, right?)

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test

* run local server, see label

### DIFF
--- a/app/frontend/src/pages/chat/Chat.tsx
+++ b/app/frontend/src/pages/chat/Chat.tsx
@@ -255,7 +255,7 @@ const Chat = () => {
 
                     <SpinButton
                         className={styles.chatSettingsSeparator}
-                        label="Retrieve this many documents from search:"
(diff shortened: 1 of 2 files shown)


chuwik (approved):

chuwik:
Great suggestion, I think your phrasing makes it more clear. 
FYI we use 'documents' to refer to the data stored in the index, but it's confusing in this case :)

pamelafox:
@chuwik I figured that was the meaning of "documents" here, but yeah most folks were conflating with PDFs.

PR #632, opened 2023-09-14
Run black on all Python files

pamelafox:
## Purpose

This PR runs black, the most popular Python PEP8-compliant formatter, on all files, and adds it to the pre-commit and CI.

This may be a controversial PR as it touches a lot of files. It would be great if we could have all our code consistently formatted, but since we didn't have formatting enforced from the beginning, adding formatting now causes a lot of changed lines. The largest changes are in prepdocs.py - I'd been running black manually on the other files when I made changes, so they were mostly compliant already.

The drawbacks of this change: 
* Any open PRs will likely have some/more merge conflicts. (But they they can run black on their own changes, yay!)
* Harder to git blame and see actual history of a line of code. There is a way to tell git to ignore this commit history for blame purposes if we want though.
* Harder for customers to manually pull our latest changes to their branches.

The benefits:
* Consistent Python formatting going forward. 
* Code reviews can be spent on higher level conversations and not style nits.

I'm researching if there is a path to formatting that'd result in less changed lines, will comment if I find anything. CC @mattmsft as formatting came up in your PR.

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test

* python3 -m pytest

### DIFF
--- a/.github/workflows/python-test.yaml
+++ b/.github/workflows/python-test.yaml
@@ -37,5 +37,7 @@ jobs:
             pip install -r requirements-dev.txt
         - name: Lint with ruff
           run: ruff .
+        - name: Check formatting with black
+          run: black . --check --verbose
         - name: Run Python tests
           run: python3 -m pytest
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -7,6 +7,10 @@ repos:
     -   id: end-of-file-fixer
     -   id: trailing-whitespace
 -   repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.0.282
+    rev: v0.0.289
     hooks:
     -   id: ruff
+-   repo: https://github.com/psf/black
+    rev: 23.9.1
+    hooks:
+    -   id: black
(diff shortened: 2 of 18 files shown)


.github/workflows/python-test.yaml
```diff
@@ -37,5 +37,7 @@ jobs:
             pip install -r requirements-dev.txt
         - name: Lint with ruff
           run: ruff .
+        - name: Check formatting with black
```
mattmsft, line 40: So . is the root directory in this case?

mattmsft (approved):

.github/workflows/python-test.yaml
pamelafox: Yes, so that it includes the Python files in both scripts and backend.

pamelafox:
Okay, good to know you're on board! It's a big change (cosmetically, not functionally) so would like one more approval from Sri or Enrique.

chuwik (approved):

srbalakr (approved):

PR #635, opened 2023-09-14
Run Autopep8 on all files

pamelafox:
## Purpose

This PR is an alternative to https://github.com/Azure-Samples/azure-search-openai-demo/pull/632
which uses autopep8, a less popular format that has more options, like ignoring line length issues. There are still a number of whitespace changes, but not as many.

Maybe we should just exclude prepdocs.py from formatting for now, since thats the one that'll cause the most conflict with other branches.

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test

* python3 -m pytest

### DIFF
--- a/.github/workflows/python-test.yaml
+++ b/.github/workflows/python-test.yaml
@@ -37,5 +37,7 @@ jobs:
             pip install -r requirements-dev.txt
         - name: Lint with ruff
           run: ruff .
+        - name: Check formatting with autopep8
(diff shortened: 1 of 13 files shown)


mattmsft:
In my opinion it's OK to cause conflicts purely due to whitespace changes. Thoughts?

pamelafox:
Abandoning since we went with black.

PR #652, opened 2023-09-19
Stabilize search query generation

srbalakr:
## Purpose
Stabilize search query generation using function calls as one approach. This example also illustrate function call to get structure response in rag sample. 

```
[ ] Yes
[ x] No
```

## Pull Request Type

```
[ ] Bugfix
[ x] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
```

## What to Check
* ...


app/backend/approaches/chatreadretrieveread.py
```diff
@@ -242,3 +264,17 @@ def get_messages_from_history(
 
         messages = message_builder.messages
         return messages
+    
+    def get_search_query(self, chat_completion: any, user_query: str):
+        response_message = chat_completion["choices"][0]["message"]
+        print(response_message)
```
pamelafox, line 270: Remove print() call

app/backend/approaches/chatreadretrieveread.py
```diff
@@ -242,3 +264,17 @@ def get_messages_from_history(
 
         messages = message_builder.messages
         return messages
+    
+    def get_search_query(self, chat_completion: any, user_query: str):
```
pamelafox, line 268: I think type for chat_completion is a dict, right?

app/backend/approaches/chatreadretrieveread.py
```diff
@@ -242,3 +264,17 @@ def get_messages_from_history(
...
+        response_message = chat_completion["choices"][0]["message"]
+        print(response_message)
+        if function_call := response_message.get("function_call"):
+            if function_call["name"] == "search_sources":
+                arg = json.loads(function_call["arguments"])
+                search_query = arg["search_query"]
+                if search_query is not None and search_query != "0":
+                    return search_query
+        elif query_text := response_message.get("content"):
+            if query_text.strip() != "0":
```
pamelafox, line 278: It might be more clear to make "0" a class constant like NO_RESPONSE, since this "0" is pretty far from the prompt.

Have you actually seen it return 0 in any cases by the way? I don't think I have.

app/backend/approaches/chatreadretrieveread.py
```diff
@@ -242,3 +264,17 @@ def get_messages_from_history(
...
         messages = message_builder.messages
         return messages
+    
+    def get_search_query(self, chat_completion: any, user_query: str):
+        response_message = chat_completion["choices"][0]["message"]
+        print(response_message)
+        if function_call := response_message.get("function_call"):
+            if function_call["name"] == "search_sources":
+                arg = json.loads(function_call["arguments"])
+                search_query = arg["search_query"]
```
pamelafox, line 274: Will search_query always be defined? I see you check is not None in the next if, but you use square brackets here instead of get(), so that implies it's always defined but could be None. Is that the case?

app/backend/approaches/chatreadretrieveread.py
pamelafox, line 268: Please add tests specifically for this function that covers its branches

app/backend/approaches/chatreadretrieveread.py
```diff
@@ -33,6 +33,7 @@ class ChatReadRetrieveReadApproach:
 Only generate questions and do not generate any text before or after the questions, such as 'Next Questions'"""
 
     query_prompt_template = """Below is a history of the conversation so far, and a new question asked by the user that needs to be answered by searching in a knowledge base about employee healthcare plans and the employee handbook.
+You have access to azure cognitive search index with 100's of documents.
```
pamelafox, line 36: I assume ChatGPT doesn't care about casing, so better to capitalize for branding reasons?

```suggestion
You have access to Azure Cognitive Search index with 100's of documents.
```

pamelafox: Also just curious how you came up with this additional prompt. Was this via testing? Or based on some research?

app/backend/approaches/chatreadretrieveread.py
```diff
@@ -80,6 +81,23 @@ async def run_until_final_call(
 
         user_q = "Generate search query for: " + history[-1]["user"]
```
pamelafox, line 82: This variable is confusingly named now, since there's also a user_query parameter in the new function that's only the latter part of this variable. Maybe this can be "user_query_request" or something. Or it might be clearer to store both vars here:

```
user_query = history[-1]["user"]
user_query_request = f"Generate search query for: {user_query}"
```

And then pass those around as needed to the functions below.

pamelafox (changes requested):
The CI is breaking with a ruff linter issue. Please also run the tests to see if they still pass, I believe you'll need to update some mocks. Let me know if I can help with the tests.

app/backend/approaches/chatreadretrieveread.py
srbalakr: Ack. updated.

app/backend/approaches/chatreadretrieveread.py
srbalakr: There was one more azure sample that had example of specifying context https://github.com/Azure-Samples/openai/blob/main/Basic_Samples/Functions/functions_with_azure_search.ipynb. Also tested extensively for quality.

app/backend/approaches/chatreadretrieveread.py
srbalakr: It appeared in corner cases where user questions aren't within the scope.

srbalakr:
> The CI is breaking with a ruff linter issue. Please also run the tests to see if they still pass, I believe you'll need to update some mocks. Let me know if I can help with the tests.



> The CI is breaking with a ruff linter issue. Please also run the tests to see if they still pass, I believe you'll need to update some mocks. Let me know if I can help with the tests.

Addressed comments and added Ut's. Help take a look.

pamelafox (approved):

PR #659, opened 2023-09-20
Smooth streaming, fix partial citation, and ensure visibility.

SteveJSteiner:
## Purpose

Allow the streamed output to become visible at a smoother cadence.  
Comparing the left (current streaming) and the right (smooth streaming) a few things are different:
For the first call the response packets arrive at relatively the same rate but ReAct is buffering and displaying them in chunks on the left and displaying them at a fixed rate on the right. Ideally, we want streaming to provide the response at a fixed rate that is faster than typical reading speed, but no faster.  (Zooming into the gif can help see the differences.)

This change includes a fix for streaming responses scrolling past the end of the screen, and another to prevent partial citation text from appearing in the answer and then 'flashing' into the correct format for citations.  

(This is a cherry-picked + squashed version of PR #618, absent the accidental merge.)

SmoothScrolling

```
[ ] Yes
[ X] No
```

## Pull Request Type

```
[X] Bugfix
[X] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
azd up
```

* Test the code

```
Ask sufficient questions to ensure the chat scrolls off the screen.
```

## What to Check
* The text of responses streams smoothly, and the answers remain visible.  Test swapping streaming off and on

### DIFF
--- a/app/frontend/src/components/Answer/AnswerParser.tsx
+++ b/app/frontend/src/components/Answer/AnswerParser.tsx
@@ -7,7 +7,7 @@ type HtmlParsedAnswer = {
     followupQuestions: string[];
 };
 
-export function parseAnswerToHtml(answer: string, onCitationClicked: (citationFilePath: string) => void): HtmlParsedAnswer {
+export function parseAnswerToHtml(answer: string, isStreaming: boolean, onCitationClicked: (citationFilePath: string) => void): HtmlParsedAnswer {
     const citations: string[] = [];
     const followupQuestions: string[] = [];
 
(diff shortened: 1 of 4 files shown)


SteveJSteiner:
@pamelafox - This should resolve the 'flashing' for citations.  I accidentally merged updates from main into the branch that was driving the original PR ... which is why I opened this one.

app/frontend/src/components/Answer/AnswerParser.tsx
```diff
@@ -20,6 +20,21 @@ export function parseAnswerToHtml(answer: string, onCitationClicked: (citationFi
     // trim any whitespace from the end of the answer after removing follow-up questions
     parsedAnswer = parsedAnswer.trim();
 
+    // Omit a citation that is still being typed during streaming
+    if (isStreaming){
```
pamelafox, line 24: I assume this is the part that changed after I last looked at it, this looks reasonable, thanks for addressing that!

pamelafox (approved):

PR #677, opened 2023-09-26
Add E2E tests using Playwright, increase accessibility of buttons

pamelafox:
## Purpose

This PR adds end-to-end tests using the Playwright SDK, specifically using the pytest-playwright plugin.

I could have chosen to have these tests run when issuing `python3 -m pytest` but we already have unit tests running then, and the playwright tests require an additional install step for the browsers. Plus, playwright tests are slower to run than unit tests. So I've opted the tests out of default test discovery by naming the file "e2e.py" (versus test_e2e.py) and I run pytest on e2e.py in a separate step in the CI and documentation.

I also want to call out how I did the mocking, since there were some design decisions there. For the environment variables that are used to setup the app, I used the mock environment variables from conftest.py, and moved those into a fixture for easy use by both the unit tests and e2e tests. When the app is actually running and the frontend is making HTTP calls to the backend, I use playwright's route() function to mock out the backend calls. For the mock JSON, I pass the results from the snapshot tests, since that ensures that our E2E tests are working with what our unit tests are outputting. (We can still suffer breakages if the ChatGPT API itself changes schema, but typically tests should not use live APIs, or at least, that should be reserved for tests on staged environments.)

While making these tests, I discovered that several of the "buttons" in the UI were not actually buttons, which is bad both for accessibility reasons and ease of E2E testing. I upgraded them to use the Button component from FluentUI, which seems to work fairly well:

Screenshot of new buttons

Some of the colors changed from light gray to dark gray as well, also largely for accessible reasons (color contrast).

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[X] Other... Please describe: TESTS!
```

## How to Test

```
python3 -m pytest tests/e2e.py
```


requirements-dev.txt
```diff
@@ -7,6 +7,6 @@ pytest-asyncio
 pytest-snapshot
 coverage
 pytest-cov
+pytest-playwright
 pre-commit
 locust
-azure-storage-file-datalake==12.12.0
```
pamelafox, line 12: @mattmsft This requirement is also in scripts/requirements.txt, so it doesn't need to be here as well, right?

tests/conftest.py
```diff
@@ -160,26 +160,32 @@ async def mock_search(self, *args, **kwargs):
...
-
-    with mock.patch("app.DefaultAzureCredential") as mock_default_azure_credential:
-        mock_default_azure_credential.return_value = MockAzureCredential()
-        quart_app = app.create_app()
-
-        async with quart_app.test_app() as test_app:
-            quart_app.config.update({"TESTING": True})
-
-            yield test_app.test_client()
+@pytest.fixture(params=envs, ids=["client0", "client1"])
```
pamelafox, line 163: This is just a refactoring so that I could use the mocked env in both the unit and e2e tests

tests/conftest.py
```diff
@@ -160,26 +160,32 @@ async def mock_search(self, *args, **kwargs):
...
+def mock_env(monkeypatch, request):
+    with mock.patch.dict(os.environ, clear=True):
```
pamelafox, line 165: One addition: Now we patch the entire environment, clearing it out, to make sure our tests are isolated from the dev's environment vars.

tests/test_app.py
```diff
@@ -8,12 +10,12 @@
 
 @pytest.mark.asyncio
 async def test_missing_env_vars():
-    quart_app = app.create_app()
+    with mock.patch.dict(os.environ, clear=True):
```
pamelafox, line 13: Now we patch the entire environment, clearing it out, to make sure our tests are isolated from the dev's environment vars. Otherwise this test was sometimes erroring when my env vars were set locally, since this test is meant to emulate the situation of no-env-vars.

.github/workflows/python-test.yaml
```diff
@@ -40,7 +40,7 @@ jobs:
         - name: Check formatting with black
           run: black . --check --verbose
         - name: Run Python tests
-          run: python3 -m pytest
+          run: python3 -m pytest --cov
```
mattmsft, line 43: is this the coverage flag? I see you removed it from an config file below

mattmsft (approved):

mattmsft:
looks like we are missing an await somewhere? python3.11 tests fail

> tests/e2e.py::test_chat[client0-chromium]
tests/e2e.py::test_chat[client1-chromium]
  /opt/hostedtoolcache/Python/3.11.5/x64/lib/python3.11/site-packages/playwright/_impl/_network.py:260: RuntimeWarning: coroutine 'Channel.send' was never awaited
    pass

pamelafox:
@mattmsft Yep :( I'm having a heck of a time getting playwright working in a 3.11 dev container locally so that I can replicate, but hope to get it working soon.

.github/workflows/python-test.yaml
pamelafox: Yes, good catch. I removed it from the overall configuration as it seems to be incompatible with running a single module (as I'm trying to do with the e2e tests). This means that you have to specify it when you want --cov so I put it in the README. I can explore other workarounds for running just the E2E tests if you're worried about adding --cov manually. Maybe keyword would work.

pamelafox:
@mattmsft The tests are now passing, though the only change I made was to add verbose output to pytest and to pin the playwright version. I don't think those changes should have fixed it, so it may have just been a timeout issue. We can increase the timeout for the E2E tests if there's continued flakiness on CI, that's configurable in many places for playwright.

mattmsft:
> @mattmsft The tests are now passing, though the only change I made was to add verbose output to pytest and to pin the playwright version. I don't think those changes should have fixed it, so it may have just been a timeout issue. We can increase the timeout for the E2E tests if there's continued flakiness on CI, that's configurable in many places for playwright.

boosting the timeout is a good idea

PR #679, opened 2023-09-26
Conform to Chat App specification for responses

pamelafox:
## Purpose

A team is working on standardizing Chat App request and response formats across our various tools and samples, so that developers can more easily swap in different technologies, use different frontends, etc.

The Chat App schema is designed to look like the ChatCompletions API, since that's already so commonly used, with additional slots for `extra_args` (anything a particular app needs, like `data_points` and `thought_process` for us) and `session_state` (memory between client/server, unused by us). See more in the proposal:
https://github.com/Azure/azureml_run_specification/pull/1066/files

This PR updates the frontend and backend to conform to just the Response part of the specification. I will send a future PR that conforms the code to the Request part of the specification.

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test

* `python3 -m pytest`


pamelafox:
Comments from Daniel:

At first look (and I might be reading it wrong), I wanted to call out that the actual chat response should be in the "delta": message (in the "content" field of the message). Extra args are for additional context and will not be streamed (unless we change that).

Again, I might be reading it wrong since the json files seem to be having the reply in the message, but for streaming it seems in the extra_args....

Also, note that a message needs to always have a role -- content can be omitted in the first message of a stream.

pamelafox:
@danielsc I took another look and I think my responses match the spec. The extra_args are a property on the first choice, but not in the delta field.


Streaming example (with formatting):

```
{
	"choices": [{
		"delta": {
			"role": "assistant"
		},
		"extra_args": {
			"data_points": ["Benefit_Options-2.pdf: There is a whistleblower policy."],
			"thoughts": "Searched for:<br>capital of France<br><br>Conversations:<br>{'role': 'system', 'content': \"Assistant helps the company employees with their healthcare plan questions, and questions about the employee handbook. Be brief in your answers.\\nAnswer ONLY with the facts listed in the list of sources below. If there isn't enough information below, say you don't know. Do not generate answers that don't use the sources below. If asking a clarifying question to the user would help, ask the question.\\nFor tabular information return it as an html table. Do not return markdown format. If the question is not in English, answer in the language used in the question.\\nEach source has a name followed by colon and the actual information, always include the source name for each fact you use in the response. Use square brackets to reference the source, e.g. [info1.txt]. Don't combine sources, list each source separately, e.g. [info1.txt][info2.pdf].\\n\\n\\n\"}<br><br>{'role': 'user', 'content': 'What is the capital of France?\\n\\nSources:\\nBenefit_Options-2.pdf: There is a whistleblower policy.'}"
		},
		"finish_reason": null,
		"index": 0
	}],
	"object": "chat.completion.chunk"
}
{
	"choices": [{
		"delta": {
			"content": "The capital of France is Paris."
		}
	}]
}
```

danielsc:
Yes, you are right -- for chunked responses content or role can be missing, just for the non-chunked response, both have to be prosent.

tests/snapshots/test_app/test_ask_rtr_hybrid/client0/result.json
```diff
@@ -1,7 +1,15 @@
...
+            },
+            "message": {
+                "content": "The capital of France is Paris."
```
danielsc, line 11: Not sure what this file is used for, but if it is a non-streaming response I would expect the `"role": "assistant"` here

pamelafox: Ah! That's due to my mocks being incomplete. I'll update them according to the OpenAI output. Thanks!

pamelafox:
This is ready for review now. I've double checked with the spec and the keys look correct. I also created issue https://github.com/Azure-Samples/azure-search-openai-demo/issues/740 to track work needed to fully implement protocol.

app/frontend/src/api/models.ts
```diff
@@ -25,13 +25,31 @@ export type AskRequest = {
     idToken?: string;
 };
 
-export type AskResponse = {
-    answer: string;
+export type ResponseMessage = {
```
mattmsft, line 28: Is this protocol documented somewhere? Wondering if we should include a link to the protocol as a comment for context

mattmsft (commented):
Going to manually test this, will sign off once I have manual success

app/frontend/src/api/models.ts
pamelafox: It's documented in a PR: https://github.com/Azure/azureml_run_specification/pull/1066/files
We could link to that or wait for that to be merged.

mattmsft (approved):
Ran a sanity test in codespaces and it worked.

app/frontend/src/api/models.ts
mattmsft: Up to you - I'm OK with a follow up to add this documentation

Issue #672, opened 2023-09-26
Functions call response exceeds token length limit of 32


KesavarajaRK:
> Please provide us with the following information:
> ---------------------------------------------------------------

### This issue is for a: (mark with an `x`)
```
- [ X] bug report -> please search issues before submitting
- [ ] feature request
- [ ] documentation issue or request
- [ ] regression (a behavior that used to work and stopped in a new release)
```

### Minimal steps to reproduce
> 

### Any log messages given by the failure
> ERROR:root:Exception in /chat
Traceback (most recent call last):
  File "D:\Codes\SS-azure-search-openai-demo\app\backend\app.py", line 109, in chat
    r = await impl.run_without_streaming(request_json["history"], request_json.get("overrides", {}))
  File "D:\Codes\SS-azure-search-openai-demo\app\backend\approaches\chatreadretrieveread.py", line 223, in run_without_streaming
    extra_info, chat_coroutine = await self.run_until_final_call(history, overrides, should_stream=False)
  File "D:\Codes\SS-azure-search-openai-demo\app\backend\approaches\chatreadretrieveread.py", line 115, in run_until_final_call
    chat_completion = await openai.ChatCompletion.acreate(
  File "D:\Codes\SS-azure-search-openai-demo\app\backend\backend_env\lib\site-packages\openai\api_resources\chat_completion.py", line 45, in acreate
    return await super().acreate(*args, **kwargs)
  File "D:\Codes\SS-azure-search-openai-demo\app\backend\backend_env\lib\site-packages\openai\api_resources\abstract\engine_api_resource.py", line 217, in acreate
    response, _, api_key = await requestor.arequest(
  File "D:\Codes\SS-azure-search-openai-demo\app\backend\backend_env\lib\site-packages\openai\api_requestor.py", line 382, in arequest
    resp, got_stream = await self._interpret_async_response(result, stream)
  File "D:\Codes\SS-azure-search-openai-demo\app\backend\backend_env\lib\site-packages\openai\api_requestor.py", line 726, in _interpret_async_response
    self._interpret_response_line(
  File "D:\Codes\SS-azure-search-openai-demo\app\backend\backend_env\lib\site-packages\openai\api_requestor.py", line 763, in _interpret_response_line
    raise self.handle_error_response(
[...]

PR #681, opened 2023-09-27
Increased maxtoken to avoid malformed JSON in chatreadretrieveread.py

KesavarajaRK:
## Purpose

Solution for a issue - https://github.com/Azure-Samples/azure-search-openai-demo/issues/672

```
[ ] Yes
[X ] No
```

## Pull Request Type

```
[ X] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
```

## What to Check
* ...

### DIFF
--- a/app/backend/approaches/chatreadretrieveread.py
+++ b/app/backend/approaches/chatreadretrieveread.py
@@ -121,7 +121,7 @@ If you cannot generate a search query, return just the number 0.
             model=self.chatgpt_model,
             messages=messages,
             temperature=0.0,
-            max_tokens=1024,
+            max_tokens=100,  # Setting too low risks malformed JSON, setting too high may affect performance
             n=1,
             functions=functions,
             function_call="auto",
--- a/app/backend/approaches/chatreadretrieveread.py
+++ b/app/backend/approaches/chatreadretrieveread.py
@@ -121,7 +121,7 @@ If you cannot generate a search query, return just the number 0.
             model=self.chatgpt_model,
             messages=messages,
             temperature=0.0,
-            max_tokens=32,
+            max_tokens=self.chatgpt_token_limit,
             n=1,
             functions=functions,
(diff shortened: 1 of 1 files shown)


app/backend/approaches/chatreadretrieveread.py
```diff
@@ -121,7 +121,7 @@ async def run_until_final_call(
             model=self.chatgpt_model,
             messages=messages,
             temperature=0.0,
-            max_tokens=32,
+            max_tokens=self.chatgpt_token_limit,
```
pamelafox, line 124: I don't know if we should up it as far as the limit, since this is just generating a keyword query, but maybe it doesn't hurt to up it to the max. It shouldn't reply with something very long given what it's being prompted for. @srbalakr what do you think?

KesavarajaRK: Got your point and modified the limit to 1024.

pamelafox (approved):

app/backend/approaches/chatreadretrieveread.py
srbalakr: @KesavarajaRK  we need point queries from users conversations. Thats why restricted to smaller length such as 32. Any reason to increase it ?

pamelafox:
Closing this until we get more information about what was wrong with the response.

pamelafox:
I've reopened this since it's definitely occurring often that 32 tokens is not enough for the full JSON for queries that contain harder-to-tokenize words. However, we don't want to increase max_tokens to an unnecessarily high amount in case it affects performance. So I've changed this PR to set it to 100, which seems to work for all the problematic queries.

pamelafox (approved):

Issue #543, opened 2023-08-18
GPT does not respond to Korean documents.


hello-lui:
English documents are normal, but a message saying that the number of tokens is exceeded occurs for Korean documents.
All installations were completed normally, but when I inquire about Korean documents, only the following answers are repeated.

![image](https://github.com/Azure-Samples/azure-search-openai-demo/assets/78183626/d935b9c5-4fe6-40ba-8cbc-0a26c68570a0)

The corresponding Korean document is also attached.
[CEOs_Digest_Weekly Report _ 2023.08.14 (n.314).pdf](https://github.com/Azure-Samples/azure-search-openai-demo/files/12378742/CEOs_Digest_Weekly.Report._.2023.08.14.n.314.pdf)

pamelafox:
Thanks for the report, sounds like this may be a limitation in the package we use to estimate token counts (in messagebuilder/modelhelper). FYI @srbalakr

srbalakr:
Ack, let me check.

jasonchanhku:
Hi I notice that this is also the same case for Chinese characters. Is it the same bug?

hello-lui:
> Hi I notice that this is also the same case for Chinese characters. Is it the same bug? Any updates on this?

@jasonchanhku 

I solved this problem. The cause was that the number of tokens occupied by Korean was nearly twice that of English, and it worked normally after distributing from gpt 3.5-turbo to gpt 3.5-turbo 16K.

hello-lui:
@jasonchanhku

I solved this problem. The cause was that the number of tokens occupied by Korean was nearly twice that of English, and it worked normally after distributing from gpt 3.5-turbo to gpt 3.5-turbo 16K.

pamelafox:
We've merged a proposed fix, please let us know if you experience this issue, and provide the exact input if so.

PR #688, opened 2023-09-29
Normalize text in messages to reduce token length

Yun-BizQA:
## Purpose

* ...
* resolve the bug: [Rag sample clasic] Fix or revisit token estimation

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[X] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## What to Check
* ...
* Korean and english words are running out of maximum token limit at the same time.

### DIFF
--- a/app/backend/approaches/chatreadretrieveread.py
+++ b/app/backend/approaches/chatreadretrieveread.py
@@ -1,6 +1,5 @@
 import json
 from typing import Any, AsyncGenerator
-import unicodedata
 
 import openai
 from azure.search.documents.aio import SearchClient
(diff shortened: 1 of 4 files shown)


pamelafox:
Thanks @Yun-BizQA ! Could you provide an example of Korean text that failed before? And then add a test for that? (See CONTRIBUTING.md for info on running the tests)

Yun-BizQA:
The fix addresses the issue here: https://github.com/Azure-Samples/azure-search-openai-demo/issues/543
The maximum context length is reached easily for korean, chinese, japanese etc because python calculates their length wrong. The fix makes their len equal as the way English char counts, so conversation lasts longer before reaching to such err. The test is performed in the local deployment instead of codes.

app/backend/approaches/chatreadretrieveread.py
```diff
@@ -259,21 +259,19 @@ def get_messages_from_history(
         for shot in few_shots:
             message_builder.append_message(shot.get("role"), shot.get("content"))
 
-        user_content = user_conv
         append_index = len(few_shots) + 1
 
         message_builder.append_message(self.USER, user_content, index=append_index)
 
         for h in reversed(history[:-1]):
+            if message_builder.token_length > max_tokens:
```
pamelafox, line 267: Moved up to account for situation of running out of tokens with just the system prompt and user message

pamelafox:
@Yun-BizQA Can you take a look at the changes I made and see if they make sense? I moved the normalization into the message builder so that its applied to all messages. I also added several tests. Let me know what you think.

also cc'ed @tonybaloney as he has experience with CJK language support.

tonybaloney:
makes sense to me 👍🏻

tonybaloney (approved):

PR #691, opened 2023-09-30
Fix embedding name so that it can be set with azd

pamelafox:
## Purpose

Discovered in debugging session yesterday that we weren't properly allowing embedding deployment name to be overridden. This fixes that and clarifies some things in readme.

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[X] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test

* Follow steps in readme for using existing instance, try name other than embedding
* Run azd up


pamelafox:
Merged from main, so tests now pass.

mattmsft (approved):

Issue #670, opened 2023-09-25
Clear Chat button causes reappearing message


mattmsft:
> Please provide us with the following information:
> ---------------------------------------------------------------

### This issue is for a: (mark with an `x`)
```
- [X] bug report -> please search issues before submitting
- [ ] feature request
- [ ] documentation issue or request
- [ ] regression (a behavior that used to work and stopped in a new release)
```

### Minimal steps to reproduce
* Start a chat session by clicking on a suggested questions
* Get a response for one of the suggested questions
* Click the "Clear Chat" button
* Click on another suggested question
* Observe the old response that was cleared come back, before being replaced by the new response

### Any log messages given by the failure
* None, only visual

### Expected/desired behavior
* Cleared messages do not come back

### OS and Version?
> Windows 7, 8 or 10. Linux (which distribution). macOS (Yosemite? El Capitan? Sierra?)
* Reproducible on Windows 11 and Linux (Codespaces) hosts

### azd version?
> run `azd version` and copy paste here.
azd version 1.3.0 (commit 6605c5768cb3ab6d9e0e6c2d892c9e8b49e1f720)

### Versions
* Reproducible on main branch

### Mention any other details that might be useful

> ---------------------------------------------------------------
> Thanks! We'll be in touch soon.

mattmsft:
[Repro Video](https://1drv.ms/v/s!An-81_3EOirQnjO1eBPkUKUJiEjA?e=xDoQaT)

nisar-snsw:
I'm able to reproduce the issue with the documentation seen on this issue. Can someone confirm that this is indeed fixed?

pamelafox:
No, I cannot replicate. Please upload a video or GIF. Thanks!

PR #716, opened 2023-10-02
Address Issue #670

srbalakr:
## Purpose
Address streaming response when clearing chat messages #670 
* ...

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
```

## What to Check
* ...

### DIFF
--- a/app/frontend/src/pages/chat/Chat.tsx
+++ b/app/frontend/src/pages/chat/Chat.tsx
@@ -42,7 +42,7 @@ const Chat = () => {
 
     const [selectedAnswer, setSelectedAnswer] = useState<number>(0);
     const [answers, setAnswers] = useState<[user: string, response: AskResponse][]>([]);
-    const [streamedAnswers, setstreamedAnswers] = useState<[user: string, response: AskResponse][]>([]);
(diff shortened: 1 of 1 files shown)


app/frontend/src/pages/chat/Chat.tsx
```diff
@@ -134,6 +134,9 @@ const Chat = () => {
         setActiveCitation(undefined);
         setActiveAnalysisPanelTab(undefined);
         setAnswers([]);
+        setstreamedAnswers([]);
```
pamelafox, line 137: Sorry I didn't notice in last PR, can you rename setstreamedAnswers to setStreamedAnswers, here and above? (Proper camel-case). This looks reasonable otherwise, all of these seem like attributes that get set on the next question when needed.

srbalakr: Done

pamelafox (approved):

PR #738, opened 2023-10-05
Remove broken langchain-based approaches

pamelafox:
## Purpose

I am proposing the removal of the 2 langchain-based approaches which have been broken since we moved to 0613 version of APIs.

Reasons to remove:

- They've been broken for months without much concern. (I am the only one that noted it) Developers put their effort into improving the main chat prompt and doc ingestion.
- Reduces complexity of code base.
- Reduces surface area of code to test and maintain.
- OpenAI just announced a new SDK so langchain will likely fall behind again. It's easier for us to stay up to date when we don't have langchain in the middle.
- Reduces dependencies, build time. Some developers aren't able to get langchain installed, and it's a shame to block them for a package that isn't working.

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[X] Other... Please describe:
```

## How to Test

*  All tests still pass
*  Also deployed and tested locally


mattmsft (approved):
LGTM - we will re-evaluate approaches at a later point

wzhwei:
Would be great if Microsoft could release an improved version some time soon! I personally think using agent in Q&A machine important to handle all types of user questions.

PR #743, opened 2023-10-05
Add more Playwright tests to check all routes and buttons

pamelafox:
## Purpose

We're aiming for "100%" test coverage. I don't have a way of instrumenting the Playwright tests to check JS code coverage, but I have at least attempted to have a test for each route and button/form field. It should be relatively easy to add regression e2e tests now, as well.

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[X] Other... Please describe: Tests
```

## How to Test

*  python3 -m pytest tests/e2e.py


tests/conftest.py
```diff
@@ -66,7 +66,7 @@ async def mock_acreate(*args, **kwargs):
         if messages[-1]["content"] == "Generate search query for: What is the capital of France?":
             answer = "capital of France"
         else:
-            answer = "The capital of France is Paris."
+            answer = "The capital of France is Paris. [Benefit_Options-2.pdf]."
```
pamelafox, line 69: I was tempted to make this Q/A make more sense (be tied to our sample data, not about Paris) but that would have been a larger change. The minimal change I needed was the citation, so the e2e tests could click something.

mattmsft: I think this is acceptable for a test

tests/e2e.py
```diff
@@ -80,3 +80,132 @@ def handle(route: Route):
...
+    expect(page.get_by_text("Whats the dental plan?")).to_be_visible()
+    expect(page.get_by_text("The capital of France is Paris.")).to_be_visible()
+    expect(page.get_by_role("button", name="Clear chat")).to_be_enabled()
+
+
+def test_ask(page: Page, live_server_url: str):
+    # Set up a mock route to the /ask endpoint
+    def handle(route: Route):
+        # Read the JSON from our snapshot results and return as the response
+        f = open("tests/snapshots/test_app/test_ask_rtr_hybrid/client0/result.json")
```
mattmsft, line 195: pretty impressive we can leverage the snapshots here as well!

mattmsft (approved):
LGTM

tests/e2e.py
pamelafox: I'm glad you thought that was cool, I thought that was cool!

PR #748, opened 2023-10-06
Chat response

pamelafox:
## Purpose

Implement the Chat Request part of the Chat App protocol:
https://github.com/Azure/azureml_run_specification/pull/1066/
That means the requests send "messages" and "context:overrides", instead of "history" and "overrides" as they sent before.

I elected to adhere to the protocol on multiple levels: in the TSX models, in the HTTP requests to the server, and in the run() method on Approach class. That should mean a high degree of compatibility with both other frontends as well as frameworks that use the protocol.
 
Also adjusts the response to return "context" instead of "extra_args" (a simple rename).

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[ ] Bugfix
[X] Feature
[ ] Other... Please describe:
```

## How to Test

* All tests still pass.
* But we could wait for my other PR to merge that adds way more e2e tests, to make sure they all pass, given this is a frontend-heavy change.


app/backend/app.py
```diff
@@ -88,61 +88,45 @@ async def ask():
...
     try:
-        impl = current_app.config[CONFIG_CHAT_APPROACH]
-        response_generator = impl.run_with_streaming(
-            request_json["history"], request_json.get("overrides", {}), auth_claims
-        )
-        response = await make_response(format_as_ndjson(response_generator))
-        response.timeout = None  # type: ignore
-        return response
+        approach = current_app.config[CONFIG_CHAT_APPROACH]
+        result = await approach.run(request_json["messages"], stream=request_json.get("stream", False), context=context)
```
mattmsft, line 121: Glad we are making this async

app/backend/approaches/chatreadretrieveread.py
```diff
@@ -249,6 +249,20 @@ async def run_with_streaming(
...
+    async def run(
+        self, messages: list[dict], stream: bool = False, session_state: Any = None, context: dict[str, Any] = {}
+    ) -> Union[dict[str, Any], AsyncGenerator[dict[str, Any], None]]:
+        overrides = context.get("overrides", {})
+        auth_claims = context.get("auth_claims", {})
+        if stream is False:
+            # Workaround for: https://github.com/openai/openai-python/issues/371
+            async with aiohttp.ClientSession() as s:
+                openai.aiosession.set(s)
+                response = await self.run_without_streaming(messages, overrides, auth_claims)
```
mattmsft, line 261: In the future if we add anything to `context`, we can refactor the `run_without_streaming` and `run_with_streaming` methods to take the context object instead of unpacking it into overrides and auth claims. But this is good for now

mattmsft (approved):
LGTM. Will run a sanity test in codespaces as well.

mattmsft:
Just ran a manual test and I can still get responses.

Looks like there are some merge conflicts though.

PR #749, opened 2023-10-07
Fixed type annotations per mypy and added mypy to CICD

pamelafox:
## Purpose

I noticed some incorrect types when running mypy on code I was writing, so I went ahead and fixed all the type annotations, as well as added a CI/CD check for mypy going forward.

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[X] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test

* Test still pass
* I will do a test "azd up" to check prepdocs. I don't have ACLs setup for testing yet, so if you have concerns about the manageacls changes, please test those.


app/backend/app.py
```diff
@@ -273,7 +273,7 @@ def create_app():
         AioHttpClientInstrumentor().instrument()
     app = Quart(__name__)
     app.register_blueprint(bp)
-    app.asgi_app = OpenTelemetryMiddleware(app.asgi_app)
+    app.asgi_app = OpenTelemetryMiddleware(app.asgi_app)  # type: ignore[method-assign]
```
pamelafox, line 276: I had to ignore this one, as apparently mypy doesn't approve of assigning to a method, but this is the official way of using middleware in Quart (and other frameworks).

pyproject.toml
```diff
@@ -16,3 +16,14 @@ source = ["scripts", "app"]
...
 [tool.coverage.report]
 show_missing = true
+
+[tool.mypy]
+check_untyped_defs = true
+python_version = 3.9
+
+[[tool.mypy.overrides]]
+module = [
+    "msal.*",
```
pamelafox, line 26: msal doesn't have types :( There's a feature request for it.

scripts/manageacl.py
```diff
@@ -2,11 +2,11 @@
 import asyncio
 import json
 import logging
-from typing import Union
+from typing import Any, Union
 
 from azure.core.credentials import AzureKeyCredential
 from azure.core.credentials_async import AsyncTokenCredential
-from azure.identity import AzureDeveloperCliCredential
+from azure.identity.aio import AzureDeveloperCliCredential
```
pamelafox, line 9: This is one I want feedback on, as the code later says it is expecting an AsyncTokenCredential, but it wasn't pulling the async version of AZDCredential. So should it be sync or async?

scripts/manageacl.py
```diff
@@ -143,14 +143,16 @@ async def enable_acls(self, endpoint: str):
...
-async def main(args: any):
+async def main(args: Any):
     # Use the current user identity to connect to Azure services unless a key is explicitly set for any of them
     azd_credential = (
         AzureDeveloperCliCredential()
         if args.tenant_id is None
         else AzureDeveloperCliCredential(tenant_id=args.tenant_id, process_timeout=60)
     )
-    search_credential = azd_credential if args.search_key is None else AzureKeyCredential(args.search_key)
+    search_credential: Union[AsyncTokenCredential, AzureKeyCredential] = azd_credential
```
pamelafox, line 153: To make mypy happy, I had to break the ternary into an if/else. It's a known issue with ternaries:
https://github.com/python/mypy/issues?q=is%3Aissue+is%3Aopen+ternary

scripts/prepdocs.py
```diff
@@ -177,8 +178,8 @@ def get_document_text(filename):
         for page_num, page in enumerate(form_recognizer_results.pages):
             tables_on_page = [
                 table
-                for table in form_recognizer_results.tables
-                if table.bounding_regions[0].page_number == page_num + 1
+                for table in (form_recognizer_results.tables or [])
```
pamelafox, line 181: These were both optional attributes that we weren't treating as optional, so I put in some guards.

scripts/prepdocs.py
```diff
@@ -118,7 +119,7 @@ def remove_blobs(filename):
     blob_container = blob_service.get_container_client(args.container)
     if blob_container.exists():
         if filename is None:
-            blobs = blob_container.list_blob_names()
+            blobs = iter(blob_container.list_blob_names())
```
pamelafox, line 122: This function was returning an iterator type object, but it wasn't being recognized as such, so calling iter() makes it clear.

scripts/prepdocs.py
```diff
@@ -469,9 +472,9 @@ def remove_from_index(filename):
         r = search_client.search("", filter=filter, top=1000, include_total_count=True)
         if r.get_count() == 0:
             break
-        r = search_client.delete_documents(documents=[{"id": d["id"]} for d in r])
+        removed_docs = search_client.delete_documents(documents=[{"id": d["id"]} for d in r])
```
pamelafox, line 475: It was confused due to the overused variable name `r`

scripts/prepdocs.py
```diff
@@ -706,24 +712,25 @@ def read_adls_gen2_files(
         if args.tenantid is None
         else AzureDeveloperCliCredential(tenant_id=args.tenantid, process_timeout=60)
     )
-    default_creds = azd_credential if args.searchkey is None or args.storagekey is None else None
     adls_gen2_creds = azd_credential if args.datalakekey is None else AzureKeyCredential(args.datalakekey)
-    search_creds = default_creds if args.searchkey is None else AzureKeyCredential(args.searchkey)
+    search_creds: Union[TokenCredential, AzureKeyCredential] = azd_credential
```
pamelafox, line 716: Same issue with ternaries here, had to turn into if/else.

https://github.com/python/mypy/issues?q=is%3Aissue+is%3Aopen+ternary

app/backend/app.py
mattmsft: makes sense

pyproject.toml
mattmsft: Bummer!

scripts/manageacl.py
mattmsft: I think this was just a typo on the import

mattmsft: Everything should be async

mattmsft: Great example of type annotations catching bugs!

mattmsft (approved):
LGTM

PR #773, opened 2023-10-11
Adding session_state to conform to ChatApp protocol

pamelafox:
## Purpose

This PR implements the session_state field of the ChatApp protocol, a mechanism for chat apps to be stateful. We do not yet use it in this app, as the only potential state is the optional authentication, and that's conveyed via a header. However, I've tried to add session_state in such a way that our app is compatible with other apps following the protocol - and so that we might be able to use it in the future (for something like storing conversation ID).

ChatApp spec on session state:
https://github.com/Azure/azureml_run_specification/pull/1066/files#diff-a9609aeac645d2c976714f5da997650e08518430117689ec1632d6cd28b57659R191

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[ ] Bugfix
[X] Feature
[ ] Other... Please describe:
```

## How to Test

* Added new unit tests and updated snapshots
* Added some asserts of POST data in e2e tests


app/backend/approaches/chatreadretrieveread.py
```diff
@@ -224,23 +224,40 @@ async def run_until_final_call(
         return (extra_info, chat_coroutine)
 
     async def run_without_streaming(
-        self, history: list[dict[str, str]], overrides: dict[str, Any], auth_claims: dict[str, Any]
+        self,
```
pamelafox, line 227: Just added the arg, most of diff is from black reformatting

app/backend/approaches/chatreadretrieveread.py
```diff
@@ -224,23 +224,40 @@ async def run_until_final_call(
...
+        chat_resp["choices"][0]["session_state"] = session_state
         return chat_resp
 
     async def run_with_streaming(
-        self, history: list[dict[str, str]], overrides: dict[str, Any], auth_claims: dict[str, Any]
+        self,
+        history: list[dict[str, str]],
+        overrides: dict[str, Any],
+        auth_claims: dict[str, Any],
+        session_state: Any = None,
```
pamelafox, line 246: Ditto

app/backend/approaches/chatreadretrieveread.py
```diff
@@ -224,23 +224,40 @@ async def run_until_final_call(
...
         extra_info, chat_coroutine = await self.run_until_final_call(
             history, overrides, auth_claims, should_stream=True
         )
         yield {
-            "choices": [{"delta": {"role": self.ASSISTANT}, "context": extra_info, "finish_reason": None, "index": 0}],
+            "choices": [
+                {
+                    "delta": {"role": self.ASSISTANT},
+                    "context": extra_info,
+                    "session_state": session_state,
```
pamelafox, line 256: Ditto

pamelafox:
cc @danielsc author of protocol spec

mattmsft (approved):
LGTM

Issue #568, opened 2023-08-30
recent document retrieval issue and max content length


professorDante:
> Please provide us with the following information:
> ---------------------------------------------------------------

### This issue is for a: (mark with an `x`)
```
- [ x] bug report -> please search issues before submitting
- [ ] feature request
- [ ] documentation issue or request
- [ ] regression (a behavior that used to work and stopped in a new release)
```

### Minimal steps to reproduce
> ask for most recent documents with default settings (gives 3). Change the max document retrieval to 5 and ask again.

### Any log messages given by the failure
>

### Expected/desired behavior
>1. Asking same question twice should give same answer (timstamp question).
2. No Max token exceeded

### OS and Version?
> Windows 7, 8 or 10. Linux (which distribution). macOS (Yosemite? El Capitan? Sierra?)
Mac OSX Monterrey
### azd version?
> run `azd version` and copy paste here.
1.1.0
### Versions
>

### Mention any other details that might be useful
1. Asking for most recent documents gives 3 (question snipped from screenshot) 
2. Changing the doc retrieval to 5 then ask again.
3. Documents are completely different to first answer.
4. Ask again
5. Now gives ' maximum content length exceeded'

![Screen Shot 2023-08-30 at 3 13 02 PM](https://github.com/Azure-Samples/azure-search-openai-demo/assets/7493788/5e3ec4a3-8722-4df8-9735-bfc47bba02c8)

> ---------------------------------------------------------------
> Thanks! We'll be in touch soon.

vprzybylo:
I'm also experiencing this issue:

Set doc retrieval to 3:
![issue](https://github.com/Azure-Samples/azure-search-openai-demo/assets/32338353/67cb1663-688e-4918-952b-c0f2c2d0f566)

Set doc retrieval to 5 and ask again and I get different files, only 4, and missing one of the docs that only came out when set to 3
![issue2](https://github.com/Azure-Samples/azure-search-openai-demo/assets/32338353/55616502-2271-47c3-ac9a-25577def1786)

Increasing to 6 gives: Error: This model's maximum context length is 4096 tokens. However, you requested 4301 tokens (3277 in the messages, 1024 in the completion). Please reduce the length of the messages or completion.

srbalakr:
This is a bug being tracked. Will fix in app code soon.

ganesh25:
I also got this error, but for two reasons.

First reason: As already mentioned above, when I increase doc retrieval I get this error.

[...]

PR #778, opened 2023-10-12
Message builder fixes for history format and token count

### LINKED ISSUE #779: Chat History not iterated
### This issue is for a:
```
- [ ] feature request
```

pamelafox:
## Purpose

While trying to replicate https://github.com/Azure-Samples/azure-search-openai-demo/issues/568 I discovered two issues with message builder logic:

1. It was still assuming the old format for messages (changed in my recent chat app protocol schema change). Unit tests didnt catch as they also used old format, so I updated them and added integration tests.
2. The logic was adding messages and only checking token count after. I changed it to first check if the added messages would exceed max tokens and only add them if they fit.

Please take a look at my logic and see if it makes sense and if you have ideas for other tests.

Fixes #779

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[X] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test

* python3 -m pytest


app/backend/approaches/chatreadretrieveread.py
```diff
@@ -282,14 +282,18 @@ def get_messages_from_history(
...
+            if message.get("role") == self.ASSISTANT:
+                ind += 1
+                while ind < len(newest_to_oldest) and newest_to_oldest[ind].get("role") != self.USER:
+                    ind += 1
+                prev_message = newest_to_oldest[ind]
+                if message_builder.calculate_token_length_with_messages([message, prev_message]) > max_tokens:
+                    break
+                message_builder.append_message(self.ASSISTANT, message["content"], index=append_index)
+                message_builder.append_message(self.USER, prev_message["content"], index=append_index)
         return message_builder.messages
```
srbalakr, line 297: @pamelafox thanks for taking care of this. I had same solution in mind. But we can implement  in a different way since this iterates over 2 loops and for 32K limit, conversation length might be longer and cause larger runtime.

My thinking is that
1. Track token count of messages in chatapproach.
2. As we currently do, start adding latest message to the message builder. 
3. Before we add a message into message builder, evaluate token length of the new message and evaluate the  condition
4. `msg_count = messagebuilder.cal_token_length_for_msg(new message)`
    ` if msg_count + total_token_count > max_tokens: break`
    `msg_builder.append(new message)`
     `total_token_count  = messagebuilder.get_tokencount()`
    This would avoid two loops and still follow LIFO to add the message builder.

srbalakr (changes requested):

app/backend/approaches/chatreadretrieveread.py
pamelafox: I've updated the code to look more like your suggestion. However, I still have it so that it only adds a *pair* at a time, so it always checks for bot/user pairs. That seems like how it was originally and I thought it was intentional to always return pairs or none at all. If we think its helpful to return a bot's answer even without the user's question, then I can adjust it. (That would be simpler code!)

srbalakr: Thats fine I think, model needs the latest context. ref: https://learn.microsoft.com/en-us/azure/ai-services/openai/how-to/chatgpt?pivots=programming-language-chat-completions#managing-conversations

srbalakr: @pamelafox also we need to reduce maxtokens to maxtokens - 1024. Because the token limit error says Model limit reached with (3K in messages and 1024 in completion) So we need to account for the max response token length returned from model which we hardcoded in the code aas 1024.

app/backend/approaches/chatreadretrieveread.py
```diff
@@ -109,12 +109,12 @@ async def run_until_final_call(
 
         # STEP 1: Generate an optimized keyword search query based on the chat history and the last question
         messages = self.get_messages_from_history(
-            self.query_prompt_template,
-            self.chatgpt_model,
-            history,
-            user_query_request,
-            self.query_prompt_few_shots,
-            self.chatgpt_token_limit - len(user_query_request),
+            system_prompt=self.query_prompt_template,
```
pamelafox, line 112: I added keyword args to make it clear what the args are, given how many there are.

app/backend/approaches/chatreadretrieveread.py
pamelafox: I've changed it so it's fine with breaking up message pairs, which is much simpler code. Also added another test for that. And I made the change to consider max token length as subtracting 1024.

pamelafox: Also added a test for the 1024 issue, plus a logging.debug statement for when that's reached, as that seems like it might be helpful for folks debugging their message histories.

pamelafox:
@srbalakr This is ready for re-review, thanks!

srbalakr (approved):

PR #780, opened 2023-10-12
support non English languages

ks6088ts:
## Purpose

* First of all, thank you for providing this useful sample code :) For now, to support non English languages, several hardcoded codes should be overwritten. In fact, we Japanese fork the repo and overwrite them to support customers (ref. [nohanaga/azure-search-openai-demo](https://github.com/nohanaga/azure-search-openai-demo), [asashiho/azure-search-openai-demo](https://github.com/asashiho/azure-search-openai-demo), [ks6088ts-labs/_azure-search-openai-demo](https://github.com/ks6088ts-labs/_azure-search-openai-demo), etc). But this approach brings some pain to catch up with the updates on new commits. So, in this PR, configurable parameters are introduced to get rid of those hardcoded codes. This fix can enlarge customers to worldwide, I think.

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
# To use Japanese languages
azd env set AZURE_SEARCH_QUERY_LANGUAGE ja-jp
azd env set AZURE_SEARCH_QUERY_SPELLER none
azd env set AZURE_SEARCH_ANALYZER_NAME ja.lucene

# deploy
azd up
```

## What to Check
* Check the App Service environment variables
* Check the analyzer settings of Cognitive Search index


pamelafox:
Neat! So you don't also have to change the English prompts?

README.md
```diff
@@ -169,7 +169,9 @@ When you run `azd up` after and are prompted to select a value for `openAiResour
   then run `azd env set AZURE_SEARCH_SERVICE_LOCATION {Location of existing service}`
 1. If the search service's SKU is not standard, then run `azd env set AZURE_SEARCH_SERVICE_SKU {Name of SKU}`. The free tier won't work as it doesn't support managed identity. ([See other possible values](https://learn.microsoft.com/azure/templates/microsoft.search/searchservices?pivots=deployment-language-bicep#sku))
 1. If you have an existing index that is set up with all the expected fields, then run `azd env set AZURE_SEARCH_INDEX {Name of existing index}`. Otherwise, the `azd up` command will create a new index.
-
+1. If you want to configure the language of the search query, then run `azd env set AZURE_SEARCH_QUERY_LANGUAGE {Name of query language}`. ([See other possible values](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/search/azure-search-documents/azure/search/documents/aio/_search_client_async.py#L226))
```
pamelafox, line 172: I was going to suggest linking to the generated reference doc at https://learn.microsoft.com/en-us/python/api/azure-search-documents/azure.search.documents.aio.searchclient?view=azure-python#azure-search-documents-aio-searchclient-search but it doesnt show those parameters. @mattmsft Do you know why theyre missing?

mattmsft: I do not know. But we can link to this which also has all the types
https://learn.microsoft.com/en-us/python/api/azure-search-documents/azure.search.documents.models.querylanguage?view=azure-python-preview

ks6088ts:
> Neat! So you don't also have to change the English prompts?

Good question.  Thanks to the cool prompt [If the question is not in English, answer in the language used in the question.](https://github.com/Azure-Samples/azure-search-openai-demo/blob/main/app/backend/approaches/chatreadretrieveread.py#L30), there seems nothing to be desired.
Actually, it works as follows!

スクリーンショット 2023-10-13 083208

I think it would be better to inject prompts from environment variable, but it is out of this PR's scope.

pamelafox:
Cool! It looks good except for updating the links to the docs when possible.

README.md
ks6088ts: Thanks for pointing it out @pamelafox @mattmsft 
It is exactly where I'm not confident.
I updated links [923a94e](https://github.com/Azure-Samples/azure-search-openai-demo/pull/780/commits/923a94ec15000dc13367e5324a028998bcab0ac8)
But unfortunately, there's no good reference for analyzer types in Python sdk docs, so I refer to dotnet sdk ones. 
https://github.com/Azure-Samples/azure-search-openai-demo/pull/780/commits/923a94ec15000dc13367e5324a028998bcab0ac8

README.md
```diff
@@ -169,7 +169,9 @@ When you run `azd up` after and are prompted to select a value for `openAiResour
   then run `azd env set AZURE_SEARCH_SERVICE_LOCATION {Location of existing service}`
 1. If the search service's SKU is not standard, then run `azd env set AZURE_SEARCH_SERVICE_SKU {Name of SKU}`. The free tier won't work as it doesn't support managed identity. ([See other possible values](https://learn.microsoft.com/azure/templates/microsoft.search/searchservices?pivots=deployment-language-bicep#sku))
 1. If you have an existing index that is set up with all the expected fields, then run `azd env set AZURE_SEARCH_INDEX {Name of existing index}`. Otherwise, the `azd up` command will create a new index.
-
+1. If you want to configure the language of the search query, then run `azd env set AZURE_SEARCH_QUERY_LANGUAGE {Name of query language}`. ([See other possible values](https://learn.microsoft.com/en-us/python/api/azure-search-documents/azure.search.documents.models.querylanguage?view=azure-python-preview))
```
pamelafox, line 172:
```suggestion
1. If you want to configure the language of the search query, then run `azd env set AZURE_SEARCH_QUERY_LANGUAGE {Name of query language}`. ([See other possible values](https://learn.microsoft.com/python/api/azure-search-documents/azure.search.documents.models.querylanguage?view=azure-python-preview))
```

README.md
```diff
@@ -169,7 +169,9 @@ When you run `azd up` after and are prompted to select a value for `openAiResour
...
+1. If you want to configure the type of the speller to use to spell-correct, then run `azd env set AZURE_SEARCH_QUERY_SPELLER {Name of speller type}`. ([See other possible values](https://learn.microsoft.com/en-us/python/api/azure-search-documents/azure.search.documents.models.queryspellertype?view=azure-python-preview))
+1. If you want to configure the name of the analyzer to use for a searchable text field, then run `azd env set AZURE_SEARCH_ANALYZER_NAME {Name of analyzer name}`. ([See other possible values](https://learn.microsoft.com/en-us/dotnet/api/microsoft.azure.search.models.field.analyzer?view=azure-dotnet-legacy&viewFallbackFrom=azure-dotnet))
```
pamelafox, line 174: These instructions are under "Use existing Azure search resource" which isn't the exact right heading, but perhaps is close enough.

I have some suggested rewording though, to make it clear what the change/defaults are:

```
You can also customize the search service (new or existing) for non-English searches:

1. To configure the language of the search query to a value other than "en-us", run `azd env set AZURE_SEARCH_QUERY_LANGUAGE {Name of query language}`. ([See other possible values](https://learn.microsoft.com/python/api/azure-search-documents/azure.search.documents.models.querylanguage?view=azure-python-preview))
1. To turn off the spell checker, run `azd env set AZURE_SEARCH_QUERY_SPELLER none`. ([See other possible values](https://learn.microsoft.com/python/api/azure-search-documents/azure.search.documents.models.queryspellertype?view=azure-python-preview))
1. To configure the name of the analyzer to use for a searchable text field to a value other than "en.microsoft", run `azd env set AZURE_SEARCH_ANALYZER_NAME {Name of analyzer name}`. ([See other possible values](https://learn.microsoft.com/dotnet/api/microsoft.azure.search.models.field.analyzer?view=azure-dotnet-legacy&viewFallbackFrom=azure-dotnet))
```

ks6088ts: @pamelafox Thank you for the suggestion. I reflected them [90c6dc4](https://github.com/Azure-Samples/azure-search-openai-demo/pull/780/commits/90c6dc44b4414679ec74a51101dfa5eef4b75362)

pamelafox (approved):
Great, I checked it out locally and was able to deploy and run local server without any changes, so this shouldn't introduce any backwards incompatible issues for devs who pull the latest. Thanks for helping this repo be useful for more international users!

Issue #676, opened 2023-09-26
 Error communicating with OpenAI: URL has an invalid label


al34n1x:
When running `azd up` as part of the extracting text document, I get the following error message for every single PDF

`Error communicating with OpenAI: URL has an invalid label. --> skipping file
Processing`

I'm not an Azure-OpenAI user, I'm leveraging my OpenAI account.

When running from Azure, or even locally, I'm not getting a proper response to my prompts.

### This issue is for a: (mark with an `x`)
```
- [ ] bug report -> please search issues before submitting
- [ ] feature request
- [x] documentation issue or request
- [ ] regression (a behavior that used to work and stopped in a new release)
```

This is the kind of response that I'm getting

![Screenshot 2023-09-26 at 19 35 05](https://github.com/Azure-Samples/azure-search-openai-demo/assets/9869723/7ccbb645-5dda-4b74-bb9f-e33379f0563d)

Thank you!

> ---------------------------------------------------------------
> Thanks! We'll be in touch soon.

pamelafox:
@carlotta94c Can you look into this, as they're using non-Azure OpenAI?

My general recommendation for debugging OpenAI errors is to set logging mode to debug in prepdocs.py:


logging.basicConfig(logging.DEBUG)


That will then show you the URL that OpenAI is trying to access, and you can compare that URL to what you expect it to be.

mattmsft:
Do these instructions work for you?
https://github.com/Azure-Samples/azure-search-openai-demo#using-a-non-azure-openai-instance

> [Using a non-Azure OpenAI instance](https://github.com/Azure-Samples/azure-search-openai-demo#using-a-non-azure-openai-instance)
> To use an existing non-Azure OpenAI account, follow these steps before running azd up

> Run azd env set OPENAI_HOST openai
> Run azd env set OPENAI_ORGANIZATION {Your OpenAI organization}
> Run azd env set OPENAI_API_KEY {Your OpenAI API key}
> Run azd up
> You can retrieve your OpenAI key by checking [your user page](https://platform.openai.com/account/api-keys) and your organization by navigating to [your organization page](https://platform.openai.com/account/org-settings). Learn more about creating an OpenAI free trial at [this link](https://openai.com/pricing). Do not check your key into source control.

al34n1x:
@pamelafox I'd try that and get back to you guys. Thanks!

@mattmsft I did follow those instructions. I'm even able to run other prompts and get responses from OpenAI. 


[...]

PR #796, opened 2023-10-14
Add arguments required for OpenAI connection

wdfox:
## Purpose

* Commit [c8b8486](https://github.com/Azure-Samples/azure-search-openai-demo/commit/c8b8486d68527026b0660a3a75cc4e5e376793c1) removed two command line arguments to the `prepdocs.sh` and `prepdocs.ps1` scripts which break the document loading process for users of the external OpenAI connection. This has been addressed in the open issue #676 
* I added the two arguments (`openaihost` and `openaikey`) back into the file, fixing the problem.
* Based on the logic of `prepdocs.py` (specifically lines 736-752), I don't believe the change will impact users of the Azure OpenAI service. However, I don't have access, so I cannot test this. 

```
[ ] Yes
[X ] No
```

## Pull Request Type

```
[X ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

Run `azd up` and specifically watch the `prepdocs` process. There should be no errors for users of the external OpenAI connection.

## What to Check
* I only have the ability to run in a Linux environment with 3rd party OpenAI connection. 
* It would help for a maintainer to validate that the change (1) doesn't affect users of the Azure OpenAI service, and (2) the changes to `prepdocs.ps1` solve the issue for Windows users.

### DIFF
--- a/scripts/prepdocs.ps1
+++ b/scripts/prepdocs.ps1
@@ -33,6 +33,5 @@ $argumentList = "./scripts/prepdocs.py `"$cwd/data/*`" $adlsGen2StorageAccountAr
 "--openaiservice `"$env:AZURE_OPENAI_SERVICE`" --openaikey `"$env:OPENAI_API_KEY`" " + `
 "--openaiorg `"$env:OPENAI_ORGANIZATION`" --openaideployment `"$env:AZURE_OPENAI_EMB_DEPLOYMENT`" " + `
 "--openaimodelname `"$env:AZURE_OPENAI_EMB_MODEL_NAME`" --index $env:AZURE_SEARCH_INDEX " + `
-"--formrecognizerservice $env:AZURE_FORMRECOGNIZER_SERVICE --tenantid $env:AZURE_TENANT_ID" + `
-"--openaihost $env:OPENAI_HOST --openaikey $env:OPENAI_API_KEY -v"
+"--formrecognizerservice $env:AZURE_FORMRECOGNIZER_SERVICE --tenantid $env:AZURE_TENANT_ID -v"
 Start-Process -FilePath $venvPythonPath -ArgumentList $argumentList -Wait -NoNewWindow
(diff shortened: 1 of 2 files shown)


wdfox:

scripts/prepdocs.ps1
```diff
@@ -33,5 +33,6 @@ $argumentList = "./scripts/prepdocs.py `"$cwd/data/*`" $adlsGen2StorageAccountAr
 "--openaiservice `"$env:AZURE_OPENAI_SERVICE`" --openaikey `"$env:OPENAI_API_KEY`" " + `
 "--openaiorg `"$env:OPENAI_ORGANIZATION`" --openaideployment `"$env:AZURE_OPENAI_EMB_DEPLOYMENT`" " + `
 "--openaimodelname `"$env:AZURE_OPENAI_EMB_MODEL_NAME`" --index $env:AZURE_SEARCH_INDEX " + `
-"--formrecognizerservice $env:AZURE_FORMRECOGNIZER_SERVICE --tenantid $env:AZURE_TENANT_ID -v"
+"--formrecognizerservice $env:AZURE_FORMRECOGNIZER_SERVICE --tenantid $env:AZURE_TENANT_ID" + `
+"--openaihost $env:OPENAI_HOST --openaikey $env:OPENAI_API_KEY -v"
```
pamelafox, line 37: I see host and key on lines 32-33 so I think they didnt get dropped from this one?

pamelafox (changes requested):
Thanks so much, 1 comment

scripts/prepdocs.ps1
wdfox: My mistake. You’re spot on. Only the `prepdocs.sh` file needs to be changed. Just removed the unnecessary args from the Powershell script.

pamelafox (approved):

scripts/prepdocs.ps1
namratam30: I am confused between OpenAI Orgnization and openaihost. Could you please help me how to pass these details-
OPENAI_API_TYPE = "azure"
OPENAI_API_VERSION = "2023-07-01-preview"
OPENAI_API_KEY = "xxxxxxx"
OPENAI_API_BASE = " https://xxxxxx.openai.azure.com/"

pamelafox: Replied to your comment on other thread

PR #815, opened 2023-10-17
Support windows by using uvicorn without extras

### LINKED ISSUE #762: Local start does not work for Windows

```
- [ ] feature request
```

run app/start.ps1 on windows computer

Windows 11

1.4.1

If the uvloop and uvicorn are commented out from the backend/requirements.txt the application will come up.

pamelafox:
## Purpose

We were previously installing uvicorn[standard] which includes a package (uvloop) that doesn't work on Windows. That'd be left out for anyone who did a straight `pip install uvicorn[standard]` on Windows, since there are environment markers in their setup that tell pip not to install uvloop on Windows. However, now that we're using an already compiled requirements.txt that I made on a Linux machine, it no longer has the environment markers, so everyone gets uvloop.

This is a drawback of using the pip-tools compile strategy. I'm not sure a better way, however.

Anyway, I've changed it to only install `uvicorn`, not `uvicorn[standard]`, as we don't seem to need those extras anyway - we only need the UvicornWorker class to work with gunicorn. I did a test deploy to App Service with the changed requirements, and all seems to be working.

Fixes #762 

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[X] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test

*  azd deploy, app should still work


pamelafox:
I've also added "windows-latest" to the GitHub actions matrix. Unfortunately I had to exclude it from running the tests due to a funky network connection issue with the unit tests (tiktoken makes a request to get the model BPE), but hopefully we can get that working in the future. Most tests pass, it's only the first test that attempts the network request that fails, so that's a good sign for windows.

.gitattributes
```diff
@@ -1 +1,2 @@
-*.sh text eol=lf
\ No newline at end of file
+*.sh text eol=lf
+*.jsonlines text eol=lf
```
pamelafox, line 2: This is needed for when we eventually can run all the tests, for the snapshot tests comparison (to avoid it comparing CRLF to LF)

tonybaloney (approved):

PR #816, opened 2023-10-17
Add front-matter needed for Sample browser

pamelafox:
## Purpose

This sample is not currently discoverable in the Azure Samples browser:
https://learn.microsoft.com/en-us/samples/browse/

We'd like all the OpenAI chat sample apps to be indexed there, but that requires adding some YAML to README.md.

Instructions here:

https://review.learn.microsoft.com/en-us/help/contribute/samples/process/onboarding?branch=main#add-metadata-to-readme

Unfortunately, the YAML will get rendered in the README.md. We've given feedback to the Learn.com that we don't want to have it visible, but for now, we don't see a way to hide it.

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test

* View readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ products:
 - azure-app-service
 - azure-openai
 urlFragment: azure-search-openai-demo
-name: ChatGPT + Enterprise data with Azure OpenAI and Cognitive Search
+name: ChatGPT + Enterprise data
 description: Deploy an app to chat on top of enterprise data using Azure OpenAI and Cognitive Search.
 ---
 
--- a/README.md
+++ b/README.md
@@ -5,15 +5,13 @@ languages:
 - typescript
 - python
 - bicep
-- html
 products:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,3 +1,19 @@
...
+description: Chat with your data using OpenAI and Cognitive Search.
+languages:
+- azdeveloper
+- typescript
+- python
+- bicep
+products:
+- azure
+- azure-app-service
+- azure-openai
```
chuwik, line 12: Can you please add ACS here too?

chuwik (changes requested):

README.md
pamelafox: Yes, great point!

chuwik (approved):

PR #831, opened 2023-10-19
Adding langchain-based approaches with a fix

Yun-BizQA:
## Purpose

* Main changes are on backend\approaches\readdecomposeask.py and readdecomposeask.py.
* This is to add back 2 langchain-based approaches which were removed 2 weeks ago at #738. The issue was caused by Azure deployment storing model name as "gpt-35-turbo" whereas old langchain api required "gpt-3.5-turbo" with dot and some extra args. This can be resolved by using another langchain api _OpenAIChat_. There is also a possibility for unfixed issues of gpt-3.5's output parser so handle_parsing_errors=True needs to be added to _AgentExecutor.from_agent_and_tools_ api.
* However, while reverting all deleted components, there are later changes like the new Chat Request of the Chat App protocol, which I feel harder to adapt. @pamelafox Could you help add back the two approaches with the rest of codes updated and refined to work?

```
[ ] Yes
[x ] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[x] Other... Please describe:
Adding back deleted features
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
```

## What to Check
* ...


pamelafox:
Interesting, I'm surprised that was the only fix, as I had to make a lot more changes when I tried to fix them, but maybe those agents finally got updated with ChatCompletion support in the last few months. 

* Can you add tests to show how they're supposed to work? See CONTRIBUTING.md
* Have you tested with non-Azure OpenAI? We now support both. See the other approaches for examples of how we alternate between Azure and non-Azure

I can adapt to chat protocol if you can do those things.

Yun-BizQA:
- Can you add tests to show how they're supposed to work? See CONTRIBUTING.md:
I tested them using old code. Give me a day to try to add test code here.

- Have you tested with non-Azure OpenAI? We now support both. See the other approaches for examples of how we alternate between Azure and non-Azure
I have, but using non-Azure OpenAI means getting rid of langchain utilitieis right? I thought we wanted to keep langchain with its benefits. @srbalakr

pamelafox:
Hm, why wouldn't you be able to use the langchain approaches with non-Azure OpenAI? I think/hope it just requires different configuration.

Yun-BizQA:
I have used latest API AzureChatOpenAI that takes argument openai.api_type, which works for both Azure and Non-Azure models.
I have also added test cases to check both approaches run without errs.
@pamelafox Could you take another look please?

PR #832, opened 2023-10-19
Improve follow-up questions and pipe into context

### LINKED ISSUE #783: Follow-up questions are not reliably rendered as links
```
- [X] bug report -> please search issues before submitting
- [ ] feature request
```

The prompt currently instructs ChatGPT to render follow-up questions with angle brackets `<<` and `>>`.  Unfortunately, ChatGPT does not format the questions in a reliable manner. Sometimes it puts them inside angle brackets as requested, but it often puts them in a numbered list. That means the UI is not always able to render them as clickable links.

I've discussed follow-up questions with colleagues working on similar codebases, and we've identified three approaches to receiving follow-up questions in a well structured way so that they can become links in the UI:

* Post-processing the response to look for the variations in how ChatGPT responds. That's fairly brittle however. (Approach used by [the Javascript sample](https://github.com/Azure-Samples/azure-search-openai-javascript/blob/67dac3efd39145cc0c28835d41c5c39e1e48b428/packages/chat-component/src/utils/index.ts#L47)) 
* Using the "function call" feature of ChatGPT to receive a list of follow-up questions in the response.  (Which I just prototyped for this repo [in a branch](https://github.com/Azure-Samples/azure-search-openai-demo/compare/main...pamelafox:azure-search-openai-demo:followup?expand=1))
* Make an additional call to ChatGPT to specifically ask it to generate follow-up questions. (Approach used by [the C# sample](https://github.com/Azure-Samples/azure-search-openai-demo-csharp/blob/b14d971be1a9305ad9c76351967138d9f132ef74/app/backend/Services/ReadRetrieveReadChatService.cs#L33))

Each approach adds complexity in different parts of the codebase, so we haven't determined which way is the best.

pamelafox:
## Purpose

Fixes #783 

Our current prompt results in ChatGPT returning the follow-up questions in assorted formats, and then the frontend fails to render them properly as links.

The JavaScript sample team figured out a better prompt that does return the follow-up questions reliably. I took that prompt and replicated the improvements. However, they were glitching while streaming, due to the angle brackets showing up before they could be turned into links. 

So I added parsing in the *backend* code to extract the followup questions, and I return them inside the context as a list of strings. That makes the backend code more complex but makes the frontend code simpler and reduces computation time in the client.

I added unit tests, integration tests, and e2e tests to test the new functionality.

I did also experiment with using function_call for the follow-up questions ([in this branch](https://github.com/Azure-Samples/azure-search-openai-demo/compare/main...pamelafox:azure-search-openai-demo:followup?expand=1)), but that greatly complicated the code due to the need to parse function_call arguments from the streamed chunks of JSON bits. I think we should avoid streaming function call results unless absolutely needed.

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[X] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test

* Run the tests
* Manual test: Select "Suggest follow-up questions", ask question, see questions as links, click a question to ask it.


srbalakr (approved):

app/backend/approaches/chatreadretrieveread.py
```diff
@@ -29,14 +30,17 @@ class ChatReadRetrieveReadApproach(Approach):
...
+Each source has a name followed by colon and the actual information, always include the source name for each fact you use in the response. Use square brackets to reference the source, for example [info1.txt]. Don't combine sources, list each source separately, for example [info1.txt][info2.pdf].
 {follow_up_questions_prompt}
 {injected_prompt}
 """
-    follow_up_questions_prompt_content = """Generate three very brief follow-up questions that the user would likely ask next about their healthcare plan and employee handbook.
-Use double angle brackets to reference the questions, e.g. <<Are there exclusions for prescriptions?>>.
-Try not to repeat questions that have already been asked.
-Only generate questions and do not generate any text before or after the questions, such as 'Next Questions'"""
+    follow_up_questions_prompt_content = """Generate 3 very brief follow-up questions that the user would likely ask next.
```
srbalakr, line 37: Is this from other JS sample ?

pamelafox: Yep! https://github.com/Azure-Samples/azure-search-openai-javascript/pull/76/commits/0fd0d7cacfe2aeca4aba90febff8f1f3e31bc02f

PR #834, opened 2023-10-19
Run prettier on files and add to precommit/CI

pamelafox:
## Purpose

Now that we have the prettier VS Code extension auto-installed (which is great), it's introducing fixes to previously checked in files (mostly changes from last few months, as I think original author used prettier). This PR adds prettier to pre-commit and CI, and runs it on the existing files.

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test

* CI should pass
* tests should pass


.github/workflows/frontend.yaml
```diff
@@ -0,0 +1,17 @@
...
+  pull_request:
+    branches: [ main ]
+
+jobs:
+    prettier:
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/checkout@v4
+            - name: Install dependencies
+              run: npm install
```
chuwik, line 15: What package.json will this install? Will the next npx command run that version too?

Ideally we want it to be the one in app/frontend/package.json, so that it install and runs the same version devs use locally. Otherwise if npx uses the latest available version, it may be ahead and propose different changes.

chuwik (approved):
Indeed, used prettier originally :) Thanks for adding full support now!

.github/workflows/frontend.yaml
chuwik: In the past I've added a prettier script in package.json, installed it and ran that command with npm run prettier. Not sure if there's a more 'Github workflow compatible' way of doing it.

pamelafox: Ah good point. Let me tinker.

pamelafox: I changed it to do everything inside "app/frontend" so it picks up the config from there. Also added a prettierignore there to ignore the JSON files, don't want to autoformat package-lock.json.

PR #835, opened 2023-10-20
Store an MD5 hash of uploaded/indexed file and check before prepdocs

tonybaloney:
## Purpose
Running prep docs on every provision is really time consuming, especially if the docs haven't changed.

This PR writes an `.md5` file with an MD5 hash of the files in `data/` when they get uploaded, then every time the prep docs script is re-run that file is checked against the current hash and the file is skipped if it hasn't changed.

I've added those md5 files to gitignore so they don't get committed and they also won't be uploaded.

```
[ ] Yes
[ ] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
```

## What to Check
* ...

### DIFF
--- a/scripts/prepdocs.py
+++ b/scripts/prepdocs.py
@@ -525,7 +525,7 @@ def read_files(
                 with open(filename, 'rb') as file:
                     existing_hash = hashlib.md5(file.read()).hexdigest()
                 if os.path.exists(filename + ".md5"):
-                    with open(filename + ".md5", "r", encoding="utf-8") as md5_f:
+                    with open(filename + ".md5", encoding="utf-8") as md5_f:
                         stored_hash = md5_f.read()
 
                 if stored_hash and stored_hash.strip() == existing_hash.strip():
--- a/scripts/prepdocs.py
+++ b/scripts/prepdocs.py
@@ -527,14 +527,14 @@ def read_files(
                 if os.path.exists(filename + ".md5"):
                     with open(filename + ".md5", "r", encoding="utf-8") as md5_f:
                         stored_hash = md5_f.read()
-                else:
-                    # Write the hash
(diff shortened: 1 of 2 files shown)


tonybaloney:
screenshot 2023-10-20 at 13 32 23
screenshot of this working..

pamelafox:
Yay! Also adding @mattmsft as a reviewer as we discussed the issue together about avoiding repetitive indexing, and also because Matt has prepdocs-related changes planned in the future. Looks like you were able to make this change fairly unobtrusively.

scripts/prepdocs.py
```diff
@@ -515,6 +516,27 @@ def read_files(
                 read_files(filename + "/*", use_vectors, vectors_batch_support)
                 continue
             try:
+                # if filename ends in .md5 skip
+                if filename.endswith(".md5"):
+                    print("Skipping md5 hash index.")
```
pamelafox, line 521: This looks fairly noisy in logs and may confuse folks, do you think its necessary?

tonybaloney: Not really, removed

mattmsft (approved):
this looks good to me, we can add tests at a later point

pamelafox (approved):

scripts/prepdocs.py
```diff
@@ -515,6 +516,27 @@ def read_files(
...
+                    continue
+
+                # if there is a file called .md5 in this directory, see if its updated
+                stored_hash = None
+                with open(filename, "rb") as file:
+                    existing_hash = hashlib.md5(file.read()).hexdigest()
```
cryptoswarm, line 526: This should be called something like : new_hash. Here we are calculating a hash of a new file.

PR #840, opened 2023-10-20
Render entire PDFs instead of single pages

### LINKED ISSUE #819: Enhancing User Experience with Access to Original Files
```
- [x] feature request
```

### Description:
Post-retrieval of the initial answer, it's likely users might wish to access the entire document for a comprehensive view. To elevate this user experience, I propose the following enhancements:

### Direct Download Link:
Providing users with a direct link to download the original file ensures they can access the complete content whenever needed.

### Improved Analysis Panel Interaction:
An even more seamless experience would be to upload only the original PDF file, and within the analysis panel, display the PDF, auto-scrolling directly to the relevant page number.

Implementing these changes could make the system more intuitive and user-friendly, thereby improving overall user satisfaction.

pamelafox:
## Purpose

Fixes #819 

According to various user studies and requests, users would generally prefer to see the entire PDF, so that they can see the context, but of course, they'd also like the PDF to start off on the relevant page. Fortunately, browsers support a pseudo-standard of "#page=N" for jumping to a particular page.

This PR changes prepdocs.py to skip the step of uploading individual pages and link to #page=N instead of filename-N.pdf, and then makes sure the content_file route can still load those.

Screenshot 2023-10-20 at 4 07 41 PM

I added tests for the content_file route as well. They're running slowly now (though passing), am hoping to get guidance from Azure SDK team as to improving my mocking approach.

This should still work with indices that have individual pages.

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[ ] Bugfix
[X] Feature
[ ] Other... Please describe:
```

## How to Test

* Test with a fresh environment, azd up, ask a question, click on citation, see it jump
* Test with an old environment, dont re-index, ask a question, click on citation, see it load whole page


pamelafox:
@tonybaloney @mattmsft You may both be interested in the mocking approach I used for this one, lower-level than what we're using for other ones by overriding the Transport class used for the network requests. It works well except for the fact that it's taking longer than other tests, am not sure if that's due to it making a hidden network request or waiting for some timeout somewhere. Maybe @tonybaloney and I can take a look next week.

pamelafox:
Update: I profiled the tests using "yappi" and figured out that the Azure Python SDK was still making network requests due to its default retry policy. I added retry_total=0 (suggestion from SDK team engineer) and now the tests run very fast as expected. So this is ready for review.
@mattmsft This PR does change prepdocs, so if you want to wait for your PR, I can refactor when yours in.

data/employee_handbook.pdf
chuwik, line 1: Deleted by mistake?

pyproject.toml
```diff
@@ -2,14 +2,17 @@
 target-version = "py38"
 select = ["E", "F", "I", "UP"]
 ignore = ["E501", "E701"] # line too long, multiple statements on one line
-src = ["app/backend"]
+src = ["app/backend", "scripts"]
+
+[tool.ruff.isort]
+known-local-folder = ["scripts"]
```
pamelafox, line 8: fyi @mattmsft This known-local-folder addition fixed the auto-import sorting that was lumping scripts together with third-party on the preplib files.

pamelafox:
Update: After bringing in Matt's prep docs refactor, I made a few changes and added full test coverage, so this now requires a re-review.

mattmsft:
Looks like playwright was flaky and failed with a timeout - otherwise LGTM

pyproject.toml
mattmsft: Thanks!

mattmsft (approved):

app/backend/app.py
```diff
@@ -69,9 +70,18 @@ async def assets(path):
 # *** NOTE *** this assumes that the content files are public, or at least that all users of the app
 # can access all the files. This is also slow and memory hungry.
 @bp.route("/content/<path>")
-async def content_file(path):
+async def content_file(path: str):
+    # Remove page number from path, filename-1.txt -> filename.txt
+    if path.find("#page=") > 0:
```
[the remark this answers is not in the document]
pamelafox: Yep! I just did another manual test to make sure, its working with an old env with individual pages.

PR #852, opened 2023-10-24
Fix order of few shots sent to MessageBuilder

### LINKED ISSUE #851: few shot reversed in get_messages_from_history

```
- [ ] feature request
```

pamelafox:
## Purpose

Fixes #851 

This PR fixes an issue where few shots were getting added in the reverse order, but also renames append_message to insert_message since the current naming can lead you to think it defaults to appending at the end.

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[X] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test

* I added a test that failed for previous code and now passes.

### DIFF
--- a/app/backend/approaches/chatreadretrieveread.py
+++ b/app/backend/approaches/chatreadretrieveread.py
@@ -335,12 +335,12 @@ If you cannot generate a search query, return just the number 0.
         message_builder = MessageBuilder(system_prompt, model_id)
 
         # Add examples to show the chat what responses we want. It will try to mimic any responses and make sure they match the rules laid out in the system message.
-        for shot in few_shots:
-            message_builder.append_message(shot.get("role"), shot.get("content"))
+        for shot in reversed(few_shots):
+            message_builder.insert_message(shot.get("role"), shot.get("content"))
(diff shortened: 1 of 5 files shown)


app/backend/approaches/chatreadretrieveread.py
```diff
@@ -335,12 +335,12 @@ def get_messages_from_history(
         message_builder = MessageBuilder(system_prompt, model_id)
 
         # Add examples to show the chat what responses we want. It will try to mimic any responses and make sure they match the rules laid out in the system message.
-        for shot in few_shots:
-            message_builder.append_message(shot.get("role"), shot.get("content"))
+        for shot in reversed(few_shots):
```
srbalakr, line 338: nitpick: do we need to reverse order for few shots ? Does it make any difference ?

pamelafox: Yep, that was actually the main point of this PR. The test will break without reversed() as they're getting put in wrong order now.

srbalakr (approved):

PR #862, opened 2023-10-25
Refactor prepdocs

mattmsft:
## Purpose

* Factor out global variables from prep docs script
* Introduce the concept of strategies to support multiple document ingestion strategies in the future
* Make prepdocs use async azure sdk classes

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
```

## What to Check
* All tests pass
* No regressions in prep docs related functionality


scripts/prepdocs/files.py
```diff
@@ -0,0 +1,749 @@
...
+
+class OpenAIEmbeddings(ABC):
+    SUPPORTED_BATCH_AOAI_MODEL = {"text-embedding-ada-002": {"token_limit": 8100, "max_batch_size": 16}}
+
+    async def create_embedding_arguments(self) -> dict[str, Any]:
+        raise NotImplementedError
+
+    def before_retry_sleep(self, retry_state):
+        if self.verbose:
+            print("Rate limited on the OpenAI embeddings API, sleeping before retrying...")
```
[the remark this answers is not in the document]
pamelafox: (The advantage would be no self.verbose check needed)

scripts/prepdocs/files.py
```diff
@@ -0,0 +1,749 @@
...
+        return {
+            "model": self.open_ai_model_name,
+            "deployment_id": self.open_ai_deployment,
+            "api_type": self.get_api_type(),
+            "api_key": await self.wrap_credential(),
+            "api_version": "2023-05-15",
+        }
+
+    def get_api_type(self) -> str:
+        return "azure_ad" if self.credential is AsyncTokenCredential else "azure"
```
pamelafox, line 163: hm. shouldnt this be isinstance? (vs. is?)
maybe I'm blanking on how Python works. Same with other is check.

scripts/prepdocs/files.py
```diff
@@ -0,0 +1,749 @@
...
+                return emb_response["data"][0]["embedding"]
+
+    async def create_embeddings(self, texts: List[str]) -> List[List[float]]:
+        if not self.disable_batch and self.open_ai_model_name in OpenAIEmbeddings.SUPPORTED_BATCH_AOAI_MODEL:
+            return await self.create_embedding_batch(texts)
+
+        return [await self.create_embedding_single(text) for text in texts]
+
+
+class AzureOpenAIEmbeddingService(OpenAIEmbeddings):
```
pamelafox, line 135: I wonder if we'll be able to replace these once OpenAI releases their latest SDK with OpenAi and AzureOpenAI. We'll see! Just a musing.

scripts/prepdocs/files.py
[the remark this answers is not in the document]
mattmsft: follow up please

scripts/prepdocs/files.py
mattmsft: you are right :)

scripts/prepdocs/files.py
```diff
@@ -0,0 +1,749 @@
...
+                    yield File(content=open(temp_file_path, "rb"), acls=acls)
+                except Exception as e:
+                    print(f"\tGot an error while reading {path.name} -> {e} --> skipping file")
+                    try:
+                        os.remove(temp_file_path)
+                    except Exception as e:
+                        print(f"\tGot an error while deleting {temp_file_path} -> {e}")
+
+
+class Page:
```
pamelafox, line 332: Page and SplitPage could use some docstrings explaining their usage? And what offset means.

scripts/prepdocs/files.py
```diff
@@ -0,0 +1,749 @@
...
+        self.text = text
+
+
+class SplitPage:
+    def __init__(self, page_num: int, text: str):
+        self.page_num = page_num
+        self.text = text
+
+
+class TextSplitter:
```
pamelafox, line 345: I wonder if chunking strategies should go in their own module? Might be nice to reduce file size if that organization makes sense.

scripts/prepdocs/files.py
```diff
@@ -0,0 +1,749 @@
...
+        self.verbose = verbose
+
+    async def upload_blob(self, file: File):
+        async with BlobServiceClient(
+            account_url=self.endpoint, credential=self.credential
+        ) as service_client, service_client.get_container_client(self.container) as container_client:
+            if not await container_client.exists():
+                await container_client.create_container()
+
+                # if file is PDF split into pages and upload each page as a separate blob
```
pamelafox, line 535: This is the logic I changed in my PR which I can either merge now or redo based on this change. It actually reduces the code complexity.

scripts/prepdocs/files.py
pamelafox: Oh I see this is already a new module! Neat. Well, not sure if it should be split further, just an idea. You could look at -accelerator and https://github.com/microsoft/sample-app-aoai-chatGPT for inspiration, to imagine whether you'd want this in a separate file in a future world where we supported more file formats, for example.
Not strongly opinionated on this either way.

scripts/prepdocs/strategy.py
```diff
@@ -0,0 +1,31 @@
+from abc import ABC
```
pamelafox, line 1: This could use module docstring and class docstrings, particularly for Strategy since thats a generic name, not clear how it's to be used.

scripts/prepdocs/files.py
mattmsft: I do think splitting this across files is the right move. Still trying to fix errors though first

pamelafox (commented):
Overall looks good! Not 100% sure on which classes belongs in which module, but I love the way you abstracted to allow for different strategies. For tests, keep in mind you could also use the mechanism I used in the wholefile branch to mock out the entire transport (i.e. HTTP requests/responses), if that ends up being cleaner than mocking individual methods.

scripts/prepdocs/files.py
mattmsft: yes, that will be a good follow up

scripts/prepdocs/files.py
mattmsft: module-ified

scripts/prepdocs/strategy.py
mattmsft: good point

scripts/prepdocs/files.py
mattmsft: agreed, offset is definitely confusing

tonybaloney:
A welcome change 👍🏻 

It might be confusing whether `prepdocs` refers to `prepdocs.py` or `prepdocs/` but since scripts isn't a submodule (doesn't have `__init__.py`) this hasn't eventuated into anything weird. If someone were to add that file in the future it could cause havoc in a seemingly unrelated PR. I'd suggest renaming the `prepdocs/` folder to `prepdocslib` or something to avoid confusion.

scripts/prepdocs/files/filestrategy.py
```diff
@@ -0,0 +1,72 @@
...
+
+    async def setup(self, search_info: SearchInfo):
+        search_manager = SearchManager(search_info, self.search_analyzer_name, self.use_acls, self.embeddings)
+        await search_manager.create_index()
+
+    async def run(self, search_info: SearchInfo):
+        search_manager = SearchManager(search_info, self.search_analyzer_name, self.use_acls, self.embeddings)
+        if self.document_action == DocumentAction.Add:
+            files = self.list_file_strategy.list()
+            async for file in files:
```
tonybaloney, line 55: love this use of async generators, it should be pretty snappy now

scripts/prepdocs.py
```diff
@@ -1,623 +1,127 @@
...
+    print("Processing files...")
+    if args.datalakestorageaccount:
+        adls_gen2_creds = credential if is_key_empty(args.datalakekey) else args.datalakekey
+        print(f"Using Data Lake Gen2 Storage Account {args.datalakestorageaccount}")
+        list_file_strategy = ADLSGen2ListFileStrategy(
+            data_lake_storage_account=args.datalakestorageaccount,
+            data_lake_filesystem=args.datalakefilesystem,
+            data_lake_path=args.datalakepath,
+            credential=adls_gen2_creds,
+            verbose=args.verbose,
```
tonybaloney, line 86: There is no `verbose` argument in the ADLSGen2ListFileStrategy init method

scripts/prepdocs/files/blobmanager.py
```diff
@@ -0,0 +1,79 @@
...
+        ) as service_client, service_client.get_container_client(self.container) as container_client:
+            if not await container_client.exists():
+                return
+            if path is None:
+                blobs = container_client.list_blob_names()
+            else:
+                prefix = os.path.splitext(os.path.basename(path))[0]
+                blobs = container_client.list_blob_names(name_starts_with=os.path.splitext(os.path.basename(prefix))[0])
+            async for b in blobs:
+                if path is not None and not re.match(f"{prefix}-\d+\.pdf", b):
```
tonybaloney, line 68: If path is None then `prefix` would be unbound here? You'll get an unbound local name exception

scripts/prepdocs/files/listfilestrategy.py
```diff
@@ -0,0 +1,170 @@
...
+    """
+
+    def __init__(self, content: IO, acls: Optional[dict[str, list]] = None):
+        self.content = content
+        self.acls = acls or {}
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self):
```
tonybaloney, line 29: `__exit__` [takes 3 arguments](https://docs.python.org/3/reference/datamodel.html#object.__exit__) (exception type, exception info and traceback). Add `*args` here if you don't care about them, without it this will raise an exception instead of closing the file handle.

scripts/prepdocs/files/listfilestrategy.py
```diff
@@ -0,0 +1,170 @@
...
+                        if acl_parts[0] == "user" and "r" in acl_parts[2]:
+                            acls["oids"].append(acl_parts[1])
+                        if acl_parts[0] == "group" and "r" in acl_parts[2]:
+                            acls["groups"].append(acl_parts[1])
+                    yield File(content=open(temp_file_path, "rb"), acls=acls)
+                except Exception as e:
+                    print(f"\tGot an error while reading {path} -> {e} --> skipping file")
+                    try:
+                        os.remove(temp_file_path)
+                    except Exception as e:
```
tonybaloney, line 169: Use a different variable name than `e` so its not confused with the surrounding scope

scripts/prepdocs.py
mattmsft: It's interesting mypy did not catch this, will fix

scripts/prepdocs/files/blobmanager.py
mattmsft: mypy didn't flag this because there's a specific `path is not None` check before using prefix. But i can make this more clear by adding a check for prefix not being none

scripts/prepdocs/files/listfilestrategy.py
mattmsft: super useful feedback - will make this change

mattmsft:
> A welcome change 👍🏻
> 
> It might be confusing whether `prepdocs` refers to `prepdocs.py` or `prepdocs/` but since scripts isn't a submodule (doesn't have `__init__.py`) this hasn't eventuated into anything weird. If someone were to add that file in the future it could cause havoc in a seemingly unrelated PR. I'd suggest renaming the `prepdocs/` folder to `prepdocslib` or something to avoid confusion.

This ended up being super confusing to resolve with mypy. Perhaps I'm not understanding something about how mypy or relative imports work. Thanks @pamelafox  for your help resolving this offline

app/mypy.ini
```diff
@@ -0,0 +1,7 @@
+[mypy]
+
+[mypy-msal.*]
+ignore_missing_imports = True
+
+[mypy-msal_extensions.*]
+ignore_missing_imports = True
```
pamelafox, line 7: Ah interesting, this had to be moved inside app/ since we're running mypy relative to each folder now. I think in theory you could point mypy at pyproject.toml or put an inner pyproject.toml in app/, but that might be confusing. This seems fine. I'm pretty torn about the best way to have multiple Python projects in one repo. Maybe we'll move to editable installs in future, but that could raise barrier for developers not used to them.

pamelafox (approved):
Let's merge and watch for related issues.

mattmsft:
> Let's merge and watch for related issues.

Agreed, there may be some

PR #866, opened 2023-10-26
Using the option --storagekey for the prepdocs script might fail

IRooc:
When using the option --storagekey for the prepdocs script the value might contain `==` base64 padding at the end. This will fail to successfully login because the script just removes the `=` signs during the split action. Copied the version from the app/start.ps1 which is better suited here.

## Purpose

* make the prepdocs script work with the --storagekey parameter with the value `$env:AZURE_STORAGE_KEY` that contain base64 padding characters

```
[ ] Yes
[ X] No
```

## Pull Request Type

```
[ X] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
* Get the code
* Get the StorageKey from the Azure Storage and make sure it has bas64 padding at the end (`==`)
* Add the AZURE_STORAGE_KEY to the azd environment `azd env set AZURE_STORAGE_KEY <<value>>`
* add the --storagekey $env:AZURE_STORAGE_KEY parameter to the script invocation in the prepdocs.ps1 file
* run the script it will complain about invalid padding because the equal signs have been stripped of
* with this fix it will work correctly

## What to Check
* processdocs script works


pamelafox:
Thanks! Assuming not an issue for shell script as that doesnt look for =.

CCing @jongio and @vhvb1989 in case there are other azd samples using same logic.

IRooc:
Just for the record, the shell script is fine and identical in both app/start and loadenv

pamelafox:
Thanks for confirming! Will merge Monday to avoid weekend merges.

pamelafox (approved):

PR #884, opened 2023-10-30
Handle errors better especially for streaming

### LINKED ISSUE #869: Triggering the AOAI content filter results in a "Type error" in the frontend
@tonybaloney -- I thought I should actually surface this as an issue out of the other thread.

If you ask a question in the frontend that triggers the content filter (e.g. "how do I make a bomb?") You get a "type error" in the frontend. (Please confirm)

I think this is because:
- the exception occurs in the OpenAI completion call
- that call is async and not actually called until _after_ execution of /chat has completed (return a 200) and streaming has started. (I noticed this stepping through debug)
- this means the exception is not caught by the except block in /chat, as you would expect it to at first glance.
- so the backend has returned the start of a stream, but no actual content because the first completion call (query generation) has errorred.
- the frontend receives an empty stream which fails type validation and hence you get a type error (instead of the actual helpful exception message)

This is what I _think_ is happening, but I'm still new to Quart and really still learning Python TBH. Be really keen to get your view.

I was also musing on how to solve it, I read that there was a generic catchall error handler in Flask, but not sure if there is one in Quart (@bp.error_handler?) Tried that but didn't work, another option would be to try/except in chatreadretrieveread around the completion blocks and return a legitimate (stream!) response, and when I realised that was a tuple of [extra_info, coroutine] I thought: this is getting too hard 😄!!

Hope you can shed some light.

_Originally posted by @leongj in https://github.com/Azure-Samples/azure-search-openai-demo/issues/849#issuecomment-1780969274_

pamelafox:
## Purpose

This PR overhauls the error handling and fixes it in the case of streaming. (Fixes #869)

Now, whenever an error happens, the UI will display a message that contains only the type (class) of the error, versus the entire error message/traceback. Generally, displaying entire error messages to users is considered a security risk, in case they contain private details, so this approach seems like a happy medium. Let me know if you'd prefer to show entire error.

Example of error when asking ChatGPT for unallowed topics:

Screenshot 2023-10-30 at 10 42 37 AM

Example of error when Azure Cognitive Search is above the semantic quota for the month:

Screenshot 2023-10-30 at 10 42 28 AM

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[X] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test

I added new unit tests, integration tests, and e2e tests. To test manually, the easiest trigger is (sadly) asking "How do i make a bomb?".

### DIFF
--- a/app/backend/app.py
+++ b/app/backend/app.py
@@ -40,6 +40,10 @@ CONFIG_CHAT_APPROACH = "chat_approach"
 CONFIG_BLOB_CONTAINER_CLIENT = "blob_container_client"
 CONFIG_AUTH_CLIENT = "auth_client"
 CONFIG_SEARCH_CLIENT = "search_client"
+ERROR_MESSAGE = """The app encountered an error processing your request.
+If you are an administrator of the app, view the full error in the logs. See aka.ms/appservice-logs for more information.
+Error type: {error_type}
+"""
 
 bp = Blueprint("routes", __name__, static_folder="static")
 
@@ -93,6 +97,10 @@ async def content_file(path: str):
     return await send_file(blob_file, mimetype=mime_type, as_attachment=False, attachment_filename=path)
 
 
+def error_dict(error: Exception) -> dict:
+    return {"error": ERROR_MESSAGE.format(error_type=type(error))}
+
+
 @bp.route("/ask", methods=["POST"])
 async def ask():
     if not request.is_json:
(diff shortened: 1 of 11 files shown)


app/backend/app.py
```diff
@@ -110,14 +118,18 @@ async def ask():
...
-        return jsonify({"error": str(e)}), 500
+    except Exception as error:
+        logging.exception("Exception in /ask: %s", error)
+        return jsonify(error_dict(error)), 500
 
 
 async def format_as_ndjson(r: AsyncGenerator[dict, None]) -> AsyncGenerator[str, None]:
-    async for event in r:
-        yield json.dumps(event, ensure_ascii=False) + "\n"
+    try:
```
pamelafox, line 127: Unfortunately, our current try/catch wasn't catching exceptions thrown from inside the async generator. I posted a question in the Quart forum to try to understand why that is (https://github.com/pallets/quart/discussions/285) but for now, this approach works.
Also, interestingly, an error while streaming means a 200 still, since status code gets sent first, but then an {error} JSON in the stream.
Not sure if that's standard for streaming protocols.

app/backend/app.py
```diff
@@ -40,6 +40,10 @@
 CONFIG_BLOB_CONTAINER_CLIENT = "blob_container_client"
 CONFIG_AUTH_CLIENT = "auth_client"
 CONFIG_SEARCH_CLIENT = "search_client"
+ERROR_MESSAGE = """The app encountered an error processing your request.
```
mattmsft, line 43: much nicer!

mattmsft (commented):
LGTM, will manually test to make sure

mattmsft (approved):
manual test LGTM

chuwik (approved):

leongj:
Looks great @pamelafox -- thank you!

PR #891, opened 2023-11-01
Automate Login Setup

mattmsft:
## Purpose

* Setup authentication with pre / post provision hooks
* Add AZURE_ENFORCE_ACCESS_CONTROL to setup mandatory access control when authentication is enabled
* Add support for integrated app services authentication

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[ ] Bugfix
[X] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
```

## What to Check
* Can deploy without any authentication, or with authentication, or with authentication and enforced access control


mattmsft:
add docs with table discussed offline

mattmsft:
add tests

mattmsft:
@charris-msft

pamelafox (changes requested):
I'm trying this out now! The first thing I notice is that the README needs updating. Definitely "Enabling authentication" in main README needs to show the `azd env set USE_AUTHENTICATION true` and then the Login MD needs updates with more `azd env set` and that table of yours.

scripts/auth_init.sh
```diff
@@ -0,0 +1,5 @@
+ #!/bin/sh
+
```
pamelafox, line 2: This is going to end up re-calling pip install even for users that aren't using authentication. This was a failing of my branch as well that you inherited.

I suggest adding an additional check here for whether USE_AUTHENTICATION is enabled before running pip install.

I'm open to other ideas for how to keep this a complete no-op for the no-auth user.

pamelafox: We might also consider -q to reduce the pip output overall, but I haven't played with it to see if it removes necessary output for debugging:
https://pip.pypa.io/en/stable/cli/pip/#cmdoption-q

tests/test_chatapproach.py
```diff
@@ -5,7 +5,7 @@
 
 def test_get_search_query():
     chat_approach = ChatReadRetrieveReadApproach(
-        None, "", "gpt-35-turbo", "gpt-35-turbo", "", "", "", "", "en-us", "lexicon"
+        None, None, "", "gpt-35-turbo", "gpt-35-turbo", "", "", "", "", "en-us", "lexicon"
```
pamelafox, line 8: At this point, you could make this a module-level fixture, as I think we reuse it a bunch of times. No big deal but have thought about that in previous PRs to this file.

scripts/auth_init.py
```diff
@@ -0,0 +1,177 @@
...
+            }
+        ],
+        "identifierUris": [f"api://{server_app_id}"],
+    }
+
+
+def create_client_app_payload(server_app_id: str, server_app_permission_setup_payload: Dict[str, Any]):
+    return {
+        "displayName": "Azure Search OpenAI Demo Client App",
+        "signInAudience": "AzureADandPersonalMicrosoftAccount",
```
jamesc, line 110: This will allow anyone in any organization with an AD tenant or anyone with a personal MSA account to try and  log in to the application.  Is this the intent?  I'd suggest setting this by default to `AzureADMyOrg` so only members of the current organization can connect.

mattmsft: @pamelafox  in your branch you used this constant, are you OK switching it to AzureADMyOrg?

infra/core/host/appservice.bicep
```diff
@@ -87,6 +88,41 @@ resource appService 'Microsoft.Web/sites@2022-03-01' = {
...
+      identityProviders: {
+        azureActiveDirectory: {
+          enabled: true
+          registration: {
+            clientId: appSettings.AZURE_CLIENT_APP_ID
+            clientSecretSettingName: !empty(appSettings.AZURE_CLIENT_APP_SECRET) ? 'AZURE_CLIENT_APP_SECRET' : ''
+            openIdIssuer: appSettings.AZURE_AUTHENTICATION_ISSUER_URI
+          }
+          login: {
+            loginParameters: ['scope=openid profile email offline_access api://${appSettings.AZURE_SERVER_APP_ID}/access_as_user']
```
jongio, line 109: I'd like to see this as a parameter to the appservice.bicep file so users could override if they need to.

infra/core/host/appservice.bicep
```diff
@@ -87,6 +88,41 @@ resource appService 'Microsoft.Web/sites@2022-03-01' = {
...
+          }
+          validation: {
+            allowedAudiences: ['api://${appSettings.AZURE_SERVER_APP_ID}']
```
jongio, line 112: Should this be a parameter as well?

infra/core/host/appservice.bicep
```diff
@@ -87,6 +88,41 @@ resource appService 'Microsoft.Web/sites@2022-03-01' = {
...
+            defaultAuthorizationPolicy: {
+              allowedApplications: []
```
jongio, line 114: Parameter?

scripts/auth_init.py
pamelafox: Yes, that's probably what most expect, they can broaden it if desired.

pamelafox:
Note: if you dont explicitly set use acls, you will get this error:
Error: "Message: Invalid expression: Could not find a property named 'oids' on type 'search.document'."

To fix that, run

./scripts/manageacl.sh --acl-action enable_acls

pamelafox:
We know this doesn't work well when you don't have consent and we need a way to surface this error better. @mattmsft

pamelafox:
I've now updated with changes that we came up with during yesterday's pairing session. I'm wondering if we need to support falling back to AZURE_TENANT_ID or if we should always explicitly have developers specify their auth tenant ID. What do you think @mattmsft ?

infra/core/host/appservice.bicep
mattmsft: We need these permissions at minimum, so it might be an "and" rather than an "or"

infra/core/host/appservice.bicep
mattmsft: In the current design I don't think this makes sense - the audience is always the server app

mattmsft:
> I've now updated with changes that we came up with during yesterday's pairing session. I'm wondering if we need to support falling back to AZURE_TENANT_ID or if we should always explicitly have developers specify their auth tenant ID. What do you think @mattmsft ?

We should default to the single tenant setup (AZURE_AUTH_TENANT_ID = AZURE_TENANT_ID), not everyone has a mutli-tenant setup like ours

app/backend/core/authentication.py
```diff
@@ -57,11 +59,14 @@ def __init__(
                 client_credential=server_app_secret,
                 token_cache=PersistedTokenCache(persistence),
             )
+        else:
```
pamelafox, line 62: Related to line 52, not this line: when deployed, App Service always logs the exception for me of "Encryption unavailable. Opting in to plain text." 
When I looked into that, I see the msal-extensions docs seem to indicate that it shouldn't be used on deployed machines-

https://pypi.org/project/msal-extensions/

"Note: It is recommended to use this library for cache persistance support for Public client applications such as Desktop apps only. In web applications, this may lead to scale and performance issues. Web applications are recommended to persist the cache in session. Take a look at this [webapp sample](https://github.com/Azure-Samples/ms-identity-python-webapp)."

Is it expected that it logs the exception in prod and uses FilePersistence instead? If its expected, should we change it to an info/warning? It adds a lot of noise to logs since it includes the full traceback.

app/backend/core/authentication.py
```diff
@@ -57,11 +59,14 @@ def __init__(
...
+            self.require_access_control = False
```
pamelafox, line 63: I also see a warning in the logs for line 56, so I've filed https://github.com/AzureAD/microsoft-authentication-library-for-python/issues/629 to get MSAL's recommendation for how to avoid that warning.

infra/main.bicep
```diff
@@ -66,12 +66,18 @@ param embeddingDeploymentName string // Set in main.parameters.json
...
+
 // Used for the optional login and document level access control system
 param useAuthentication bool = false
+param enforceAccessControl bool = false
 param serverAppId string = ''
 @secure()
 param serverAppSecret string = ''
 param clientAppId string = ''
+@secure()
+param clientAppSecret string = ''
```
pamelafox, line 80: This needs to be stored in Key Vault

infra/main.bicep
```diff
@@ -174,10 +180,14 @@ module backend 'core/host/appservice.bicep' = {
       OPENAI_ORGANIZATION: openAiApiOrganization
       // Optional login and document level access control system
       AZURE_USE_AUTHENTICATION: useAuthentication
+      AZURE_ENFORCE_ACCESS_CONTROL: enforceAccessControl
       AZURE_SERVER_APP_ID: serverAppId
       AZURE_SERVER_APP_SECRET: serverAppSecret
       AZURE_CLIENT_APP_ID: clientAppId
-      AZURE_TENANT_ID: tenant().tenantId
+      AZURE_CLIENT_APP_SECRET: clientAppSecret
```
pamelafox, line 187: This should become a keyvault reference instead.

app/backend/core/authentication.py
mattmsft: We don’t use sessions in our application. We can add support for this in app.py and remove the file persistence cache completely.

the encrypted one only works if you install g object which I found difficult

mattmsft: Discussed offline - replace with secure cookie sessions. Requires key vault support

infra/main.bicep
mattmsft: will update when we have key vault support

infra/main.bicep
mattmsft: will update when we have keyvault support

mattmsft:
> Note: if you dont explicitly set use acls, you will get this error: Error: "Message: Invalid expression: Could not find a property named 'oids' on type 'search.document'."
> 
> To fix that, run
> 
> ./scripts/manageacl.sh --acl-action enable_acls

i'll create an error on startup if you run this on an index without the correct fields

mattmsft:
> Note: if you dont explicitly set use acls, you will get this error: Error: "Message: Invalid expression: Could not find a property named 'oids' on type 'search.document'."
> 
> To fix that, run
> 
> ./scripts/manageacl.sh --acl-action enable_acls

generate an AuthError message more effectively for this case

infra/core/host/appservice.bicep
mattmsft: added

app/backend/core/authentication.py
mattmsft: got it, looks like this is going to be fixed in the msal library?

app/backend/requirements.txt
```diff
@@ -266,17 +266,23 @@ six==1.16.0
     #   azure-core
     #   isodate
     #   python-dateutil
+taskgroup==0.0.0a4
+    # via hypercorn
 tiktoken==0.5.1
     # via -r requirements.in
+tomli==2.0.1
```
pamelafox, line 273: Can you re-run compile in 3.11?

app/backend/core/authentication.py
pamelafox: Yep they merged a fix

pamelafox (approved):

PR #903, opened 2023-11-02
PoC: Resource-group scoped deployment

weikanglim:
Proof-of-concept of what a resource-group scoped OpenAI demo deployment would look like.


pamelafox:
Does this not work? I notice you closed it.

PR #907, opened 2023-11-05
Test and fix listfilestrategy.py

### LINKED ISSUE #905: Error on check_md5 with folder
```
- [ ] feature request
```

  File "\scripts\prepdocs.py", line 256, in <module>
    loop.run_until_complete(main(file_strategy, azd_credential, args))
  File "C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.10_3.10.3056.0_x64__qbz5n2kfra8p0\lib\asyncio\base_events.py", line 649, in run_until_complete
    return future.result()
  File "\scripts\prepdocs.py", line 131, in main
    await strategy.run(search_info)
  File "\scripts\prepdocslib\filestrategy.py", line 54, in run
    async for file in files:
  File "\scripts\prepdocslib\listfilestrategy.py", line 82, in list
    if not self.check_md5(path):
  File "\scripts\prepdocslib\listfilestrategy.py", line 92, in check_md5
    with open(path, "rb") as file:
PermissionError: [Errno 13] Permission denied: 'rootPath/data\\SomeFolder'

    def check_md5(self, path: str) -> bool:
        # if path is a folder skip it
        if os.path.isdir(path):
            return True

pamelafox:
## Purpose

Fixes #905 and adds full test coverage for relevant classes (File, LocalListFileStrategy).

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[X] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test

* New tests should pass
* For manual test, make dir in data and run prepdocs, it should index contents.

### DIFF
--- a/scripts/prepdocslib/listfilestrategy.py
+++ b/scripts/prepdocslib/listfilestrategy.py
@@ -23,12 +23,6 @@ class File:
         self.content = content
         self.acls = acls or {}
 
-    def __enter__(self):
-        return self
-
-    def __exit__(self, *args):
-        self.close()
-
     def filename(self):
         return os.path.basename(self.content.name)
 
@@ -48,11 +42,11 @@ class ListFileStrategy(ABC):
     """
 
     async def list(self) -> AsyncGenerator[File, None]:
-        if False:
+        if False:  # pragma: no cover - this is necessary for mypy to type check
             yield
 
     async def list_paths(self) -> AsyncGenerator[str, None]:
-        if False:
+        if False:  # pragma: no cover - this is necessary for mypy to type check
             yield
 
 
@@ -74,8 +68,9 @@ class LocalListFileStrategy(ListFileStrategy):
             if os.path.isdir(path):
                 async for p in self._list_paths(f"{path}/*"):
                     yield p
-
-            yield path
+            else:
+                # Only list files, not directories
+                yield path
 
     async def list(self) -> AsyncGenerator[File, None]:
         async for path in self.list_paths():
--- a/tests/test_prepdocs.py
+++ b/tests/test_prepdocs.py
@@ -1,29 +1,12 @@
-import io
-
 import openai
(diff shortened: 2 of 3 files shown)


scripts/prepdocslib/listfilestrategy.py
```diff
@@ -23,12 +23,6 @@ def __init__(self, content: IO, acls: Optional[dict[str, list]] = None):
         self.content = content
         self.acls = acls or {}
 
-    def __enter__(self):
```
pamelafox, line 26: We dont seem to be using this class as a context manager anywhere, so I deleted the code.

tests/test_prepdocs.py
```diff
@@ -1,29 +1,12 @@
...
 
-from scripts.prepdocslib.blobmanager import File
 from scripts.prepdocslib.embeddings import (
     AzureOpenAIEmbeddingService,
     OpenAIEmbeddingService,
 )
-from scripts.prepdocslib.listfilestrategy import ADLSGen2ListFileStrategy
-
-
-def test_filename_to_id():
```
pamelafox, line 16: Just moved files

scripts/prepdocslib/listfilestrategy.py
tonybaloney: this class looks like it should mostly be used as a context manager if it has a `.close()` method, so I'm surprised it's not being used in this way

tonybaloney (approved):

scripts/prepdocslib/listfilestrategy.py
pamelafox: I think it'd be possible for it to be used that way, but it'd be a larger change affecting a few files. Right now it's used inside a try/finally, where close() is called in the finally clause.

cc @mattmsft

PR #898, opened 2023-11-02
Added insights dashboard

pamelafox:
## Purpose

Adds a dashboard that can be pulled up via `azd monitor`.

Screenshot 2023-11-02 at 7 40 30 AM

## Does this introduce a breaking change?

```
[ ] Yes
[X] No
```

## Pull Request Type
What kind of change does this Pull Request introduce?

```
[ ] Bugfix
[X] Feature
[ ] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[ ] Documentation content changes
[ ] Other... Please describe:
```

## How to Test

* Run azd up, then run azd monitor. Run a loadtest to generate data and wait 5 minutes for it to show up.

PR #918, opened 2023-11-06
Upgrade infra/core to latest from azd

jongio:
Update the infra/core bicep to the latest security improvements

### DIFF
--- a/infra/core/monitor/applicationinsights.bicep
+++ b/infra/core/monitor/applicationinsights.bicep
@@ -1,6 +1,10 @@
+metadata description = 'Creates an Application Insights instance based on an existing Log Analytics workspace.'
 param name string
+param dashboardName string
 param location string = resourceGroup().location
 param tags object = {}
+param includeDashboard bool = true
+param logAnalyticsWorkspaceId string
 
 resource applicationInsights 'Microsoft.Insights/components@2020-02-02' = {
   name: name
@@ -9,6 +13,16 @@ resource applicationInsights 'Microsoft.Insights/components@2020-02-02' = {
   kind: 'web'
   properties: {
     Application_Type: 'web'
+    WorkspaceResourceId: logAnalyticsWorkspaceId
+  }
+}
+
+module applicationInsightsDashboard 'applicationinsights-dashboard.bicep' =  if (includeDashboard) {
+  name: 'application-insights-dashboard'
+  params: {
+    name: dashboardName
+    location: location
+    applicationInsightsName: applicationInsights.name
   }
 }
 
--- a/infra/core/storage/storage-account.bicep
+++ b/infra/core/storage/storage-account.bicep
@@ -1,24 +1,31 @@
+metadata description = 'Creates an Azure storage account.'
 param name string
 param location string = resourceGroup().location
 param tags object = {}
 
-@allowed([ 'Hot', 'Cool', 'Premium' ])
+@allowed([
+  'Cool'
+  'Hot'
+  'Premium' ])
 param accessTier string = 'Hot'
-param allowBlobPublicAccess bool = false
+param allowBlobPublicAccess bool = true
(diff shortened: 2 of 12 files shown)


infra/core/monitor/applicationinsights.bicep
```diff
@@ -9,6 +13,16 @@ resource applicationInsights 'Microsoft.Insights/components@2020-02-02' = {
   kind: 'web'
   properties: {
     Application_Type: 'web'
+    WorkspaceResourceId: logAnalyticsWorkspaceId
+  }
+}
+
+module applicationInsightsDashboard 'applicationinsights-dashboard.bicep' =  if (includeDashboard) {
```
pamelafox, line 20: I have an existing PR that adds a dashboard (and the related Bicep that's here), and my dashboard is pretty different than the azd one since I couldn't get browser and user stats to work. Do those stats actually work for all the azd samples? How?

pamelafox: My PR: https://github.com/Azure-Samples/azure-search-openai-demo/pull/898

infra/core/storage/storage-account.bicep
```diff
@@ -1,25 +1,32 @@
...
 param tags object = {}
 
-@allowed([ 'Hot', 'Cool', 'Premium' ])
+@allowed([
+  'Cool'
+  'Hot'
+  'Premium' ])
 param accessTier string = 'Hot'
-param allowBlobPublicAccess bool = false
+param allowBlobPublicAccess bool = true
```
pamelafox, line 11: This became true, so you'll need to pass it in as False in main.bicep

jongio: Done

infra/core/monitor/applicationinsights.bicep
jongio: It did work when we initially published, I haven't tested it in a while.  Would be good to see what diff you made and integrate into core if we can.

jongio: I think my appinsights core changes are the same as yours (other than dashboard), so I'll keep in mine, we can test and merge it, then get the latest in yours.  Or the other way around. LMK what works for you.

pamelafox: My dashboard is just vastly simplified to the stats that I could confirm were showing up, I don't think you'd want it in azd.

jongio: Can you figure out a way to add yours without overwriting the one in core?

pamelafox: I'm not sure that dashboards should be part of core? I would ideally customize our dashboard even more, to show openai-specific metrics, group by openai error rate, etc. It feels like something that should be passed in if that's possible.

And it's hard to have a "core" dashboard and add to it, since you literally have to configure the x/y position of each thing in the dashboard.

pamelafox: Okay so how about, we bring in your default dashboard, but don't actually use it? (i.e. includeDashboard stays at false) I think I can refer to my simplified dash in main.bicep . And if someone wants to refer to the default one for ideas, they can still do so.

pamelafox:
That storage security issue is noted here, by the way: https://github.com/Azure-Samples/azure-search-openai-demo/security/code-scanning/22

infra/core/monitor/applicationinsights.bicep
jongio: yes, FYI, 
The default dashboard was scraped from the default one you get when you click new dashboard in the portal.

jongio:
@pamelafox I tested these changes locally and it appears to all be working on my end.  It would be good if you could smoke it as well.

pamelafox (approved):
I ran `azd up` and this worked for me.

PR #970, opened 2023-11-18
Fix ignored capacity field for openAI deployment

vhvb1989:
The field `capacity` can't be directly used at the root level of a deployment object for the cognitive services module.

The module expects either the node `sku` or nothing:

```
sku: contains(deployment, 'sku') ? deployment.sku : {
    name: 'Standard'
    capacity: 20
  }
```

This PR won't change any functionality, as the default value for sku provided by the module is currently equal to the value set in main.bicep.  But fixes the bug for any user trying to use the capacity field with other value, as it is ignored


pamelafox (approved):
thanks!

pamelafox (approved):

PR #986, opened 2023-11-22
Fixing app service settings

pamelafox:
## Purpose

I've helped multiple developers debug an issue where their App Service was lacking environment variables, and I've seen it mostly the last few weeks. I'm thinking it's due to the change in https://github.com/Azure-Samples/azure-search-openai-demo/pull/918 that brought in the latest core templates, where appservice.bicep moved app settings into its own module. I'm not sure why exactly that'd break app service, but based on the timing and the issue, it seems likely.
Also, I just helped a developer fix their bug by re-provisioning with the old appservice.bicep.

Thus, I propose going back to the old approach of not putting the settings in a module.

FYI @jongio @wbreza 

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[X] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test

* azd up in a new environment, should work!

### DIFF
--- a/infra/core/host/appservice.bicep
+++ b/infra/core/host/appservice.bicep
@@ -65,6 +65,18 @@ resource appService 'Microsoft.Web/sites@2022-03-01' = {
 
   identity: { type: managedIdentity ? 'SystemAssigned' : 'None' }
 
+  resource configAppSettings 'config' = {
+    name: 'appsettings'
+    properties: union(appSettings,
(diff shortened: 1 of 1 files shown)


pamelafox:
cc @charris-msft I think this is also what you ran into.

mattmsft (approved):

PR #1022, opened 2023-11-30
Update launch.json with python backend path

pamelafox:
## Purpose

This PR adds the python property to the Quart launch.json configuration, so that the debugger selects Python interpreter in the correct virtual environment.

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[X] Other... Please describe:
```

## How to Test

* Without running pip install at the root, try to use the Quart debug configuration. It should work!


.vscode/launch.json
```diff
@@ -10,6 +10,7 @@
             "request": "launch",
             "module": "quart",
             "cwd": "${workspaceFolder}/app/backend",
+            "python": "${workspaceFolder}/app/backend/backend_env/bin/python",
```
tonybaloney, line 13: Is backend_env the name we use in a setup script?

pamelafox: yes, in https://github.com/Azure-Samples/azure-search-openai-demo/blob/a1170562395ad122d2be0e8e29d33cac7ff169c2/app/start.sh#L20
is there a less hard codey approach?

tonybaloney: If you leave it out, VS Code will select the configured interpreter or there is a special variable to inherit that. https://code.visualstudio.com/docs/python/debugging#_python

pamelafox: Ah I think I don't typically configure an interpreter for my workspace, so I assume it defaults to the Python of the dev container. That won't always work since people may not have also installed into the dev container directly. So I think we need it to be backend_env specifically, since we're assuming people are setting that up? Or, should there be a prerequisite step to the debugger that runs pip install at the root?

pamelafox: Ping on this discussion. Just working on my slides about local development and remembered this was pending.

pamelafox:
I've also updated "python" to "debugpy" as that recently changed in VSCode, "python" is on the deprecation path.

tonybaloney (approved):

PR #1062, opened 2023-12-14
Use AAD instead of key vault for Computer Vision API

pamelafox:
## Purpose

I've discovered from the computer vision team that we can use Bearer header with Azure credential token provider. I'm testing this change more thoroughly but initial experiments look good.

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[ ] Bugfix
[X] Feature
[ ] Other... Please describe:
```

## How to Test

* Testing azd up with GPT-4-V, both locally and deployed should work


pamelafox:
Update: I need to update prepdocs as well.

pamelafox:
This is blocked on the Azure credential bearer token not working in App Service. Works fine locally.

pamelafox:
Code to try it in SSH (failing now)

```
>>> import aiohttp
>>> from azure.identity import DefaultAzureCredential, get_bearer_token_provider
>>> azure_credential = DefaultAzureCredential(exclude_shared_token_cache_credential=True)
>>> token_provider = get_bearer_token_provider(azure_credential, "https://cognitiveservices.azure.com/.default")
>>> url = 'https://cog-cv-qn2a735t7pshk.cognitiveservices.azure.com/computervision/retrieval:vectorizeText?api-version=2023-02-01-preview&modelVersion=latest'
>>> headers = {"Content-Type": "application/json"}
>>> params = {"api-version": "2023-02-01-preview", "modelVersion": "latest"}
>>> data = {"text": 'puppy'}
>>> headers["Authorization"] = "Bearer " + token_provider()
>>> import requests
>>> response = requests.post(url, json=data, headers=headers, params=params)
```

app/backend/app.py
```diff
@@ -9,7 +9,6 @@
 
 from azure.core.exceptions import ResourceNotFoundError
 from azure.identity.aio import DefaultAzureCredential, get_bearer_token_provider
-from azure.keyvault.secrets.aio import SecretClient
```
mattmsft, line 12: Will we add this back as part of the key vault auth changes

infra/main.bicep
```diff
@@ -47,12 +46,12 @@ param openAiServiceName string = ''
...
 param useGPT4V bool = false
 
+param keyVaultResourceGroupName string = ''
 param keyVaultServiceName string = ''
-param computerVisionSecretName string = 'computerVisionSecret'
 param searchServiceSecretName string = 'searchServiceSecret'
 
 @description('Location for the OpenAI resource group')
-@allowed(['canadaeast', 'eastus', 'eastus2', 'francecentral', 'switzerlandnorth', 'uksouth', 'japaneast', 'northcentralus', 'australiaeast', 'swedencentral'])
+@allowed([ 'canadaeast', 'eastus', 'eastus2', 'francecentral', 'switzerlandnorth', 'uksouth', 'japaneast', 'northcentralus', 'australiaeast', 'swedencentral' ])
```
pamelafox, line 54: I ran az bicep format so you'll notice some spacing changes

pamelafox:
I realized I was missing a role, this now works for me locally and in production. @mattgotteiner Please try it out if you have the time.

app/backend/app.py
```diff
@@ -348,16 +345,16 @@ async def setup_clients():
     )
 
     if USE_GPT4V:
-        if vision_key is None:
-            raise ValueError("Vision key must be set (in Key Vault) to use the vision approach.")
+
+        token_provider = get_bearer_token_provider(azure_credential, "https://cognitiveservices.azure.com/.default")
```
mattgotteiner, line 349: did we still want to support keys for folks that didn't want to use RBAC? (not sure why you would, so I think it's OK to drop this)

infra/main.bicep
```diff
@@ -582,6 +573,17 @@ module searchReaderRoleBackend 'core/security/role.bicep' = if (useAuthenticatio
   }
 }
 
+// For computer vision access by the backend
```
mattgotteiner, line 576: did we need to add a role for access by principal in case you are bringing an existing vision account you didn't create?

scripts/prepdocslib/embeddings.py
```diff
@@ -192,15 +192,17 @@ class ImageEmbeddings:
...
         self.endpoint = endpoint
         self.verbose = verbose
 
     async def create_embeddings(self, blob_urls: List[str]) -> List[List[float]]:
-        headers = {"Ocp-Apim-Subscription-Key": self.credential}
-        params = {"api-version": "2023-02-01-preview", "modelVersion": "latest"}
         endpoint = urljoin(self.endpoint, "computervision/retrieval:vectorizeImage")
+        headers = {"Content-Type": "application/json"}
+        params = {"api-version": "2023-02-01-preview", "modelVersion": "latest"}
+        headers["Authorization"] = "Bearer " + await self.token_provider()
```
mattgotteiner, line 204: It's so interesting to me that this isn't actually documented, great find https://learn.microsoft.com/en-us/rest/api/computer-vision/?view=rest-computervision-2023-02-01-preview

mattgotteiner (commented):
looks good, thanks for making this change. will manually test to confirm. I'll sign off after my manual test

app/backend/app.py
pamelafox: I considered that, since it was already implemented, and since -accelerator is supporting both keys and RBAC. However, I think we agree that keyless auth is the future, and if we were to support keys for this API, we should support for all of them, which would significantly complicate code. We also know there are issues with using keyvault with azd, particularly in combination with CI/CD (due to the current user losing access). So we want to reduce use of key vault when possible.
So I decided to remove it. Existing users of GPT4V will be upgraded to managed identity when they re-provision.

PR #1082, opened 2023-12-20
Add more tests for prepdocslib

tonybaloney:
This PR adds some tests to load Pdf, use the local file strategy and use the text splitter.


tonybaloney:
This increases our overall coverage to 91%

pamelafox (approved):
Yay more tests!

PR #1083, opened 2023-12-21
gh definition and enhances

vhvb1989:
- Configure Azure Developer Pipeline
- Configure Azure Developer Pipeline
- Configure Azure Developer Pipeline
- Configure Azure Developer Pipeline
- Configure Azure Developer Pipeline
- Configure Azure Developer Pipeline

## Purpose

* ...

```
[ ] Yes
[ ] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

```
```

## What to Check
* ...


pamelafox:
@vhvb1989 How does the pipeline know the value of openAiResourceGroupLocation that was stored in the environment? (That's not in an env variable)

vhvb1989:
> @vhvb1989 How does the pipeline know the value of openAiResourceGroupLocation that was stored in the environment? (That's not in an env variable)

that's azd's magic :D
So, during `azd pipeline config` azd will prompt for a value for `openAiResourceGroupLocation` and azd will ask if you want to persist the value. When you click yes, the value is saved within the environment configuration (a json file next to the azure .env file).
Then, azd serializes the json file, environment configuration, and set it as a secret for your pipeline (using the name `AZD_INITIAL_ENVIRONMENT_CONFIG`). When azd runs provision, it uses the value from that secret to re-construct the json file as the initial environment configuration.
That means that the value you set during the prompt will be accessible from CI/CD, just like it is on your local machine.

Notes:
- If you want to change the selection, you need to either create a new environment, or manually edit the environment configuration. Azd doesn't have any gesture to manage the per-environment configuration.  After editing the value, you can run `azd pipeline config` again to refresh the secret for CI/CD
-  You must select `yes` when the prompt ask you about persisting the selection, otherwise, the selection will not be reflected on CI/CD.  I'll create a PR soon to mitigate this little detail :)

vhvb1989:
azd 1.6.0 is released, so this PR is ready to go :)

pamelafox (approved):

Issue #1051, opened 2023-12-12
Cookie-Based Token Cache for Authentication


mattgotteiner:
> Please provide us with the following information:
> ---------------------------------------------------------------

### This issue is for a: (mark with an `x`)
```
- [ ] bug report -> please search issues before submitting
- [X] feature request
- [ ] documentation issue or request
- [ ] regression (a behavior that used to work and stopped in a new release)
```

Authentication should use a cookie-based token cache so restarting the server doesn't log all users out

mattmsft:
I actually don't think this is the case - there's session based storage in the app that actually does sign on

mattmsft:
```
    def get_auth_setup_for_client(self) -> dict[str, Any]:
        # returns MSAL.js settings used by the client app
        return {
            "useLogin": self.use_authentication,  # Whether or not login elements are enabled on the UI
            "requireAccessControl": self.require_access_control,  # Whether or not access control is required to use the application
            "msalConfig": {
                "auth": {
                    "clientId": self.client_app_id,  # Client app id used for login
                    "authority": self.authority,  # Directory to use for login https://learn.microsoft.com/azure/active-directory/develop/msal-client-application-configuration#authority
                    "redirectUri": "/redirect",  # Points to window.location.origin. You must register this URI on Azure Portal/App Registration.
                    "postLogoutRedirectUri": "/",  # Indicates the page to navigate after logout.
                    "navigateToLoginRequestUrl": False,  # If "true", will navigate back to the original request location before processing the auth code response.
                },
                "cache": {
                    "cacheLocation": "sessionStorage",
                    "storeAuthStateInCookie": False,
                },  # Configures cache location. "sessionStorage" is more secure, but "localStorage" gives you SSO between tabs.  # Set this to "true" if you are having issues on IE11 or Edge
            },
```

Since the frontend app uses sessionStorage - I think the in memory cache for the backend app is just fine. The backend app doesn't actually manage any login state

PR #1117, opened 2024-01-05
Update Authentication to use local storage and easy auth refresh

mattgotteiner:
## Purpose

* Ensure tokens are refreshed when using easy auth so the full refresh window is used before requiring login again
* Store tokens in local storage so sessions are preserved across tabs

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[ ] Bugfix
[X] Feature
[ ] Other... Please describe:
```

## What to Check
*  Get the code
*  Validate .auth/refresh is called when logging into the app using the network tab on developer settings
* Validate auth_setup shows localStorage as the msal cache location

### DIFF
--- a/app/frontend/src/authConfig.ts
+++ b/app/frontend/src/authConfig.ts
@@ -89,35 +89,32 @@ export const getRedirectUri = () => {
 // Get an access token if a user logged in using app services authentication
 // Returns null if the app doesn't support app services authentication
 const getAppServicesToken = (): Promise<AppServicesToken | null> => {
-    return fetch(appServicesAuthTokenRefreshUrl)
-        .then(r => {
-            if (r.ok) {
-                return fetch(appServicesAuthTokenUrl)
-                    .then(r => {
-                        if (r.ok) {
-                            return r.json()
-                                .then(json => {
-                                    if (json.length > 0) {
-                                        return {
-                                            id_token: json[0]["id_token"] as string,
-                                            access_token: json[0]["access_token"] as string,
-                                            user_claims: json[0]["user_claims"].reduce((acc: Record<string, any>, item: Record<string, any>) => {
-                                                acc[item.typ] = item.val;
-                                                return acc;
-                                            }, {}) as Record<string, any>
-                                        };
-                                    }
-
-                                    return null;
-                                });
+    return fetch(appServicesAuthTokenRefreshUrl).then(r => {
+        if (r.ok) {
+            return fetch(appServicesAuthTokenUrl).then(r => {
+                if (r.ok) {
+                    return r.json().then(json => {
+                        if (json.length > 0) {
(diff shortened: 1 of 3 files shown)


mattmsft:
#1051

pamelafox:
I think some snapshot tests need updating to reflect change to localStorage?

app/frontend/src/authConfig.ts
```diff
@@ -88,18 +89,24 @@ export const getRedirectUri = () => {
 // Get an access token if a user logged in using app services authentication
 // Returns null if the app doesn't support app services authentication
 const getAppServicesToken = (): Promise<AppServicesToken | null> => {
-    return fetch(appServicesAuthTokenUrl).then(r => {
+    return fetch(appServicesAuthTokenRefreshUrl).then(r => {
```
pamelafox, line 92: So this is constantly refreshing? Is that the general recommendation? (Versus refreshing only based on expiration or some such?) Might be a naive question

mattmsft: There doesn't seem to be a solid recommendation on how often to do this

https://learn.microsoft.com/en-us/azure/app-service/configure-authentication-oauth-tokens#refresh-auth-tokens

mattmsft:
> I think some snapshot tests need updating to reflect change to localStorage?

How do you do this?

app/frontend/src/authConfig.ts
pamelafox: Okay, CC @jamesc in case he has any thoughts, but this seems fine to ship.

jamesc:
[like]  James Casey reacted to your message:
________________________________

jamesc:
[like]  James Casey reacted to your message:
________________________________

mirojs:
Hello @mattgotteiner @pamelafox 

Upon refreshing the app page from the browser (Chrome), the developer tool reported error:
GET https://app-backend-***.azurewebsites.net/.auth/refresh 404 (Not Found)
rs @ authConfig.ts:92
(anonymous) @ authConfig.ts:120

Besides, 
  const authSetup = await fetchAuthSetup(); Line 61
  export const appServicesToken = await getAppServicesToken(); Line 120
were labelled: 'await' function is only allowed within an async function

Could you help to check further? Thanks!

matthewgotteiner:
Thanks for reporting for.

regarding .auth not found - this is expected if easy auth is not setup, we might need to add more checks for this to avoid refreshing.

regarding the typescript compile error what version of node are you using?

mirojs:
@mattgotteiner Thanks for the quick response! My node version is v20.11.0

ntabernacle:
I was also getting .auth not found and this caused me issues with login. The fix was changing localStorage to sessionStorage in authentication.py

Issue #1152, opened 2024-01-18
Small documets are not indexing


RezaMatin:
> Please provide us with the following information:
> ---------------------------------------------------------------

### This issue is for a: (mark with an `x`)
```
- [X] bug report -> please search issues before submitting
- [ ] feature request
- [ ] documentation issue or request
- [ ] regression (a behavior that used to work and stopped in a new release)
```

### Minimal steps to reproduce
init
azd up
azd deploy

### Any log messages given by the failure
No error, empty search index entry

### Expected/desired behavior
Search index should be field from given PDF file

### OS and Version?
> Windows 7, 8 or 10. Linux (which distribution). macOS (Yosemite? El Capitan? Sierra?)
Created from code space

### azd version?
> run `azd version` and copy paste here.
1.5.1
### Versions
>

### Mention any other details that might be useful

> ---------------------------------------------------------------
> Thanks! We'll be in touch soon.

mattgotteiner:
Hi @RezaMatin, we need some more details to understand this issue

1. When you say small documents, what do you mean? Are these short pdfs, text files, something else?
2. Are you using the existing sample data as well or have you totally replaced it? Does your app work with the sample data?
Thanks

RezaMatin:
Hi @mattgotteiner  
Thanks for your reply. Everything works fine with the included data. I have removed the data and tested it with different PDF sizes. PDF document with just a few lines does not appear in the search index.
Thanks

marekjakimiuk1:
> Hi @RezaMatin, we need some more details to understand this issue
> 
> 1. When you say small documents, what do you mean? Are these short pdfs, text files, something else?
> 2. Are you using the existing sample data as well or have you totally replaced it? Does your app work with the sample data?
>    Thanks

Hi @mattgotteiner and @RezaMatin ,

I've pushed fix for this bug: https://github.com/Azure-Samples/azure-search-openai-demo/pull/1155

Best regards,
Marek

Issue #1153, opened 2024-01-18
Document is not indexed when section_overlap is bigger than the text size


marekjakimiuk1:
> Please provide us with the following information:
> ---------------------------------------------------------------
When section_overlap is bigger than text size than the split_pages function in textsplitter.py is not yielding any SplitPage. The result is that the document is not being indexed.

### This issue is for a: (mark with an `x`)
```
- [ X] bug report -> please search issues before submitting
- [ ] feature request
- [ ] documentation issue or request
- [ ] regression (a behavior that used to work and stopped in a new release)
```

### Minimal steps to reproduce
init
azd up
azd deploy

### Any log messages given by the failure
Empty search index entry. No error logs.

### Expected/desired behavior
Search index should be field from given file

### OS and Version?
> Windows 7, 8 or 10. Linux (which distribution). macOS (Yosemite? El Capitan? Sierra?)
Azure services creaded by the source code.

### azd version?
1.5.1

### Versions
>

### Mention any other details that might be useful

> ---------------------------------------------------------------
> Thanks! We'll be in touch soon.

mattgotteiner:
Thanks - this seems like a bug. we'll take a look

PR #1155, opened 2024-01-18
bug #1152: Fix for indexing small docs

marekjakimiuk1:
## Purpose

* ...

```
[ X] Yes
[ ] No
```

## Pull Request Type

It solves the problem with indexing small documents, where amount of text is less than section_overlap. In this case it's enought to check if  amount of text is less than max_section_length and return only one section without additional processing. 
This approach assumes that the section_overlap < max_section_length

```
[X] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test
*  Get the code
Index document whith amount of text less than max_section_length

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

* Test the code

Index document whith amount of text less than max_section_length

## What to Check
After indexing document whith amount of text less than max_section_length, check if document content has been added to the search index

### DIFF
--- a/scripts/prepdocslib/textsplitter.py
+++ b/scripts/prepdocslib/textsplitter.py
@@ -44,7 +44,7 @@ class TextSplitter:
         length = len(all_text)
 
         if length <= self.max_section_length:
-            yield (all_text, find_page(0))
+            yield SplitPage(page_num=find_page(0), text=all_text)
         else:
             start = 0
             end = length
--- a/scripts/prepdocslib/textsplitter.py
+++ b/scripts/prepdocslib/textsplitter.py
@@ -1,5 +1,4 @@
 from typing import Generator, List
-
 from .pdfparser import Page
 
 
--- a/scripts/prepdocslib/textsplitter.py
+++ b/scripts/prepdocslib/textsplitter.py
@@ -40,8 +40,10 @@ class TextSplitter:
             return pages[num_pages - 1].page_num
 
         all_text = "".join(page.text for page in pages)
-        length = len(all_text)
+        if len(all_text.strip()) == 0:
+            return
 
+        length = len(all_text)
         if length <= self.max_section_length:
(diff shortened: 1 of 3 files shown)


scripts/prepdocslib/textsplitter.py
```diff
@@ -42,59 +42,63 @@ def find_page(offset):
...
-        while start + self.section_overlap < length:
-            last_word = -1
-            end = start + self.max_section_length
-
-            if end > length:
-                end = length
-            else:
-                # Try to find the end of the sentence
+
+        if length <= self.max_section_length:
```
mattgotteiner, line 46: I see - the diff is coming from the indentation after the else?

marekjakimiuk1: Hi @mattgotteiner ,

I've added if statement befor while loop. For small documents where amount of text is less than section_overlap the condition: 'while start + self.section_overlap < length:' is always false same as the last if statement in the function ('if start + self.section_overlap < end:') so no section is return. For this solution, the assumptions is that section_overlap < max_section_length - but it seems to be obvious.
I've also reported bug related to usage of section_overlap in while loop: https://github.com/Azure-Samples/azure-search-openai-demo/issues/1153, but this is separate case.
Hope, this make it more clear.

Best regards,
Marek

mattgotteiner: Thanks, this makes sense to me, I will manually test this and add some test cases. Appreciate it!

mattgotteiner (approved):

mattgotteiner:
@marekjakimiuk1  thank you for your contribution!

marekjakimiuk1:
> @marekjakimiuk1 thank you for your contribution! The PR has been edited slightly to reduce diff size and help with git blame in the future

@mattgotteiner Thank you too!

PR #1160, opened 2024-01-20
Apply acls to citations

mattgotteiner:
## Purpose

* Enforce access control when using the /content endpoint to view citations and AZURE_ENFORCE_ACCESS_CONTROL is set

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[ ] Bugfix
[X] Feature
[ ] Other... Please describe:
```

## How to Test
* Enable AZURE_ENFORCE_ACCESS_CONTROL
* Citations should not be viewable without the appropriate Authorization header obtained through login

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```

## What to Check
* When AZURE_ENFORCE_ACCESS_CONTROL is set, no files under /content are accessible without an Authorization header. The default view in the web browser should return HTTP 403

### DIFF
--- a/app/backend/app.py
+++ b/app/backend/app.py
@@ -91,15 +91,18 @@ async def content_file(path: str):
     # If authentication is enabled, validate the user can access the file
     auth_helper = current_app.config[CONFIG_AUTH_CLIENT]
     search_client = current_app.config[CONFIG_SEARCH_CLIENT]
+    authorized = False
     try:
         auth_claims = await auth_helper.get_auth_claims_if_enabled(request.headers)
-        if not await auth_helper.check_path_auth(path, auth_claims, search_client):
-            abort(403)
+        authorized = await auth_helper.check_path_auth(path, auth_claims, search_client)
     except AuthError:
         abort(403)
     except Exception as error:
         logging.exception("Problem checking path auth %s", error)
         return error_response(error, route="/content")
+
+    if not authorized:
+        abort(403)
 
     # Remove page number from path, filename-1.txt -> filename.txt
     if path.find("#page=") > 0:
--- a/app/backend/app.py
+++ b/app/backend/app.py
@@ -36,7 +36,7 @@ from approaches.chatreadretrieveread import ChatReadRetrieveReadApproach
 from approaches.chatreadretrievereadvision import ChatReadRetrieveReadVisionApproach
 from approaches.retrievethenread import RetrieveThenReadApproach
 from approaches.retrievethenreadvision import RetrieveThenReadVisionApproach
-from core.authentication import AuthenticationHelper
+from core.authentication import AuthenticationHelper, AuthError
 
 CONFIG_OPENAI_TOKEN = "openai_token"
 CONFIG_CREDENTIAL = "azure_credential"
@@ -84,10 +84,23 @@ async def assets(path):
 
 
 # Serve content files from blob storage from within the app to keep the example self-contained.
-# *** NOTE *** this assumes that the content files are public, or at least that all users of the app
-# can access all the files. This is also slow and memory hungry.
+# *** NOTE *** if AZURE_ENFORCE_ACCESS_CONTROL is not set or false, all content is accessible
+# This is also slow and memory hungry.
 @bp.route("/content/<path>")
 async def content_file(path: str):
+    # If authentication is enabled, validate the user can access the file
+    auth_helper = current_app.config[CONFIG_AUTH_CLIENT]
+    search_client = current_app.config[CONFIG_SEARCH_CLIENT]
+    try:
+        auth_claims = await auth_helper.get_auth_claims_if_enabled(request.headers)
+        if not await auth_helper.check_path_auth(path, auth_claims, search_client):
+            abort(403)
+    except AuthError:
+        abort(403)
+    except Exception as error:
+        logging.exception("Problem checking path auth %s", error)
+        return error_response(error, route="/content")
+
     # Remove page number from path, filename-1.txt -> filename.txt
     if path.find("#page=") > 0:
         path_parts = path.rsplit("#page=", 1)
--- a/app/backend/core/authentication.py
+++ b/app/backend/core/authentication.py
@@ -220,7 +220,8 @@ class AuthenticationHelper:
         # Start with the standard security filter for all queries
         security_filter = self.build_security_filters(overrides={}, auth_claims=auth_claims)
         # Filter down to only chunks that are from the specific source file
-        filter = f"{security_filter} and (sourcefile eq '{path}')"
+        filter = f"{security_filter} and (sourcepage eq '{path}')"
+        print(f"path filter {filter}")
 
         # If the filter returns any results, the user is allowed to access the document
         # Otherwise, access is denied
--- a/app/backend/core/authentication.py
+++ b/app/backend/core/authentication.py
@@ -5,6 +5,7 @@ import logging
 from typing import Any, Optional
 
 import aiohttp
+from azure.search.documents.aio import SearchClient
 from azure.search.documents.indexes.models import SearchIndex
 from msal import ConfidentialClientApplication
 from msal.token_cache import TokenCache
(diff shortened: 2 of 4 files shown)


mattgotteiner:
add tests

app/backend/app.py
```diff
@@ -84,10 +84,26 @@ async def assets(path):
 
 
 # Serve content files from blob storage from within the app to keep the example self-contained.
-# *** NOTE *** this assumes that the content files are public, or at least that all users of the app
-# can access all the files. This is also slow and memory hungry.
+# *** NOTE *** if AZURE_ENFORCE_ACCESS_CONTROL is not set or false, all content is accessible
```
pamelafox, line 87: To clarify, if they enforce authentication (but not access control), won't easyauth kick out non-logged in users with an HTTP 401 before returning this route?

app/backend/app.py
```diff
@@ -84,10 +84,26 @@ async def assets(path):
...
+# This is also slow and memory hungry.
 @bp.route("/content/<path>")
 async def content_file(path: str):
+    # If authentication is enabled, validate the user can access the file
```
pamelafox, line 91: It could be nice design wise if this was a decorator, like @access_control required, but the path-checking part is specific to this route, so am not sure if decorator would be reusable in other scenarios. Typically in Python apps, I've used decorator's to denote different kinds of auth situations, and then it's easy to attach them to various routes (auth_required, admin_required, etc). It would also make the code cleaner for developers not using authentication. What do you think, would that be a useful pattern here?

app/backend/core/authentication.py
```diff
@@ -216,3 +217,17 @@ async def get_auth_claims_if_enabled(self, headers: dict) -> dict[str, Any]:
             if self.require_access_control:
                 raise
             return {}
+
+    async def check_path_auth(self, path: str, auth_claims: dict[str, Any], search_client: SearchClient) -> bool:
+        # Start with the standard security filter for all queries
+        security_filter = self.build_security_filters(overrides={}, auth_claims=auth_claims)
+        # Filter down to only chunks that are from the specific source file
+        filter = f"{security_filter} and (sourcepage eq '{path}')"
+        print(f"path filter {filter}")
```
pamelafox, line 226: Rm print statement when done debugging

app/backend/core/authentication.py
```diff
@@ -216,3 +217,17 @@ async def get_auth_claims_if_enabled(self, headers: dict) -> dict[str, Any]:
...
+
+        # If the filter returns any results, the user is allowed to access the document
+        # Otherwise, access is denied
+        results = await search_client.search(search_text="*", top=1, filter=filter, include_total_count=True)
```
pamelafox, line 230: Doesn't include_total_count have a performance impact? (Docs say so). We only need to confirm that 1 matches, right?

app/frontend/src/components/AnalysisPanel/AnalysisPanel.tsx
```diff
@@ -23,6 +27,26 @@ export const AnalysisPanel = ({ answer, activeTab, activeCitation, citationHeigh
...
+    const fetchCitation = async () => {
+        const token = client ? await getToken(client) : undefined;
+        if (activeCitation) {
+            const response = await fetch(activeCitation, {
+                method: "GET",
+                headers: getHeaders(token)
+            });
+            const citationContent = await response.blob();
+            const citationObjectUrl = URL.createObjectURL(citationContent);
+            console.log(`object url {citationObjectUrl}`);
```
pamelafox, line 43: rm console.log when done debugging

app/frontend/src/components/AnalysisPanel/AnalysisPanel.tsx
```diff
@@ -23,6 +27,26 @@ export const AnalysisPanel = ({ answer, activeTab, activeCitation, citationHeigh
...
+
+    const fetchCitation = async () => {
+        const token = client ? await getToken(client) : undefined;
+        if (activeCitation) {
+            const response = await fetch(activeCitation, {
+                method: "GET",
+                headers: getHeaders(token)
+            });
+            const citationContent = await response.blob();
+            const citationObjectUrl = URL.createObjectURL(citationContent);
```
pamelafox, line 42: Ooh thats what you mean by object URL. TIL!

app/backend/core/authentication.py
mattgotteiner: You are right - i can change this

app/backend/app.py
mattgotteiner: Yes, I'll take a look at this pattern

app/backend/app.py
mattgotteiner: Correct, but easy-auth is auth only, it doesn't understand authorization (who can access what document). I can update the comment

app/backend/app.py
```diff
@@ -83,10 +83,57 @@ async def assets(path):
     return await send_from_directory(Path(__file__).resolve().parent / "static" / "assets", path)
 
 
+# Dectorator for routes that request a specific file that might require access control enforcement
```
pamelafox, line 86:
```suggestion
# Decorator for routes that request a specific file that might require access control enforcement
```

app/backend/app.py
```diff
@@ -83,10 +83,57 @@ async def assets(path):
...
+
+        return await route_fn(path)
+
+    # Rename decorator to avoid view function mapping overwrite error
+    auth_handler.__name__ = route_fn.__name__
+
+    return auth_handler
+
+
+# Decorator for routes that might require access control. Unpacks Authorization header informaiton into an auth_claims dictionary
```
pamelafox, line 113:
```suggestion
# Decorator for routes that might require access control. Unpacks Authorization header information into an auth_claims dictionary
```

app/backend/app.py
```diff
@@ -83,10 +83,57 @@ async def assets(path):
     return await send_from_directory(Path(__file__).resolve().parent / "static" / "assets", path)
 
 
+# Decorator for routes that request a specific file that might require access control enforcement
```
pamelafox, line 86: PEP8 nit: This comment here seems like it's basically a function docstring and should be inside the function. https://peps.python.org/pep-0257/

Not a big deal for top-level functions but a good practice generally for better intellisense/hover-ability.

app/backend/app.py
```diff
@@ -83,10 +83,57 @@ async def assets(path):
...
+    return auth_handler
+
+
 # Serve content files from blob storage from within the app to keep the example self-contained.
-# *** NOTE *** this assumes that the content files are public, or at least that all users of the app
-# can access all the files. This is also slow and memory hungry.
+# *** NOTE *** if you are using app services authentication, this route will return unauthorized to all users that are not logged in
+# if AZURE_ENFORCE_ACCESS_CONTROL is not set or false, logged in users can access all files regardless of access control
+# if AZURE_ENFORCE_ACCESS_CONTROL is set to true, logged in users can only access files they have access to
+# This is also slow and memory hungry.
```
pamelafox, line 134: Ditto for these other function comments, here and above. Another advantage of docstrings is that if someone collapses a function to declutter up their working environment, the docstring collapses with it.

app/backend/app.py
```diff
@@ -83,10 +83,57 @@ async def assets(path):
...
+        auth_helper = current_app.config[CONFIG_AUTH_CLIENT]
+        try:
+            auth_claims = await auth_helper.get_auth_claims_if_enabled(request.headers)
+        except AuthError:
+            abort(403)
+
+        return await route_fn(auth_claims)
+
+    # Rename decorator to avoid view function mapping overwrite error
+    auth_handler.__name__ = route_fn.__name__
```
pamelafox, line 125: The Flask documentation suggests that using functools.wraps() is the best practice here: https://flask.palletsprojects.com/en/3.0.x/patterns/viewdecorators/#login-required-decorator

mattgotteiner: I was wondering why this was so weird, thanks for the suggestion!

app/backend/app.py
```diff
@@ -36,7 +36,7 @@
 from approaches.chatreadretrievereadvision import ChatReadRetrieveReadVisionApproach
 from approaches.retrievethenread import RetrieveThenReadApproach
 from approaches.retrievethenreadvision import RetrieveThenReadVisionApproach
-from core.authentication import AuthenticationHelper
+from core.authentication import AuthenticationHelper, AuthError
```
pamelafox, line 39: Love the decorators! One overall comment: what do you think of moving them into their own file, route_decorators.py or some such? That would keep this file a more manageable size.

app/backend/app.py
```diff
@@ -36,24 +36,20 @@
 from approaches.chatreadretrievereadvision import ChatReadRetrieveReadVisionApproach
 from approaches.retrievethenread import RetrieveThenReadApproach
 from approaches.retrievethenreadvision import RetrieveThenReadVisionApproach
+from config import (
```
pamelafox, line 39: I think we didn't end up saving lines of code with moving them to a file, but I don't have a strong preference either way.

pamelafox (approved):
Love that you added more functionality while hiding the complexity!

PR #1166, opened 2024-01-22
Make it possible to deploy on free App Service and free Search Service tier

pamelafox:
## Purpose

This PR makes it easier to deploy to free tiers.

* App Service: I just needed to set two properties correctly, use32BitWorkerProcesses and alwaysOn, based on the tier.
* Search Service: This was the larger change that required storing the key inside the key vault and turning off managed identity. I did not add any logic to explicitly turn off semantic ranker, I just put instructions about how to disable in the docs.

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[ ] Bugfix
[X] Feature
[ ] Other... Please describe:
```

## How to Test

* I followed steps in my new doc to deploy to free tiers only, and confirmed it worked.
* I also tested both ./scripts/prepdocs.sh and ./scripts/prepdocs.ps1 in that environment.


infra/main.bicep
```diff
@@ -201,14 +205,17 @@ module backend 'core/host/appservice.bicep' = {
     serverAppId: serverAppId
     clientSecretSettingName: !empty(clientAppSecret) ? 'AZURE_CLIENT_APP_SECRET' : ''
     authenticationIssuerUri: authenticationIssuerUri
+    use32BitWorkerProcess: appServiceSkuName == 'F1'
```
pamelafox, line 208: @jongio Wondering if I should change core/appservice to always set these based on the SKU?

scripts/prepdocs.ps1
```diff
@@ -37,13 +36,15 @@ if ($env:AZURE_VISION_KEY) {
 
 # If vision keys are stored in keyvault provide the keyvault name and secret name
 if ($env:AZURE_KEY_VAULT_NAME) {
-  $visionKeyVaultName = "--visionKeyVaultName $env:AZURE_KEY_VAULT_NAME"
+  $keyVaultName = "--keyvaultname $env:AZURE_KEY_VAULT_NAME"
```
pamelafox, line 39: I renamed arg to not be vision dependent.

scripts/prepdocs.ps1
```diff
@@ -37,13 +36,15 @@ if ($env:AZURE_VISION_KEY) {
...
 }
 if ($env:VISION_SECRET_NAME) {
-  $visionKeyVaultkey = "--visionKeyVaultkey $env:VISION_SECRET_NAME"
+  $visionSecretNameArg = "--visionsecretname $env:VISION_SECRET_NAME"
```
pamelafox, line 42: I also renamed this arg to match new one, and because its not a key.

docs/deploy_lowcost.md
```diff
@@ -0,0 +1,69 @@
...
+
+    ```shell
+    azd env set AZURE_SEARCH_SERVICE_SKU free
+    ```
+
+    Limitations:
+    1. You are only allowed one free search service across all regions.
+    If you have one already, either delete that service or follow instructions to
+    reuse your [existing search service](../README.md#existing-azure-ai-search-resource).
+    2. The free tier does not support semantic ranker, so you will need to disable
```
mattgotteiner, line 25: Is it possible to check the env var in the client and disable semantic ranker if the search service tier is free?

docs/deploy_lowcost.md
```diff
@@ -0,0 +1,69 @@
...
+    but it is an option for those who want to minimize costs.
+
+4. Disable vector search:
+
+    By default, the application computes vector embeddings for documents during the data ingestion phase,
+    and then computes a vector embedding for user questions asked in the application.
+    Those computations require an embedding model, which incurs costs per tokens used. The costs are fairly low,
+    so the benefits of vector search would typically outweigh the costs, but it is possible to disable vector support.
+
+    Modify `prepdocs.sh` or `prepdocs.ps1` to add the `--novectors` argument to the command at the bottom.
```
mattgotteiner, line 60: should we add an env var to make this easier?

docs/deploy_lowcost.md
pamelafox: Yes, I can do that, in progress. Just need to make sure it'll play nicely with existing deployments.

docs/deploy_lowcost.md
pamelafox: I've added an env var for vectors now as well. I had a hard time deciding whether it should be USE_VECTORS or DONT_USE_VECTORS, but I went with USE_VECTORS for the sake of avoiding double negatives in the code. I just have to be very careful that the default is true if not defined, which I believe I've done in all cases.

docs/deploy_lowcost.md
```diff
@@ -0,0 +1,69 @@
...
+
+    ```shell
+    azd env set AZURE_FORMRECOGNIZER_SKU free
+    ```
+
+    Limitation: The free tier will only scan the first two pages of each PDF.
+    In our sample documents, those first two pages are just title pages,
+    so you won't be able to get answers from the documents.
+    You can either use your own documents that are only 2-pages long,
+    or you can use a local Python package for PDF parsing by adding the `--localpdfparser`
```
pamelafox: I went ahead and added an env var for LOCAL_PDF_PARSER too, just so that noone ever has to touch prepdocs.sh (eek).

tests/test_app.py
```diff
@@ -54,24 +54,6 @@ async def test_missing_env_vars():
                 test_app.test_client()
 
 
-@pytest.mark.asyncio
```
pamelafox, line 57: Moved to test_app_config.py

infra/main.bicep
```diff
@@ -330,36 +341,41 @@ module computerVision 'core/ai/cognitiveservices.bicep' = if (useGPT4V) {
...
     computerVisionId: useGPT4V ? computerVision.outputs.id : ''
     computerVisionSecretName: computerVisionSecretName
+    storeSearchServiceSecret: useSearchServiceKey
+    searchServiceId: useSearchServiceKey ? searchService.outputs.id : ''
+    searchServiceSecretName: searchServiceSecretName
   }
 }
 
+
+var searchServiceSemanticRankerLevel = (searchServiceSkuName == 'free') ? 'disabled' : 'free'
```
mattgotteiner, line 378: nit: Can we allow the customer to pass in a variable to have a standard tier semantic ranker?

scripts/prepdocs.ps1
```diff
@@ -37,27 +36,45 @@ if ($env:AZURE_VISION_KEY) {
...
 if ($env:USE_GPT4V -eq $true) {
   $searchImagesArg = "--searchimages"
 }
 
+if ($env:USE_VECTORS -eq $false) {
+  $disableVectorsArg="--novectors"
+}
+
+if ($env:USE_LOCAL_PDF_PARSER -eq $true) {
+  $localPdfParserArg = "--localpdfparser"
```
mattgotteiner, line 57: Good addition!

mattgotteiner (approved):
Great work!

infra/main.bicep
mattgotteiner: essentially just allowing AZURE_SEARCH_SEMANTIC_RANKER as a parameter in addition to being an output

mattgotteiner: LGTM

PR #1175, opened 2024-01-24
Update to tools parameter

tonybaloney:
`functions` is a deprecated argument. We should be using the tools parameter

### DIFF
--- a/app/backend/approaches/chatapproach.py
+++ b/app/backend/approaches/chatapproach.py
@@ -68,7 +68,7 @@ class ChatApproach(Approach, ABC):
         else:
             return override_prompt.format(follow_up_questions_prompt=follow_up_questions_prompt)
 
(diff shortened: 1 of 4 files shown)


srbalakr (approved):

tonybaloney:
`openai.BadRequestError: Error code: 400 - {'error': {'message': "Invalid value for 'function_call': 'function_call' is only allowed when 'functions' are specified.", 'type': 'invalid_request_error', 'param': 'function_call', 'code': None}}`

pamelafox:
I assume the OpenAI version we're supporting does support the new tools parameter?

tonybaloney:
> I assume the OpenAI version we're supporting does support the new tools parameter?

Yes it does

pamelafox (approved):
Thank you!

PR #1186, opened 2024-01-26
Add token validation

mattgotteiner:
## Purpose

* Validate authentication tokens before calling Microsoft.Graph on behalf of the user

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[ ] Bugfix
[X] Feature
[ ] Other... Please describe:
```

## How to Test
*  Enable authentication
* Validate all endpoints that require authentication still work with your client app registration
* Re-run auth_init.ps1/sh and confirm the authentication still works even after a token version upgrade

```
git clone [repo-address]
cd [repo-name]
git checkout [branch-name]
npm install
```


app/backend/core/authentication.py
```diff
@@ -237,3 +257,74 @@ async def check_path_auth(self, path: str, auth_claims: dict[str, Any], search_c
             break
 
         return allowed
+
+    # See https://github.com/Azure-Samples/ms-identity-python-on-behalf-of/FlaskAPI/helpers/authorization.py#L44
```
pamelafox, line 261: This is a 404

mattgotteiner: good catch!

app/backend/core/authentication.py
```diff
@@ -237,3 +257,74 @@ async def check_path_auth(self, path: str, auth_claims: dict[str, Any], search_c
...
 
         return allowed
+
+    # See https://github.com/Azure-Samples/ms-identity-python-on-behalf-of/blob/939be02b11f1604814532fdacc2c2eccd198b755/FlaskAPI/helpers/authorization.py#L44
+    async def validate_access_token(self, token: str):
+        """
+        Validate an access token is issued by Entra
+        """
+        jwks = None
+        async for attempt in AsyncRetrying(
```
pamelafox, line 267: It looks like you added the tenacity retry, is that because you found it flaky in your experience?

app/backend/core/authentication.py
```diff
@@ -237,3 +257,74 @@ async def check_path_auth(self, path: str, auth_claims: dict[str, Any], search_c
...
+                async with aiohttp.ClientSession() as session:
+                    async with session.get(url=self.key_url) as resp:
+                        resp_status = resp.status
+                        if resp_status in [500, 502, 503, 504]:
+                            raise AuthError(
+                                error=f"Failed to get keys info: {await resp.text()}", status_code=resp_status
+                            )
+                        jwks = await resp.json()
+
+        if not jwks or "keys" not in jwks:
```
pamelafox, line 282: I'm trying to parse the rest of the differences between your code and the sample code, is it mostly additional error message raising? Anything else notable?

mattgotteiner: There's 1 big change: I didn't realize we forgot to set the requested token version in the manifest to 2. Because we didn't do that, we are allowing v1 token issuer/audience in addition to v2. auth_init.py will fix this for folks running it again, but I wanted to make sure it would work if customers changed the code without rerunning provision hooks.

I also changed retries to use tenacity. Other than that they are the same

app/backend/core/authentication.py
mattgotteiner: The sample had retries using session adapters, so I used the existing retry pattern in our repo instead

pamelafox:
Merged black changes, so you can update this now

pamelafox (approved):

PR #1191, opened 2024-01-26
Rename OneShot to Ask, update Black style

pamelafox:
## Purpose

This PR changes "OneShot" to "Ask", to match its UI title. With the current naming, its harder for developers to find the right file, and its misleading since we actually use a "few-shot" technique in aspects of the "Ask" approach. A more correct term would be "Multi-turn" for "Chat" and "Single-turn" for "Ask, but I don't want to rename "Chat" as well.

The drawback of this change is that it may be an annoying merge for existing developers who have customized those files, so you can decide if it's worth it for clarity going forwards.

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test

* Start up the UI, ask a question, open up the dev settings


app/frontend/src/components/GPT4VSettings/GPT4VSettings.tsx
```diff
@@ -38,7 +38,6 @@ export const GPT4VSettings = ({ updateGPT4VInputs, updateUseGPT4V, isUseGPT4V, g
             {useGPT4V && (
                 <Dropdown
                     selectedKey={vectorFieldOption}
-                    className={styles.oneshotSettingsSeparator}
```
pamelafox, line 41: I removed these styles as they didn't exist anyway. Looks fine.

mattgotteiner (approved):

mattgotteiner:
thanks for fixing black

PR #1193, opened 2024-01-26
Add document describing the HTTP chat protocol

pamelafox:
## Purpose

This describes the HTTP protocol used by this app as well as other applications. This will be linked from some documentation pages. It was supposed to live in another non-language specific repo, but that is taking some time.

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test

* Read doc 
* Click link from main README to new doc


docs/protocol.md
```diff
@@ -0,0 +1,364 @@
+# HTTP protocol for RAG chat apps
```
mattgotteiner, line 1: I almost feel like there should be a separate repository documenting this protocol that has samples on how to integrate with it and stand up simple servers / clients. But for now I think this is OK

docs/protocol.md
```diff
@@ -0,0 +1,364 @@
...
+```json
+{
+    "messages": [
+        {
+            "content": "What is included in my Northwind Health Plus plan that is not in standard?",
+            "role": "user"
+        }
+    ],
+    "stream": false,
+    "context": {
```
mattgotteiner, line 31: Context is how servers extend the protocol? Are clients that are context-unaware guaranteed to work with the server?

docs/protocol.md
```diff
@@ -0,0 +1,364 @@
...
+An HTTP request to `/chat` or `/ask` can contain these properties, in JSON:
+
+* `"messages"`: A list of messages, each containing "content" and "role", where "role" may be "system", "assistant", or "user"
+* `"stream"`: A boolean indicating whether the response should be streamed or not.
+* `"context"`: An object containing any additional context about the request. This app only sets the "overrides" property, which is an object with additional settings for the request, like how many search results to return.
+* `"session_state"`: An object containing the "memory" for the chat app, such as a user ID. Not currently used by this app.
+
+Here's an example JSON request:
+
+```json
```
mattgotteiner, line 22: We are missing a version field in the protocol, we might want to add / remove fields in the future. How will clients for version X deal with version X+1, X-1 from the server?

docs/protocol.md
```diff
@@ -0,0 +1,364 @@
...
+{
+    "messages": [
+        {
+            "content": "What is included in my Northwind Health Plus plan that is not in standard?",
+            "role": "user"
+        }
+    ],
+    "stream": false,
+    "context": {
+        "overrides": {
```
diberry, line 32: Can these overrides either be doc'd here or link to the source of truth.

docs/protocol.md
```diff
@@ -0,0 +1,364 @@
...
+            "semantic_captions": false,
+            "suggest_followup_questions": false,
+            "use_oid_security_filter": false,
+            "use_groups_security_filter": false,
+            "vector_fields": ["embedding"],
+            "use_gpt4v": false,
+            "gpt4v_input": "textAndImages"
+        }
+    },
+    "session_state": null
```
diberry, line 45: Is there anything interesting about `session_state` that is specific to OpenAI/RAG?

docs/protocol.md
```diff
@@ -0,0 +1,364 @@
...
+with additional properties needed to display sources and citations properly.
+
+### Non-streaming response
+
+This response is based off the [OpenAI chat completion object](https://platform.openai.com/docs/api-reference/chat/object),
+with additional properties needed to display sources and citations properly.
+
+A non-streaming JSON response contains the following properties:
+
+* `"choices"`: A list of responses from the LLM, typically containing only 1 response as our app sets `n=1` when requesting a completion. Each response contains:
```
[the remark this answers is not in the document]
pamelafox: At this point, I think the OpenAI protocol is a defacto standard - notice how many open source model running servers are emulating it (like llamafile, but I found several others last weekend). It's possible that a real standard may come about that's OpenAI-independent, but this has been fairly stable. (The SDK changed, but the API stayed the same.)

docs/protocol.md
pamelafox: I think the idea was to utilize it for conversation history and user authentication, but we don't have conversation history and we use headers for user ID, so it's null for us. We do honor it though, so we pass it back and forth. There's more about it in Dan's original doc, if you can view this: https://github.com/Azure/azureml_run_specification/pull/1066/files

docs/protocol.md
```diff
@@ -0,0 +1,364 @@
...
+```json
+{
+    "choices": [
+        {
+            "delta": {
+                "role": "assistant"
+            },
+            "context": {
+                "data_points": {
+                    "text": [
```
diberry, line 264: is there a gpt4/vision response in the doc

docs/protocol.md
pamelafox: I think the version field is helpful for conveying that things changed, but also practically, we're not intending to serve older versions at same time, so I think either a client supports version X+1 or it doesn't, right?
It's good to have though, for communication purposes.

docs/protocol.md
pamelafox: Yep, context is basically the customization mechanism for both requests/responses. I try to treat context as optional where possible, to increase compatibility - i.e. in theory a non-RAG chat app client could interface with our backend and still work, it just wouldn't show citations.

docs/protocol.md
pamelafox: I'm happy to stick this in a separate repository, I have a reputation for repository creation. However, the "context" properties should be documented here, since they could vary per server/client. Let me know!

docs/protocol.md
mattgotteiner: We've extended this protocol with Authorization headers which aren't captured here (pure json request / response). Should we document this extension in the protocol spec?

docs/protocol.md
mattgotteiner: Version updates will break clients, IMO if there's no clear version field there will be issues with mismatched clients / servers

docs/protocol.md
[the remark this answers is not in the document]
mattgotteiner: OK, makes sense to me

docs/protocol.md
pamelafox: Made a new repo and made changes there, see: https://github.com/Azure-Samples/ai-chat-app-protocol/pull/1

docs/protocol.md
pamelafox: I've added a version number to top of the doc in new repo-
https://github.com/Azure-Samples/ai-chat-app-protocol/pull/1/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L1

And filed an issue with more thoughts:
https://github.com/Azure-Samples/ai-chat-app-protocol/issues/2

docs/protocol.md
pamelafox: Yes, documented here: https://github.com/Azure-Samples/ai-chat-app-protocol/pull/1/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R30
Let me know if that's described correctly.

docs/protocol.md
pamelafox: I've documented them in the new repo here: https://github.com/Azure-Samples/ai-chat-app-protocol/pull/1/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R59

They feel fairly app specific right now, so it feels strange to say they're part of a generic protocol. Oh well, we're iterating.

docs/protocol.md
pamelafox: I added an images example here:

https://github.com/Azure-Samples/ai-chat-app-protocol/pull/1/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R532

mattgotteiner (approved):
Approved but sounds like the spec is evolving. this can be a working copy, thanks!

docs/protocol.md
pamelafox: I'm going to remove the protocol document from here, since it's now in a separate repo. For this repo, my goal is to bring in Quart-schema as that can generate up-to-date documentation for anyone using the API.

pamelafox: Quart-schema issue filed here: https://github.com/Azure-Samples/azure-search-openai-demo/issues/1248

pamelafox:
@mattgotteiner Are you okay with me rm'ing the local doc, and just linking to the repo?

mattgotteiner:
> @mattgotteiner Are you okay with me rm'ing the local doc, and just linking to the repo?

Yes

PR #1214, opened 2024-01-31
Add suggestion to use [azd auth login] in the free low-cost deploy tutorial

elbruno:
## Purpose
Add suggestion to use [azd auth login] in the free low-cost deploy tutorial

[ ] Yes
[X ] No

## Pull Request Type

[ ] Bugfix
[ ] Feature
[X ] Documentation content changes
[ ] Other... Please describe:

## What to Check

### DIFF
--- a/docs/deploy_lowcost.md
+++ b/docs/deploy_lowcost.md
@@ -15,6 +15,13 @@ However, if your goal is to minimize costs while prototyping your application, f
     Enter a name that will be used for the resource group.
     This will create a new folder in the `.azure` folder, and set it as the active environment for any calls to `azd` going forward.
 
+
+    > **Important:** To use Azure Developer CLI, you need to authenticate with Azure first. You can do this by running the command `azd auth login --use-device-code`, which will prompt you to enter a device code in a browser.
+    
+    ```shell
+    azd auth login  --use-device-code
+    ```
+
 2. Use the free tier of App Service:
 
     ```shell


docs/deploy_lowcost.md
```diff
@@ -15,6 +15,13 @@ However, if your goal is to minimize costs while prototyping your application, f
     Enter a name that will be used for the resource group.
     This will create a new folder in the `.azure` folder, and set it as the active environment for any calls to `azd` going forward.
 
+
+    > **Important:** To use Azure Developer CLI, you need to authenticate with Azure first. You can do this by running the command `azd auth login --use-device-code`, which will prompt you to enter a device code in a browser.
```
jongio, line 19: You shouldn't need --use-device-code here

// @vhvb1989

vhvb1989: yeah, there's no need for the flag (unless without a specific reason)

pamelafox: It's necessary in both Codespaces and Dev Containers, which are our recommended environments, so I'm now defaulting to specifying it in my demos. Most people will need it, I think.

vhvb1989: when you run `azd auth login` on codespaces (web client / browser), azd sets `--use-device-code` automatically. 

When you are on codespaces (local client / vscode native app), vscode creates the port-forwarding automatically which makes `azd auth login` work w/o using device code.  Same thing applies here to devcontainers (WLS, for example)

You want to explicitly set `--use-device-code` flag when:
- Connection to devcontainer or codespaces w/o using VSCode local client. For example, using Putty terminal or some other SSH connection which won't provide auto port forwarding.


Important Note:  when using the `device-code` flag on Azure tenants with conditional access policies, azd gets limited to the operations it can perform and will fail on some commands like `azd pipeline config` (azd won't be able to create a service principal if the device does not meet conditional-access policies).

pamelafox: Er I don't think I'm doing anything special in my dev container, but the port forwarding isn't working. Dina experienced the same thing, she mentioned it in an issue somewhere.

Maybe the --use-device-code is getting auto-added correctly on codespaces, I've just gotten so used to needing it in dev containers that I use it everywhere. I get sad when I see another failed localhost.

pamelafox: I've filed https://github.com/Azure/azure-dev/issues/3277

I'll remove use-device-code here for consistency with the other README, but I have been showing it in my talks and videos lately since I need it for my environment.

docs/deploy_lowcost.md
```diff
@@ -15,6 +15,13 @@ However, if your goal is to minimize costs while prototyping your application, f
...
+    
+    ```shell
+    azd auth login  --use-device-code
```
pamelafox, line 22:
```suggestion
    azd auth login
```

docs/deploy_lowcost.md
```diff
@@ -15,6 +15,13 @@ However, if your goal is to minimize costs while prototyping your application, f
     Enter a name that will be used for the resource group.
     This will create a new folder in the `.azure` folder, and set it as the active environment for any calls to `azd` going forward.
 
+
+    > **Important:** To use Azure Developer CLI, you need to authenticate with Azure first. You can do this by running the command `azd auth login --use-device-code`, which will prompt you to enter a device code in a browser.
+    
+    ```shell
+    azd auth login
+    ```
```
pamelafox, line 23: @elbruno Can you make this the first step? I guess azd env new works without login but I think logging in first makes sense, flow wise, and matches main README better.

pamelafox (approved):

PR #1225, opened 2024-02-02
Improve uses of get, fix temperature bug

pamelafox:
This pull request primarily focuses on improving the handling of dictionary `get` method calls in the codebase. The changes involve replacing the `or` operator with the default parameter of the `get` method to provide cleaner and more Pythonic code. 

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test

* Automated tests

### DIFF
--- a/app/backend/approaches/approach.py
+++ b/app/backend/approaches/approach.py
@@ -96,7 +96,7 @@ class Approach:
         self.openai_host = openai_host
 
     def build_filter(self, overrides: dict[str, Any], auth_claims: dict[str, Any]) -> Optional[str]:
-        exclude_category = overrides.get("exclude_category") or None
+        exclude_category = overrides.get("exclude_category")
         security_filter = self.auth_helper.build_security_filters(overrides, auth_claims)
         filters = []
         if exclude_category:
--- a/app/backend/approaches/chatreadretrievereadvision.py
+++ b/app/backend/approaches/chatreadretrievereadvision.py
@@ -110,7 +110,7 @@ class ChatReadRetrieveReadVisionApproach(ChatApproach):
         chat_completion: ChatCompletion = await self.openai_client.chat.completions.create(
             model=self.gpt4v_deployment if self.gpt4v_deployment else self.gpt4v_model,
             messages=messages,
-            temperature=overrides.get("temperature") or 0.0,
+            temperature=0.0,
             max_tokens=100,
             n=1,
         )
@@ -194,7 +194,7 @@ class ChatReadRetrieveReadVisionApproach(ChatApproach):
         chat_coroutine = self.openai_client.chat.completions.create(
             model=self.gpt4v_deployment if self.gpt4v_deployment else self.gpt4v_model,
             messages=messages,
-            temperature=overrides.get("temperature") or 0.7,
+            temperature=overrides.get("temperature", 0.7),
             max_tokens=response_token_limit,
             n=1,
             stream=should_stream,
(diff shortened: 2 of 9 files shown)


app/backend/approaches/chatreadretrievereadvision.py
```diff
@@ -110,7 +110,7 @@ async def run_until_final_call(
         chat_completion: ChatCompletion = await self.openai_client.chat.completions.create(
             model=self.gpt4v_deployment if self.gpt4v_deployment else self.gpt4v_model,
             messages=messages,
-            temperature=overrides.get("temperature") or 0.0,
```
pamelafox, line 113: This was a bug, temperature parameter shouldn't be used in the first call.

app/backend/approaches/retrievethenread.py
```diff
@@ -118,7 +118,7 @@ async def run(
                 # Azure Open AI takes the deployment name as the model name
                 model=self.chatgpt_deployment if self.chatgpt_deployment else self.chatgpt_model,
                 messages=message_builder.messages,
-                temperature=overrides.get("temperature") or 0.3,
```
pamelafox, line 121: Interesting that our default is 0.3 for RTR, didnt realize thst.

scripts/prepdocslib/filestrategy.py
```diff
@@ -59,8 +59,8 @@ async def run(self, search_info: SearchInfo):
             async for file in files:
                 try:
                     key = file.file_extension()
-                    processor = self.file_processors[key]
```
pamelafox, line 62: This was a bug, this needs to be a get.

app/backend/approaches/approach.py
```diff
@@ -96,7 +96,7 @@ def __init__(
         self.openai_host = openai_host
 
     def build_filter(self, overrides: dict[str, Any], auth_claims: dict[str, Any]) -> Optional[str]:
-        exclude_category = overrides.get("exclude_category") or None
+        exclude_category = overrides.get("exclude_category")
```
mattgotteiner, line 99: right .get defaults to None as the returned value if it's not present. I think I helped copy paste this everywher sorry about that

app/backend/approaches/chatreadretrievereadvision.py
mattgotteiner: Is there a doc that indicates this? Would be helpful to add it in a comment for folks reading this

app/backend/approaches/retrievethenread.py
mattgotteiner: I think these choices were made a long time ago and never revisited. Perhaps we can use the evaluation repo on the samples to figure out which one is the best :)

app/backend/approaches/retrievethenread.py
```diff
@@ -118,7 +118,7 @@ async def run(
                 # Azure Open AI takes the deployment name as the model name
                 model=self.chatgpt_deployment if self.chatgpt_deployment else self.chatgpt_model,
                 messages=message_builder.messages,
```
mattgotteiner, line 120: I sometimes see issues related to "output isn't consistent". In a follow up PR would love to see us using the seed parameter and let customers put it in developer settings

https://learn.microsoft.com/en-us/azure/ai-services/openai/how-to/reproducible-output?tabs=pyton

scripts/prepdocslib/filestrategy.py
mattgotteiner: great catch!

mattgotteiner (approved):

app/backend/approaches/chatreadretrievereadvision.py
pamelafox: Sure, added comment to both chat files.

app/backend/approaches/retrievethenread.py
pamelafox: Yep good idea! I've documented for now.

PR #1238, opened 2024-02-06
Change default temp to 0.3 for final chat completion call in chat approaches

pamelafox:
## Purpose

This pull request includes changes to the `run_until_final_call` function in two files: `app/backend/approaches/chatreadretrieveread.py` and `app/backend/approaches/chatreadretrievereadvision.py`. The changes involve adjusting the temperature parameter for the OpenAI chat completion function from 0.7 to 0.3. This will affect the randomness of the AI's responses, making them more deterministic and less random.

The ask approach already uses 0.3, so this makes chat match it.

The 0.3 did well in evaluations as compared to current baseline:

Screenshot 2024-02-05 at 5 01 03 PM

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[X] Other... Please describe:
```

## How to Test

* Ran evaluation

### DIFF
--- a/app/backend/approaches/chatreadretrieveread.py
+++ b/app/backend/approaches/chatreadretrieveread.py
@@ -196,7 +196,7 @@ class ChatReadRetrieveReadApproach(ChatApproach):
             # Azure Open AI takes the deployment name as the model name
             model=self.chatgpt_deployment if self.chatgpt_deployment else self.chatgpt_model,
             messages=messages,
-            temperature=overrides.get("temperature", 0.7),
+            temperature=overrides.get("temperature", 0.3),
             max_tokens=response_token_limit,
             n=1,
             stream=should_stream,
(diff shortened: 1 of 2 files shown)


app/backend/approaches/chatreadretrieveread.py
```diff
@@ -196,7 +196,7 @@ async def run_until_final_call(
             # Azure Open AI takes the deployment name as the model name
             model=self.chatgpt_deployment if self.chatgpt_deployment else self.chatgpt_model,
             messages=messages,
-            temperature=overrides.get("temperature", 0.7),
+            temperature=overrides.get("temperature", 0.3),
```
pamelafox, line 199: I could also make a class variable on approach class, but there's no precedent for that yet, so this is more straightforward change.

pamelafox: Note that slider PR or this PR will require mods, depending which gets approved first.

mattgotteiner: I think this is OK for now

mattgotteiner (approved):

PR #1273, opened 2024-02-14
Adding support for custom Azure deployments

andredewes:
## Purpose
Currently, the app only supports talking to Azure OpenAI endpoints directly, since it manually builds the URLs with suffix '*.openai.azure.com'. This won't allow any other Azure service like API Management to be a proxy or load balancer since we can't specify a custom URL.  
This PR allows the user to set the environment variable "OPENAI_HOST" to "azure_custom" and then you can define any custom URL by setting the "OPENAI_BASE_URL" variable. 
The managed identity is kept, which allows us to keep this integration keyless.

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

## How to Test
```
- Run "azd env set OPENAI_HOST azure_custom"
- Run "azd env set OPENAI_BASE_URL {azure-custom-url}"
- Run "azd up"
```


app/backend/app.py
```diff
@@ -307,6 +309,14 @@ async def setup_clients():
         )
     elif OPENAI_HOST == "local":
         openai_client = AsyncOpenAI(base_url=os.environ["OPENAI_BASE_URL"], api_key="no-key-required")
+    elif OPENAI_HOST == "azure_custom":
+        token_provider = get_bearer_token_provider(azure_credential, "https://cognitiveservices.azure.com/.default")
+        # Store on app.config for later use inside requests
+        openai_client = AsyncAzureOpenAI(
+            api_version="2023-07-01-preview",
+            azure_endpoint=os.environ["OPENAI_BASE_URL"],
```
pamelafox, line 317: The only difference in this code is the endpoint, right? In that case, I think it'd be preferred to change the original if to "startswith("azure")" and then alter the endpoint only inside that block. Or are you thinking there'll future differentiation between the setup?

infra/main.bicep
```diff
@@ -229,6 +230,7 @@ module backend 'core/host/appservice.bicep' = {
       APPLICATIONINSIGHTS_CONNECTION_STRING: useApplicationInsights ? monitoring.outputs.applicationInsightsConnectionString : ''
       // Shared by all OpenAI deployments
       OPENAI_HOST: openAiHost
+      OPENAI_BASE_URL: openAiBaseUrl
```
pamelafox, line 233: I think this should be AZURE_OPENAI_BASE_URL or AZURE_OPENAI_CUSTOM_URL, to clarify that it's being used with an Azure client.

pamelafox (changes requested):
See comments

andredewes:
> See comments

Fixed both!

pamelafox:
Great, I will do some test deploys today of both existing environments and a new environment with this change.

pamelafox (approved):
My deploy went through successfully, so this doesnt introduce regressions in existing environments. Thanks!

PR #1282, opened 2024-02-16
Clarify that gpt-4-v cant be used with integrated vectorization

pamelafox:
## Purpose

This clarifies that you must make sure to not use integrated vectorization if you're trying to use GPT-4-vision approach. A customer ran into this today and was perplexed. We could also add an error in prepdocs to warn about it.

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```


mattgotteiner (approved):
Thanks, we can eventually make an update to support this but it will require a custom skill

PR #1308, opened 2024-02-23
Pass along status codes for errors

pamelafox:
## Purpose

I realized during loadtesting that RateLimitError is being returned as a 500. This PR adjusts the error handling code so that the status code is passed along, and they'll show as 429 instead.

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[X] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test

* Added a new unit test with mock

### DIFF
--- a/tests/test_app.py
+++ b/tests/test_app.py
@@ -6,7 +6,7 @@ from unittest import mock
 import pytest
 import quart.testing.app
 from httpx import Request, Response
-from openai import BadRequestError
+from openai import BadRequestError, RateLimitError
 
 import app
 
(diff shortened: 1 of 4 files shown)


pamelafox:
Hm. Now I'm debating if it actually should be a 500.. as other errors should probably be 500s:
https://github.com/openai/openai-python/blob/a7115b5f33acd27326e5f78e19beb0d73bd3268e/src/openai/_exceptions.py

Or I could treat 429 specially and keep treating others as 500s.

mattgotteiner (approved):

tests/test_app.py
```diff
@@ -115,6 +125,23 @@ async def test_ask_handle_exception_contentsafety(client, monkeypatch, snapshot,
     snapshot.assert_match(json.dumps(result, indent=4), "result.json")
 
 
+@pytest.mark.asyncio
+async def test_ask_handle_exception_ratelimit(client, monkeypatch, snapshot, caplog):
+    monkeypatch.setattr(
```
tonybaloney, line 130: 🙈

tonybaloney (approved):

pamelafox:
I decided not to do this, as it doesn't feel okay to pass down errors to users that are often errors with the server code, not with their own usage of the app. We should only pass down error codes that are definitely a user eror, like the content_safety_filter (and we pass that now as a 400). Closing!

Issue #1318, opened 2024-02-26
Improve OpenAI observability with openai specific instrumentation


tonybaloney:
Currently we're using OpenTelemetry and the httpx instrumentation to capture the time taken for the HTTP calls to Azure OpenAI (and OpenAI public) calls. This includes some information like the URL and host, as well as HTTP errors, but little else.

![screenshot 2024-02-26 at 10 27 04](https://github.com/Azure-Samples/azure-search-openai-demo/assets/1532417/b23e7e5c-4e2f-41db-929d-75933a5fc9d5)

I've looked at opentelemetry-instrumentation-openai which is part of the [traceloop llmetry](https://github.com/traceloop/openllmetry/) project. I think it improves the tracing of this app by capturing more properties about the chat completions and embeddings calls:

<img width="645" alt="screenshot 2024-02-26 at 11 03 29" src="https://github.com/Azure-Samples/azure-search-openai-demo/assets/1532417/bbedf02e-9a31-4983-b649-531b2d93ce69">

And for embeddings
<img width="656" alt="screenshot 2024-02-26 at 11 04 50" src="https://github.com/Azure-Samples/azure-search-openai-demo/assets/1532417/a4de0132-97d0-49c4-bd83-169374cf39f2">

The only downside I've noticed is that it doesn't capture the HTTP URL as a property of the span, so you can't see the OpenAI deployment in that specific call. Have raised a request in the downstream repo.  https://github.com/traceloop/openllmetry/issues/479

tonybaloney:
Implemented

PR #1319, opened 2024-02-26
Implement better LLM tracing with llmetetry OpenAI instrumentation

tonybaloney:
See #1318

### DIFF
--- a/app/backend/app.py
+++ b/app/backend/app.py
@@ -65,9 +65,9 @@ bp = Blueprint("routes", __name__, static_folder="static")
 mimetypes.add_type("application/javascript", ".js")
 mimetypes.add_type("text/css", ".css")
 
-httpx_transport = httpx.HTTPTransport()
+httpx_transport = httpx.AsyncHTTPTransport()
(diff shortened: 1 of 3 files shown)


app/backend/app.py
```diff
@@ -411,8 +416,10 @@ def create_app():
         configure_azure_monitor()
         # This tracks HTTP requests made by aiohttp:
         AioHttpClientInstrumentor().instrument()
-        # This tracks HTTP requests made by httpx/openai:
+        # This tracks HTTP requests made by httpx:
         HTTPXClientInstrumentor().instrument()
```
pamelafox, line 420: Is this one still tracing anything? I haven't seen traces from it recently, likely due to the OpenAI regression, and it looks like httpx is only required by openai in this codebase. I'm wondering if it can be removed.

pamelafox: Ping on this question.

pamelafox: Anthony says that httpx works sometimes, and maybe will work consistently in future, and its fine to have multiple spans, so just keep it in.

pamelafox (approved):

PR #1326, opened 2024-02-28
Improve text splitter for non-English documents

### LINKED ISSUE #1328: Text splitter incorrectly assumes that sections at ~1000 characters will be below <500 tokens in cl100k_base BPE
If you upload PDFs that contain non Indo-European languages, the average tokens per word is 2-5x higher than English, French etc. 

In those cases, the text splitter will split into sections around 1000 characters.

Instead I think we should verify the tokenised length of the section and then, if too big, split it again. This could be achieved by a recursive splitter.

Before I implement this, I'll add tests with some test documents and verify that none of the sections are beyond 500 tokens.

tonybaloney:
## Purpose

This PR is to demonstrate that the rule of "1000" characters does not fall within the limit of 500 tokens for non-english documents and to improve the splitter by adding some more CJK stop-words/punctuation. 

The splitter is then recursive if the size of the tokenized string is beyond the max amount (500). 

Included regression tests (see https://github.com/Azure-Samples/azure-search-openai-demo/pull/1326/commits/e63f9fc63a7790f3c47d59aecb37a9cae0925fa9) so the fix is trackable

Fixes #1328 

```
[ ] Yes
[x] No
```

## Pull Request Type

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```


scripts/prepdocslib/textsplitter.py
```diff
@@ -16,16 +18,92 @@ def split_pages(self, pages: List[Page]) -> Generator[SplitPage, None, None]:
             yield  # pragma: no cover - this is necessary for mypy to type check
 
 
+STANDARD_LINE_BREAKS = [",", ";", ":", " ", "(", ")", "[", "]", "{", "}", "\t", "\n"]
+CJK_LINE_BREAKS = [
+    "。",
```
tonybaloney, line 23: This aren't used by the recursive splitter, but might make the overall quality better

scripts/prepdocslib/textsplitter.py
```diff
@@ -16,16 +18,92 @@ def split_pages(self, pages: List[Page]) -> Generator[SplitPage, None, None]:
...
+    """
+    Splits page by tokens
+    """
+    tokens = cl100k.encode(text)
+    if len(tokens) < max_tokens:
+        yield SplitPage(page_num=page_num, text=text)
+    else:
+        # Split page in half and call function again
+        # TODO : Overlap first and second half
+        first_half = text[: len(text) // 2]
```
tonybaloney, line 90: FWIW we should never split + encode cl100k tokens because 1 token doesn't always produce a Unicode character, in CJK and other languages you need 2 or 3 tokens for a single character.

scripts/prepdocslib/textsplitter.py
```diff
@@ -16,16 +18,92 @@ def split_pages(self, pages: List[Page]) -> Generator[SplitPage, None, None]:
...
+def split_page_by_max_tokens(page_num: int, text: str, max_tokens: int) -> Generator[SplitPage, None, None]:
+    """
+    Splits page by tokens
+    """
+    tokens = cl100k.encode(text)
+    if len(tokens) < max_tokens:
+        yield SplitPage(page_num=page_num, text=text)
+    else:
+        # Split page in half and call function again
+        # TODO : Overlap first and second half
```
tonybaloney, line 89: I put a todo here but the overlap should be smaller than 100 characters, maybe 5%?

tonybaloney: implemented as 5%

scripts/prepdocslib/textsplitter.py
```diff
@@ -16,16 +18,92 @@ def split_pages(self, pages: List[Page]) -> Generator[SplitPage, None, None]:
...
+    "⁈",
+    "⁉",
+    "⸮",
+    "ⸯ",
+    "。",
+    "々",
+    "〻",
+    "〼",
+]
+cl100k = tiktoken.get_encoding("cl100k_base")
```
tonybaloney, line 77: This should also be configurable incase there is another BPE in the future?

pamelafox: The repo has an existing incompatibility with ollama and local language models as our token-counting code in https://github.com/Azure-Samples/azure-search-openai-demo/blob/main/app/backend/core/modelhelper.py assumes OpenAI models. I see here that you assume cl100k_base versus using encoding_for_model(). On the one hand, that means that this code will work for any model, but that also means it may work deceptively. (i.e. it thinks its tokenizing correctly, but it isn't).
On the other hand:
* the current azure openai models all use that BPE- https://github.com/openai/tiktoken/blob/1b9faf2779855124f05174adf1383e53689ed94b/tiktoken/model.py#L19
* supporting non-openai models would require bringing in the tokenizers library so we're far out from that.

I think my vote is to use tiktoken.encoding_for_model() with the name of the embedding model, since then people would realize the error with using non-openai models. I believe prepdocs already has access to the model name somewhere.

scripts/prepdocslib/textsplitter.py
```diff
@@ -16,16 +18,92 @@ def split_pages(self, pages: List[Page]) -> Generator[SplitPage, None, None]:
...
+
+
+def split_page_by_max_tokens(page_num: int, text: str, max_tokens: int) -> Generator[SplitPage, None, None]:
+    """
+    Splits page by tokens
```
pamelafox, line 82:
```suggestion
    Splits given text into SplitPage objects that are shorter than the max_tokens.
```

The current docstring is basically the function name, so I suggest making it more specific or removing it.

scripts/prepdocslib/textsplitter.py
```diff
@@ -16,16 +18,92 @@ def split_pages(self, pages: List[Page]) -> Generator[SplitPage, None, None]:
             yield  # pragma: no cover - this is necessary for mypy to type check
 
 
+STANDARD_LINE_BREAKS = [",", ";", ":", " ", "(", ")", "[", "]", "{", "}", "\t", "\n"]
+CJK_LINE_BREAKS = [
```
pamelafox, line 22: Can you link to a source for these line breaks? (Are they in spacy somewhere? Or a wikipedia article or something?)

scripts/prepdocslib/textsplitter.py
```diff
@@ -16,16 +18,92 @@ def split_pages(self, pages: List[Page]) -> Generator[SplitPage, None, None]:
...
+]
+cl100k = tiktoken.get_encoding("cl100k_base")
+
+
+def split_page_by_max_tokens(page_num: int, text: str, max_tokens: int) -> Generator[SplitPage, None, None]:
+    """
+    Splits page by tokens
+    """
+    tokens = cl100k.encode(text)
+    if len(tokens) < max_tokens:
```
pamelafox, line 85: Just checking < is intentional, versus <=
Not a big deal.

scripts/prepdocslib/textsplitter.py
```diff
@@ -16,16 +18,92 @@ def split_pages(self, pages: List[Page]) -> Generator[SplitPage, None, None]:
...
 class SentenceTextSplitter(TextSplitter):
     """
     Class that splits pages into smaller chunks. This is required because embedding models may not be able to analyze an entire page at once
     """
 
-    def __init__(self, has_image_embeddings: bool, verbose: bool = False):
-        self.sentence_endings = [".", "!", "?"]
-        self.word_breaks = [",", ";", ":", " ", "(", ")", "[", "]", "{", "}", "\t", "\n"]
+    def __init__(self, has_image_embeddings: bool, verbose: bool = False, max_tokens_per_section: int = 500):
+        self.sentence_endings = [".", "!", "?"] + ["。", "！", "？"]
```
pamelafox, line 102: I think it'd be nice and consistent to make these constants too? STANDARD_SENTENCE_ENDINGS and CJK_SENTENCE_ENDINGS
with links for CJK if possible

scripts/prepdocslib/textsplitter.py
```diff
@@ -16,16 +18,92 @@ def split_pages(self, pages: List[Page]) -> Generator[SplitPage, None, None]:
...
+    "ⸯ",
+    "。",
+    "々",
+    "〻",
+    "〼",
+]
+cl100k = tiktoken.get_encoding("cl100k_base")
+
+
+def split_page_by_max_tokens(page_num: int, text: str, max_tokens: int) -> Generator[SplitPage, None, None]:
```
pamelafox, line 80: Curious on decision to make this a function versus a method. If it was a method, it'd have access to self.max_tokens_per_section and not need to be passed in.

tests/test_prepdocslib_textsplitter.py
```diff
@@ -77,3 +78,41 @@ def test_sentencetextsplitter_split_pages():
...
+
+
+def pytest_generate_tests(metafunc):
+    """Parametrize the test_doc fixture with all the pdf files in the test-data directory."""
+    if "test_doc" in metafunc.fixturenames:
+        metafunc.parametrize("test_doc", [pdf for pdf in Path("tests", "test-data").glob("*.pdf")])
+
+
+@pytest.mark.asyncio
+async def test_sentencetextsplitter_multilang(test_doc, tmp_path):
```
pamelafox, line 90: This is great! Love seeing more tests.

tests/test_prepdocslib_textsplitter.py
```diff
@@ -77,3 +78,41 @@ def test_sentencetextsplitter_split_pages():
     assert split_pages[2].page_num == 2
     assert split_pages[2].text == 'e page"}'
     assert len(split_pages[2].text) <= max_object_length
+
+
+def pytest_generate_tests(metafunc):
+    """Parametrize the test_doc fixture with all the pdf files in the test-data directory."""
+    if "test_doc" in metafunc.fixturenames:
+        metafunc.parametrize("test_doc", [pdf for pdf in Path("tests", "test-data").glob("*.pdf")])
```
pamelafox, line 86: It might be helpful to put "_chinese" "_japanese" and "_english" in the filenames to make it clear to non-speakers which data file tests what. I can't tell from a quick glance what languages they are.

pamelafox: (Comment is here since I can't comment on filenames)

scripts/prepdocslib/textsplitter.py
```diff
@@ -16,16 +18,92 @@ def split_pages(self, pages: List[Page]) -> Generator[SplitPage, None, None]:
...
     """
     Class that splits pages into smaller chunks. This is required because embedding models may not be able to analyze an entire page at once
     """
 
-    def __init__(self, has_image_embeddings: bool, verbose: bool = False):
-        self.sentence_endings = [".", "!", "?"]
-        self.word_breaks = [",", ";", ":", " ", "(", ")", "[", "]", "{", "}", "\t", "\n"]
+    def __init__(self, has_image_embeddings: bool, verbose: bool = False, max_tokens_per_section: int = 500):
+        self.sentence_endings = [".", "!", "?"] + ["。", "！", "？"]
+        self.word_breaks = STANDARD_LINE_BREAKS + CJK_LINE_BREAKS
```
pamelafox, line 103: It confuses me that the constants are called LINE_BREAKS but the variable name is called word_breaks. Which is it? A comment may help as well.

pamelafox (changes requested):
Awesome! Most of my comments are requests for clarification.

scripts/prepdocslib/textsplitter.py
```diff
@@ -16,16 +18,92 @@ def split_pages(self, pages: List[Page]) -> Generator[SplitPage, None, None]:
...
+def split_page_by_max_tokens(page_num: int, text: str, max_tokens: int) -> Generator[SplitPage, None, None]:
+    """
+    Splits page by tokens
+    """
+    tokens = cl100k.encode(text)
+    if len(tokens) < max_tokens:
+        yield SplitPage(page_num=page_num, text=text)
+    else:
+        # Split page in half and call function again
+        # Overlap first and second halves by 5%
```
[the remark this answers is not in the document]
tonybaloney: I went based on half the amount that was demonstrated in the article https://techcommunity.microsoft.com/t5/ai-azure-ai-services-blog/azure-ai-search-outperforming-vector-search-with-hybrid/ba-p/3929167 10% might make more sense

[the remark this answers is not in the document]
tonybaloney: It now inherits the default in the text splitter class.

pamelafox:
Can you add a table test? And/or snapshot test our sample docs?

tonybaloney:
Added a test for the table section check, but looking at the code I think it could be fragile and generally needs more testing (but not in this PR).

Added a snapshot test and reverted the changes to `textsplitter.py` and ran it manually to verify:

```default
asyncio: mode=Mode.STRICT
collected 1 item

tests/test_prepdocslib_textsplitter.py .                                 [100%]

============================== 1 passed in 6.40s ===============================
Finished running tests!
```

The snapshots will need updating if we change the algorithm or the test data

pamelafox:
looks like tests are failing on ubuntu, line endings issue maybe?

tonybaloney:
Ran the tests on Linux, it's failing because of the order for the dictionary keys

pamelafox (approved):

PR #1337, opened 2024-02-28
Show users a special message for context length errors

### LINKED ISSUE #1332: Error when setting search results too high

```
- [ ] feature request
```

If you are an administrator of the app, view the full error in the logs. See aka.ms/appservice-logs for more information.
Error type: <class 'openai.BadRequestError'>`

pamelafox:
## Purpose

Fixes #1332 

This PR adds a special message for the context length exceeded error. This can happen in several situations:

1) If a user uses developer settings to request more sources than can be fit inside the context window. My opinion is that this setting should usually be disabled in production or at least capped, but perhaps many people leave it in. The user can retrieve less sources to resolve the error.
2) Some sort of incorrect setup with the indexing. For example, if the user indexed Chinese documents with the current code, they can easily exceed the context window with the default source count. The developer should improve the indexing in this case, as the user can't do much in this case.

This should help users and developers to realize more quickly what the possible issues are.

How it appears:

Screenshot of error

```
[ ] Yes
[X] No
```

```
[ ] Yes
[X] No
```

## Type of change

```
[ ] Bugfix
[X] Feature
[ ] Other... Please describe:
```

## Code quality checklist

### DIFF
--- a/app/backend/error.py
+++ b/app/backend/error.py
@@ -9,7 +9,7 @@ Error type: {error_type}
 """
 ERROR_MESSAGE_FILTER = """Your message contains content that was flagged by the OpenAI content filter."""
 
-ERROR_MESSAGE_LENGTH = """Your message exceeded the context length limit for this OpenAI model. Please shorten your message or retrieve less sources from the search index."""
+ERROR_MESSAGE_LENGTH = """Your message exceeded the context length limit for this OpenAI model. Please shorten your message or change your settings to retrieve fewer search results."""
 
 
 def error_dict(error: Exception) -> dict:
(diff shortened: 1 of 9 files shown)


app/backend/error.py
```diff
@@ -9,10 +9,14 @@
 """
 ERROR_MESSAGE_FILTER = """Your message contains content that was flagged by the OpenAI content filter."""
 
+ERROR_MESSAGE_LENGTH = """Your message exceeded the context length limit for this OpenAI model. Please shorten your message or retrieve less sources from the search index."""
```
mattgotteiner, line 12: Should we tell the user how to do "retrieve less sources from the search index"?

mattgotteiner (approved):

app/backend/error.py
pamelafox: Rewording to "Please shorten your message or change your settings to retrieve fewer search results."

I don't want to get too exact in case customer changes UI but that feels both fairly generic but also gives a good hint in terms of the settings.

PR #1346, opened 2024-02-29
Readme refactor

pamelafox:
## Purpose

This pull request includes changes to the `README.md` file to reduce the size, moving the deployment customizations into their own docs. This does mean that developers cant ctrl+f so easily, but should help new developers wading through the doc. Also made some wording changes for clarity.

```
[ ] Yes
[X] No
```

```
[ ] Yes
[X] No
```

## Type of change

```
[ ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## Code quality checklist


docs/deploy_features.md
```diff
@@ -0,0 +1,58 @@
...
+* [Enabling Integrated Vectorization](#enabling-integrated-vectorization)
+* [Enabling authentication](#enabling-authentication)
+* [Enabling login and document level access control](#enabling-login-and-document-level-access-control)
+* [Enabling CORS for an alternate frontend](#enabling-cors-for-an-alternate-frontend)
+
+## Using GPT-4
+
+We generally find that most developers are able to get high quality answers using GPT 3.5. However, if you want to try GPT-4, you can do so by following these steps:
+
+* In `infra/main.bicep`, change `chatGptModelName` to 'gpt-4' instead of 'gpt-35-turbo'.
```
charris-msft, line 18: Nice to have: it would be great to create and environment variable that the user can use to change the model instead of modifying the bicep directly

charris-msft (approved):
Looks great.
Just one suggestion for the future.

docs/deploy_features.md
pamelafox: I agree, I haven't figured out how to do that with the way we currently support OpenAI as well. Maybe if I stare hard enough I'll figure it out.

pamelafox: Err actually I think I can do it, it'll just be a nested ternary. Doable though.

PR #1348, opened 2024-02-29
Improvements to load balancer setup script

pamelafox:
## Purpose

I had the chance to try out the aca-lb last night and ran into some issues with the script. However, I think it's easier to grab the URL from the outputs, so I'm curious if we need this script. 

```
[ ] Yes
[X] No
```

## Pull Request Type

```
[ ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## How to Test

*  Run the script after filling in vars

### DIFF
--- a/scripts/load-balance-aca-setup.sh
+++ b/scripts/load-balance-aca-setup.sh
@@ -1,13 +1,20 @@
  #!/bin/sh
 
-# To set up AZD environment variabless 
+# To set up AZD environment variabless
 # for use with ACA load balancer https://github.com/Azure-Samples/openai-aca-lb
-#
-# Prereq: az login --use-device-code
-# Example usage: bash load-balance-aca-setup.sh <ACA-LB-RESOURCE-GROUP-NAME> <ACA-LB-CONTAINER-APP-URI>
 
-resourceGroupName="your-resource-group-name"
-containerAppName="your-container-app-name"
+# Fill in your resource group name and container app name:
+resourceGroupName="lb-aca-rg"
+containerAppName="lb-aca-t5bgpwpdsq2a-ca"
+
+if [ -z "$(az account show)" ]; then
+    echo "You are not logged in. Please run 'az login' or 'az login --use-device-code' first."
+    exit 1
+fi
+
+echo "Running provisioning using this subscription:"
+az account show --query "{subscriptionId:id, name:name}"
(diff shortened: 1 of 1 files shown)


diberry:
The purpose of the script is to reduce the manual steps a customer has to do to set up the chat bot to use the ACA load balancer. The FQDN is one part but the env set is the other part. Every manual step in docs is an opportunity for the customer to not be able to complete it. I would like to keep the script in the repo.

scripts/load-balance-aca-setup.sh
```diff
@@ -31,4 +39,4 @@ fi
 
 azd env set OPENAI_HOST azure_custom
 azd env set AZURE_OPENAI_CUSTOM_URL $fqdn
-
+echo "Set OPENAI_HOST to azure_custom and AZURE_OPENAI_CUSTOM_URL to $fqdn"
```
diberry, line 42: If this is a script to help customers complete steps faster, then the echo line should indicate the set env variables are done. If you need the customer to manually complete these steps for some reason, then keep this line and remove the azd env set lines.

diberry:
Since I'm working on the APIM doc now, should this script be generalized to handle either LB, or should there be a separate script for APIM @pamelafox ?

scripts/load-balance-aca-setup.sh
diberry: This is the format the ACA infra outputs. 'https://${app.properties.configuration.ingress.fqdn}'

pamelafox:
@diberry For APIM, would the arguments be the same?

I'm okay with keeping the script, I just felt like I was doing a fair amount of work just to set an env variable. (Looking up the resource group and container apps name). I also had to download azure CLI, connect it to the right subscription, etc. So my thought was that it might be just as much work / just as much that could go wrong with the aca-lb repo outputting the URI.

scripts/load-balance-aca-setup.sh
pamelafox: Ah sorry this is meant to be "set" in the past tense, like "Script has set..." but that same as imperative in English! So I should reword.  I just want it to be clear what its done.

pamelafox, line 42:
```suggestion
echo "Successfully set OPENAI_HOST to azure_custom and AZURE_OPENAI_CUSTOM_URL to $fqdn"
```

diberry (approved):

charris-msft (approved):
LGTM

PR #1351, opened 2024-02-29
feat: add local html parser

john0isaac:
## Purpose

Add [Beautiful Soup](https://beautiful-soup-4.readthedocs.io/en/latest/) to parse HTML files locally instead of using Azure services to decrease the costs of running the project.

### Deployed from my branch:
image

### Citation:
image

Live Here ==> https://app-backend-o5u5phxddyxay.azurewebsites.net/
If you test my deployment please ignore the other citation as I was downloading the html blogs in the wrong format. 😅

```
[ ] Yes
[X] No
```

```
[ ] Yes
[X] No
```

## Type of change

```
[ ] Bugfix
[X] Feature
[ ] Other... Please describe:
```

## Code quality checklist


scripts/prepdocslib/htmlparser.py
```diff
@@ -0,0 +1,41 @@
...
+    Args:
+        data: (str): The data to clean up.
+    Returns:
+        str: The cleaned up data.
+    """
+    output = re.sub(r"\n{2,}", "\n", data)
+    output = re.sub(r"[^\S\n]{2,}", " ", output)
+    output = re.sub(r"-{2,}", "--", output)
+
+    return output.strip()
```
john0isaac, line 21: This function decreases the size of the content very much.
Making the parsing faster and fewer spaces in the text.

scripts/prepdocslib/htmlparser.py
```diff
@@ -0,0 +1,41 @@
...
+        Returns:
+            Page: The parsed html Page.
+        """
+        data = content.read()
+        soup = BeautifulSoup(data, "html.parser")
+
+        # Get text only from html file
+        result = soup.get_text()
+
+        yield Page(0, 0, text=cleanup_data(result))
```
john0isaac, line 41: Don't think we need to split the HTML file into pages as getting the text only greatly decreases it's size.

scripts/prepdocs.py
```diff
@@ -70,7 +71,7 @@ async def setup_file_strategy(credential: AsyncTokenCredential, args: Any) -> St
     sentence_text_splitter = SentenceTextSplitter(has_image_embeddings=args.searchimages)
     file_processors = {
         ".pdf": FileProcessor(pdf_parser, sentence_text_splitter),
-        ".html": FileProcessor(doc_int_parser, sentence_text_splitter),
+        ".html": FileProcessor(LocalHTMLParser(), sentence_text_splitter),
```
pamelafox, line 74: I think we need to do this similar to how we do the pdf parser - add an option for using the local parser - with the default of using Document Intelligence. It's a pain adding new options since you have to modify the sh/ps1 scripts, but it's important to make it an option. If you have other ideas for switching between the two, let me know.

scripts/prepdocslib/htmlparser.py
pamelafox: Yeah, I think some use cases could warrant splitting HTML pages based on top-level headers, but in the general case, I agree with a single page, so this looks good.

scripts/prepdocslib/htmlparser.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+def cleanup_data(data: str) -> str:
+    """Cleans up the given content using regexes
+    Args:
+        data: (str): The data to clean up.
+    Returns:
+        str: The cleaned up data.
+    """
+    output = re.sub(r"\n{2,}", "\n", data)
```
pamelafox, line 17: Can you comment these, since humans are bad at grokking regexes?

john0isaac: yes, of course

pamelafox (changes requested):
Thanks so much! Can you also add tests?

scripts/prepdocs.py
john0isaac: will review and see if I don't find anything better will implement it

john0isaac:
All is done, I'm working on the tests now. Sorry, I missed it.

john0isaac:
Done what is left is the comment I left above about the docs.
The test coverage is 100%

pamelafox:
@john0isaac Yep you can add those types to scripts/requirements.in and re-generate requirements.txt from inside that folder. We have types-Pillow there already as an example.

john0isaac:
@pamelafox Done

pamelafox:
@john0isaac Thank you, it'd be great if you could update those two docs as well!

john0isaac:
Re docs working on it.

It seems wired to me that the tests are failing on black check, while all the reformatting done by black here is not related to the code I added https://github.com/Azure-Samples/azure-search-openai-demo/pull/1351/commits/aa5a9c477bde7490cec91b7de2f4084d6ea823e5 .
How was it able to pass the tests before 🤔

john0isaac:
@pamelafox done

README.md
```diff
@@ -224,6 +225,15 @@ If your existing resource is in one of those regions, then you can re-use it by
 You can also use existing Azure AI Document Intelligence and Storage Accounts. See `./infra/main.parameters.json` for list of environment variables to pass to `azd env set` to configure those existing resources.
 
 
+### Using local resources
```
john0isaac, line 228: I think this should be included to have different local options other than the parsers in the future.

pamelafox: Sure! I can add local OpenAI to it later.

pamelafox: * Err, local LLMs with OpenAI compatible endpoints.

docs/deploy_lowcost.md
```diff
@@ -38,21 +38,35 @@ However, if your goal is to minimize costs while prototyping your application, f
...
     ```shell
     azd env set AZURE_DOCUMENTINTELLIGENCE_SKU F0
     ```
 
-    Limitation: The free tier will only scan the first two pages of each PDF.
-    In our sample documents, those first two pages are just title pages,
-    so you won't be able to get answers from the documents.
-    You can either use your own documents that are only 2-pages long,
-    or you can use a local Python package for PDF parsing by setting:
+    **Limitaion for PDF files:**
```
pamelafox, line 48:
```suggestion
    **Limitation for PDF files:**
```

README.md
john0isaac: Yep, I saw you taking about this in the live streams multiple times

scripts/prepdocs.py
```diff
@@ -245,7 +254,9 @@ async def main(strategy: Strategy, credential: AsyncTokenCredential, args: Any):
     )
     parser.add_argument("files", nargs="?", help="Files to be processed")
     parser.add_argument(
-        "--datalakestorageaccount", required=False, help="Optional. Azure Data Lake Storage Gen2 Account name"
+        "--datalakestorageaccount",
```
pamelafox, line 257: Do you know what led to the change of formatting of these lines? Is your black configured differently?

pamelafox: I'm not bothered by the formatting itself, but it's nice to avoid formatting changes in unrelated PRs to reduce git noise.

john0isaac: See this:
https://github.com/Azure-Samples/azure-search-openai-demo/pull/1351#issuecomment-1972104927

tests/test_htmlparser.py
```diff
@@ -0,0 +1,28 @@
...
+                            <h1>Test header</h1>
+                            <p>Test paragraph</p>
+                        </body>
+                       </html>
+                       """
+    )
+    file.name = "test.json"
+    htmlparser = LocalHTMLParser()
+    pages = [page async for page in htmlparser.parse(file)]
+    print(pages[0].text)
```
pamelafox, line 24:
```suggestion
```

tests/test_htmlparser.py
```diff
@@ -0,0 +1,27 @@
+import io
+
+import pytest
+
+from scripts.prepdocslib.htmlparser import LocalHTMLParser
+
+
```
pamelafox, line 7: Can you add a test specifically for the cleanup function? Maybe one test for spaces, one for newlines, one for hyphens?

john0isaac: Will work on it

scripts/prepdocs.py
john0isaac: I'm working inside github codespaces.
It's something related to line length but even setting the line length the closest I got to no change is by setting the line length to 90 only one line was changed

pamelafox (approved):
I pulled the branch, tested it out with local/cloud PDF/HTML, made a few changes like printing what parser we're using. Also moved the readme section into deploy_features since we're trying to keep the main readme down. Looks good, thanks so much for the change! I'll merge once workflows complete.

john0isaac:
Thank you for reviewing the PR! Looks good to me too.

PR #1366, opened 2024-03-03
Restrict GitHub workflows run

### LINKED ISSUE #1356: Restrict GitHub workflows run according to trigger on related changes

The Python workflows runs on docs changes which I don't think is very efficient and not the best scenario.
We should restrict workflow runs to relevant changes

I can work on this btw if you want.

```
- [X] feature request
```

john0isaac:
## Purpose
Fix #1356 

- restrict frontend workflow to changes in the frontend folder
- restrict python workflow runs to changes in python files

```
[ ] Yes
[X] No
```

```
[ ] Yes
[X] No
```

## Type of change

```
[ ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## Code quality checklist

### DIFF
--- a/.github/workflows/frontend.yaml
+++ b/.github/workflows/frontend.yaml
@@ -8,7 +8,7 @@ on:
   pull_request:
     branches: [ main ]
     paths:
-      - "infra/frontend/**"
+      - "app/frontend/**"
 
 jobs:
     prettier:
--- a/.github/workflows/frontend.yaml
+++ b/.github/workflows/frontend.yaml
@@ -3,8 +3,12 @@ name: Frontend linting
 on:
   push:
     branches: [ main ]
+    paths:
+      - "app/frontend/**"
   pull_request:
     branches: [ main ]
+    paths:
+      - "infra/frontend/**"
 
 jobs:
     prettier:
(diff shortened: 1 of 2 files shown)


.github/workflows/python-test.yaml
```diff
@@ -3,8 +3,12 @@ name: Python check
 on:
   push:
     branches: [ main ]
+    paths:
+      - "**.py"
   pull_request:
     branches: [ main ]
+    paths:
+      - "**.py"
```
pamelafox, line 11: This workflow also does E2E tests, so it could be affected by frontend changes as well. Is it possible to say "run this if the change affects anything but MD files"? I think quite a few changes could affect this workflow.

.github/workflows/frontend.yaml
```diff
@@ -3,8 +3,12 @@ name: Frontend linting
 on:
   push:
     branches: [ main ]
+    paths:
+      - "app/frontend/**"
   pull_request:
     branches: [ main ]
+    paths:
+      - "app/frontend/**"
```
pamelafox, line 11: This change makes sense since we're literally cd'ing into app/frontend before running prettier, so it's a pretty well scoped workflow.

pamelafox (changes requested):
Second workflow needs to be run in more cases.

.github/workflows/python-test.yaml
john0isaac: Yep we can

pamelafox (approved):
thanks!

PR #1378, opened 2024-03-06
Add support for using new ada models with different dimensions

### LINKED ISSUE #1369: Make it easy to switch ada embedding models via azd environment variables
```
- [ ] feature request
- [x] documentation issue or request
```

param embeddingModelName string = 'text-embedding-ada-002'
to
param embeddingModelName string = 'text-embedding-3-small'

pamelafox:
## Purpose

Fixes #1369 

This PR enables developers to try out the new ada models, by letting them specify model name, version (which is "1" for the new models, vs "2" for the old), and dimensions. I've tried to write it in a way that is backward compatible, assuming 1536 dimensions, but I need to test on more old environments.

```
[ ] Yes
[X] No
```

```
[ ] Yes
[X] No
```

## Type of change

```
[ ] Bugfix
[X] Feature
[ ] Other... Please describe:
```

## Code quality checklist

- [x] I added tests that prove my fix is effective or that my feature works --- Not clear what tests can be added.


docs/deploy_features.md
```diff
@@ -19,6 +20,23 @@ We generally find that most developers are able to get high quality answers usin
 * In `infra/main.bicep`, change `chatGptModelName` to 'gpt-4' instead of 'gpt-35-turbo'.
 * You may also need to adjust the capacity above that line depending on how much TPM your account is allowed.
 
+## Using ada-3 embedding models
```
mattgotteiner, line 23: Can you please add a note that this isn't supported with int vectorization?

scripts/prepdocs.sh
```diff
@@ -58,6 +58,10 @@ if [ "$USE_VECTORS" = false ]; then
   disableVectorsArg="--novectors"
 fi
 
+if [ "$OPENAI_EMB_DIMENSIONS" = false ]; then
```
mattgotteiner, line 61: = false or present?

mattgotteiner: also, aren't you supposed to add dimensions as an actual argument here? Looks like you are passing it like a flag

scripts/prepdocslib/embeddings.py
```diff
@@ -110,7 +115,7 @@ async def create_embedding_single(self, text: str) -> List[float]:
             before_sleep=self.before_retry_sleep,
         ):
             with attempt:
-                emb_response = await client.embeddings.create(model=self.open_ai_model_name, input=text)
+                emb_response = await client.embeddings.create(model=self.open_ai_model_name, input=text, dimensions=256)
```
mattgotteiner, line 118: Hardcoded to 256 here? Should we be picking up the variable?

mattgotteiner (changes requested):
Thanks, this looks good, just a few minor comments. Appreciate it!

docs/deploy_features.md
pamelafox: I added that note under the int vect heading, but I could also note here.

scripts/prepdocs.sh
pamelafox: yes! I wonder if I changed this after testing. Will do another test run.

pamelafox: I think I must have been accidentally using 1536 in my testing, not passing on the env vars, as I uncovered more errors when fixing this. 

That does bring up an issue that @tonybaloney mentioned in a meeting earlier- in theory someone could change all the variables, *not* re-index any data, and then discover absolute nonsense search results in the app, because the app query model doesn't match the indexed model. We could add a field to the index to track the model used, but we really don't want to add fields in this repo since that isn't backwards compatible. I am okay with that being a possible situation, given that people have to read the README anyway in order to set this up, and it says re-indexing is needed. But wanted to mention it.

app/backend/app.py
```diff
@@ -222,6 +222,7 @@ async def setup_clients():
     OPENAI_HOST = os.getenv("OPENAI_HOST", "azure")
     OPENAI_CHATGPT_MODEL = os.environ["AZURE_OPENAI_CHATGPT_MODEL"]
     OPENAI_EMB_MODEL = os.getenv("AZURE_OPENAI_EMB_MODEL_NAME", "text-embedding-ada-002")
+    OPENAI_EMB_DIMENSIONS = int(os.getenv("AZURE_OPENAI_EMB_DIMENSIONS", "1536"))
```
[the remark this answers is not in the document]
tonybaloney: Default to 1536 as an integer to save the cast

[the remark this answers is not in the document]
tonybaloney: Unless pyright whinges

docs/deploy_features.md
```diff
@@ -19,6 +20,40 @@ We generally find that most developers are able to get high quality answers usin
...
+
+By default, the deployed Azure web app uses the `text-embedding-ada-002` embedding model. If you want to use one of the ada 3 embedding models, you can do so by following these steps:
```
tonybaloney, line 25: OpenAI never calls the new models "ada 3", they're always called `text-embedding-3`, I don't think ada 3 means anything (unless it's called that in the model). https://openai.com/blog/new-embedding-models-and-api-updates?ref=blog.salesforceairesearch.com

docs/deploy_features.md
```diff
@@ -19,6 +20,40 @@ We generally find that most developers are able to get high quality answers usin
...
+    azd env set AZURE_OPENAI_EMB_DIMENSIONS 256
+    ```
+
+* Set the model version to "1" (the only version as of March 2024):
+
+    ```shell
+    azd env set AZURE_OPENAI_EMB_MODEL_VERSION 1
+    ```
+
+If you have already done a deployment:
```
tonybaloney, line 49: "If you have already deployed"

docs/deploy_features.md
```diff
@@ -34,6 +69,8 @@ To enable integrated vectorization with this sample:
 3. Run `azd up` to update system and user roles
 4. You can view the resources such as the indexer and skillset in Azure Portal and monitor the status of the vectorization process.
 
+This feature is not currently compatible with GPT4-vision or the newer ada models.
```
tonybaloney, line 72: ada/text embedding

tests/test_chatapproach.py
```diff
@@ -16,6 +16,7 @@ def chat_approach():
         chatgpt_deployment="chat",
         embedding_deployment="embeddings",
         embedding_model="text-",
+        embedding_dimensions=1536,
```
tonybaloney, line 19: Can you make 1536 a constant somewhere and reuse it in these tests for when we need to update that

tonybaloney:
Assuming the following:

1. The user has already deployed
2. They don't need the documentation
3. They update to the new model
4. They use the same dimensions

What happens? Does the app just break, does it give an error, work, or just proceed to return nonsense because the similarity scores are broken?

docs/deploy_features.md
pamelafox: I think I hallucinated it.

pamelafox:
@tonybaloney I tested out that situation.
First, it took me a long time to find a valid previous environment, given that the new models are only available in a few regions. I'll note that on README.
When I asked a question in the app, I initially got a great response. I realized thats because the app defaults to hybrid search. When I changed it to vectors only, the results were unrelated, and it wasn't able to answer the question.
I'm not sure how to help developers in this scenario.

infra/main.parameters.json
```diff
@@ -77,6 +77,15 @@
...
     },
+    "embeddingModelName": {
+      "value": "${AZURE_OPENAI_EMB_MODEL_NAME=text-embedding-ada-002}"
+    },
+    "embeddingDimensions": {
+      "value": "${AZURE_OPENAI_EMB_DIMENSIONS=1536}"
+    },
+    "embeddingModelVersion": {
+      "value": "${AZURE_OPENAI_EMB_MODEL_VERSION=2}"
+    },
```
pamelafox, line 88: TODO for Pamela in this PR: Update workflows

tonybaloney:
> @tonybaloney I tested out that situation.
> 
> First, it took me a long time to find a valid previous environment, given that the new models are only available in a few regions. I'll note that on README.
> 
> When I asked a question in the app, I initially got a great response. I realized thats because the app defaults to hybrid search. When I changed it to vectors only, the results were unrelated, and it wasn't able to answer the question.
> 
> I'm not sure how to help developers in this scenario.
> 
> 

If there's nothing programmatic we can do to spot this and give an error, then we just need to be aware that it can (and probably will) happen for now. As the new model becomes more available and is adopted by more users I think this situation is going to come up more often.

Estrogenetics:
@pamelafox thanks so much for your work on this, it's be running well with 256 large. However, I recently noticed that a previous pdf file that embedded fine on ada-2 now gives me the following error when attempting to embed. Been looking for a solution, but haven't tested a working fix yet.

Batch Completed. Batch size  16 Token count 4821
Batch Completed. Batch size  16 Token count 4629
Traceback (most recent call last):
  File "/workspaces/username/./scripts/prepdocs.py", line 436, in <module>
    loop.run_until_complete(main(ingestion_strategy, azd_credential, args))
  File "/usr/local/lib/python3.11/asyncio/base_events.py", line 654, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "/workspaces/username/./scripts/prepdocs.py", line 247, in main
    await strategy.run(search_info)
  File "/workspaces/username/scripts/prepdocslib/filestrategy.py", line 76, in run
    await search_manager.update_content(sections, blob_image_embeddings)
  File "/workspaces/username/scripts/prepdocslib/searchmanager.py", line 217, in update_content
    embeddings = await self.embeddings.create_embeddings(
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspaces/username/scripts/prepdocslib/embeddings.py", line 130, in create_embeddings
    return await self.create_embedding_batch(texts)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspaces/username/scripts/prepdocslib/embeddings.py", line 97, in create_embedding_batch
    async for attempt in AsyncRetrying(
  File "/workspaces/username/scripts/.venv/lib/python3.11/site-packages/tenacity/_asyncio.py", line 71, in __anext__
    do = self.iter(retry_state=self._retry_state)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspaces/username/scripts/.venv/lib/python3.11/site-packages/tenacity/__init__.py", line 314, in iter
    return fut.result()
           ^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/concurrent/futures/_base.py", line 449, in result
    return self.__get_result()
           ^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/concurrent/futures/_base.py", line 401, in __get_result
    raise self._exception
  File "/workspaces/username/scripts/prepdocslib/embeddings.py", line 104, in create_embedding_batch
    emb_response = await client.embeddings.create(
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspaces/username/scripts/.venv/lib/python3.11/site-packages/openai/resources/embeddings.py", line 214, in create
    return await self._post(
           ^^^^^^^^^^^^^^^^^
  File "/workspaces/username/scripts/.venv/lib/python3.11/site-packages/openai/_base_client.py", line 1725, in post
    return await self.request(cast_to, opts, stream=stream, stream_cls=stream_cls)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspaces/username/scripts/.venv/lib/python3.11/site-packages/openai/_base_client.py", line 1428, in request
    return await self._request(
           ^^^^^^^^^^^^^^^^^^^^
  File "/workspaces/username/scripts/.venv/lib/python3.11/site-packages/openai/_base_client.py", line 1519, in _request
    raise self._make_status_error_from_response(err.response) from None
openai.BadRequestError: Error code: 400 - {'error': {'message': "'$.input' is invalid. Please check the API reference: https://platform.openai.com/docs/api-reference.", 'type': 'invalid_request_error', 'param': None, 'code': None}}

pamelafox:
@Estrogenetics That looks like the error we recently fixed, but I haven't merged main into this branch. I'll merge in main today.

pamelafox:
This is ready for re-review. I did two more manual tests today with standard text-embedding-ada-002 and then switching to a new model.

app/backend/app.py
```diff
@@ -222,6 +222,7 @@ async def setup_clients():
     OPENAI_HOST = os.getenv("OPENAI_HOST", "azure")
     OPENAI_CHATGPT_MODEL = os.environ["AZURE_OPENAI_CHATGPT_MODEL"]
     OPENAI_EMB_MODEL = os.getenv("AZURE_OPENAI_EMB_MODEL_NAME", "text-embedding-ada-002")
+    OPENAI_EMB_DIMENSIONS = int(os.getenv("AZURE_OPENAI_EMB_DIMENSIONS", 1536))
```
mattgotteiner, line 225: will this crash on the int() cast if the env var is a string or something?

infra/main.bicep
```diff
@@ -101,11 +101,13 @@ param embeddingModelName string = ''
 param embeddingDeploymentName string = ''
 param embeddingDeploymentVersion string = ''
 param embeddingDeploymentCapacity int = 0
+param embeddingDimensions int = 0
 var embedding = {
   modelName: !empty(embeddingModelName) ? embeddingModelName : 'text-embedding-ada-002'
   deploymentName: !empty(embeddingDeploymentName) ? embeddingDeploymentName : 'embedding'
   deploymentVersion: !empty(embeddingDeploymentVersion) ? embeddingDeploymentVersion : '2'
   deploymentCapacity: embeddingDeploymentCapacity != 0 ? embeddingDeploymentCapacity : 30
+  dimensions: embeddingDimensions != 0 ? embeddingDimensions : 1536
```
mattgotteiner, line 110: will this work on ada002 as well?

mattgotteiner (approved):
looks good - just 2 minor questions

app/backend/app.py
pamelafox: int('256') will not crash, but int('hi') will crash - which I think is desirable, right?

mattgotteiner: fair

infra/main.bicep
pamelafox: I am doing more tests and am now getting an error with the old environments. Apparently the OpenAI API won't let us specify dimensions for ada002 models at all, even if its set to 1536. So I have to optionally include in all our create() calls only for the newer models.
Not sure why I didn't catch that before. I think I keep forgetting to delete md5's when switching environments.

pamelafox: Made changes, tested in several old environments, both Azure and non-Azure.

tonybaloney:
I understood that `text-embedding-3-small` supports 512 and 1536 as dimensions. `text-embedding-3-large` supports 256, 1024, and 3072.

Therefore:
- 1536 isn't a suitable default for `text-embedding-3-large`
- There isn't any validation for those values in the code
- There isn't a mention of this in the docs

pamelafox:
I didn't realize you could only shorten to those dimensions. I see the table in https://openai.com/blog/new-embedding-models-and-api-updates?ref=upstract.com , is that where you're getting that from? I thought those were example dimension sizes but I probably mis-interpreted.

We can't do an @allows very easily since this would be a conditional allows. I could throw an error in the OpenAIEmbeddingService in prepdocs, but I assume that an error would also be thrown by OpenAI call, and it might be better to defer to them in case something changes? 

Ideally I'd link to documentation but am not seeing a clear source for allowed sizes in either https://platform.openai.com/docs/guides/embeddings/frequently-asked-questions or https://openai.com/blog/new-embedding-models-and-api-updates?ref=upstract.com

I can also ask the SDK team

pamelafox:
Hm yeah I'm able to get arbitrarily sized vectors:

vector = openai_client.embeddings.create(model="text3", input=['hello'], dimensions=400)
len(vector.data[0].embedding) # 400

pamelafox:
With an error thrown by API if dimension is too high:

BadRequestError: Error code: 400 - {'error': {'message': "Invalid value for 'dimensions' = 5000. Must be less than or equal to 3072.", 'type': 'invalid_request_error', 'param': None, 'code': None}}

Estrogenetics:
> @Estrogenetics That looks like the error we recently fixed, but I haven't merged main into this branch. I'll merge in main today.

Hi Pamela, I can confirm the merge resolved the upload issue with ada3 I was having. Upload is going smooth with no errors. I also noticed that ada3 allows documents that were scanned sideways to also correctly embed. I did not notice this ability with ada2. Thank you again, and the team, for bringing us to the forefront. Absolutely brilliant. Can't wait to dive into your courses!

tonybaloney (approved):

PR #1386, opened 2024-03-07
Allow manual setting of the OpenAI Deployment Name, Version, Capacity and Model

### LINKED ISSUE #1352: Be able to specify GPT-4 via env variable
It would be great to create and environment variable that the user can use to change the model instead of modifying the bicep directly

_Originally posted by @charris-msft in https://github.com/Azure-Samples/azure-search-openai-demo/pull/1346#discussion_r1507832085_

john0isaac:
## Purpose

fix #1352

Default params are used unless manually set using `azd env set PARAMETER`

```
[ ] Yes
[X] No
```

```
[ ] Yes
[X] No
```

## Type of change

```
[ ] Bugfix
[X] Feature
[ ] Other... Please describe:
```

## Code quality checklist


docs/deploy_features.md
```diff
@@ -16,8 +16,18 @@ You should typically enable these features before running `azd up`. Once you've
 
 We generally find that most developers are able to get high quality answers using GPT 3.5. However, if you want to try GPT-4, you can do so by following these steps:
 
-* In `infra/main.bicep`, change `chatGptModelName` to 'gpt-4' instead of 'gpt-35-turbo'.
-* You may also need to adjust the capacity above that line depending on how much TPM your account is allowed.
+* Execute the following command inside your terminal:
+
+    ```bash
+    azd env set AZURE_OPENAI_CHATGPT_DEPLOYMENT chat4
```
pamelafox, line 22: Hm, I'm not sure it's a good idea to prescribe the deployment name used by developers, as developers often want to reuse existing deployments or have some naming scheme to their deployments. Can you instead add an additional variable like USE_GPT4, so that developers can set the deployment name as desired? (The default can still be chat4 for gpt-4 deployment)

docs/deploy_features.md
```diff
@@ -16,8 +16,18 @@ You should typically enable these features before running `azd up`. Once you've
...
+    ```
+
+* You may also need to adjust the capacity in `infra/main.bicep` depending on how much TPM your account is allowed.
+
+> [!NOTE]
+> To revert back to GPT 3.5, use this command `azd env set AZURE_OPENAI_CHATGPT_DEPLOYMENT chat`, then redeploy the application using `azd provision`.
```
pamelafox, line 28:
```suggestion
> To revert back to GPT 3.5, use this command `azd env set AZURE_OPENAI_CHATGPT_DEPLOYMENT chat`, then redeploy the application using `azd up`.
```

pamelafox (changes requested):
See comments

docs/deploy_features.md
john0isaac: Nope, it's azd provision.
Testing and working 🎉
No need to redeploy the app too.

infra/main.bicep
```diff
@@ -625,4 +624,4 @@ output AZURE_STORAGE_RESOURCE_GROUP string = storageResourceGroup.name
 
 output AZURE_USE_AUTHENTICATION bool = useAuthentication
 
-output BACKEND_URI string = backend.outputs.uri
+output BACKEND_URI string = backend.outputs.uri
```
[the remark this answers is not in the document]
john0isaac: I don't think it's the new line, I ran `az bicep format` on the file that's why I think this changed.

docs/deploy_features.md
john0isaac: on it

pamelafox:
@john0isaac Ah yes, I did file an issue with the az repo about `az bicep format` and the newlines issue. I'm not sure what the current status is, there might be an argument to preserve them now. I think there was a bug when I used it a year ago.

john0isaac:
I deleted the comments as it was fixed when I added a new line like you said but I'm sure it was the format command.

I finished my quotas for the region I was testing inside, now testing in another region hope it works 🤞

docs/deploy_features.md
pamelafox: Has App Service been restarting for you consistently after environment variable updates? I've been having mixed success with that, which is why I've just been recommending "azd up" across the board.

john0isaac: No, I reviewed the vars and it was updated, I also tested the application itself and it was faster for the same prompt.

john0isaac:
It's ready for review now.

docs/deploy_features.md
```diff
@@ -16,8 +16,24 @@ You should typically enable these features before running `azd up`. Once you've
...
+1. Execute the following command inside your terminal to use GPT-4:
+
+    ```bash
+    azd env set USE_GPT4 true
+    ```
+
+2. Execute the following command inside your terminal to set the name of the GPT-4 deployment:
+
+    ```bash
+    azd env set AZURE_OPENAI_CHATGPT_DEPLOYMENT chat4
```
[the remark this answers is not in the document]
pamelafox: Ah yes thats fair, I have the same directive in my embedding-3 readme. You could say "If you've already deployed..." but it's also okay to recommend it to everyone.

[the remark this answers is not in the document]
john0isaac: I'm rewording it as I noticed a problem. 😆

pamelafox (approved):
This looks good! I'm going to test it myself soon before merging.

.azdo/pipelines/azure-dev.yml
```diff
@@ -67,6 +67,7 @@ steps:
       OPENAI_ORGANIZATION: $(OPENAI_ORGANIZATION)
       AZURE_USE_APPLICATION_INSIGHTS: $(AZURE_USE_APPLICATION_INSIGHTS)
       USE_VECTORS: $(USE_VECTORS)
+      USE_GPT4: $(USE_GPT4)
```
pamelafox, line 70: Thanks for remembering the workflows! I need to update those in my embedding-3 PR.

john0isaac: I just followed the `USE_GPT4v` changes 😅

docs/deploy_features.md
```diff
@@ -16,8 +16,27 @@ You should typically enable these features before running `azd up`. Once you've
...
+    ```
+
+You may also need to adjust the capacity in `infra/main.bicep` depending on how much TPM your account is allowed.
+
+> [!NOTE]
+> To revert back to GPT 3.5, run the following commands:
+> - `azd env set USE_GPT4 false` to disable GPT-4.
+> - `azd env set AZURE_OPENAI_CHATGPT_DEPLOYMENT <YOUR-DEPLOYMENT-NAME>` to use the name for your GPT-3 deployment.
```
john0isaac, line 36: @pamelafox you also need to ask people to set the name if they want to revert back to a previous thing as it will try to use the old name when they just disable the feature without going back to the old name or adding a new name.

pamelafox (changes requested):
I realize that 1) we have developers who want gpt-4-32K for some environments and 2) we have developers who want to specify particular versions, so I think we'll need a different approach than what I originally suggested, unfortunately, sorry!

I'm wondering if we can support them setting the deployment_name, model, and version, and just document that they need to use a different model name for OpenAI.com vs Azure OpenAI. What do you think?

john0isaac:
It's okay.. 
it's just another if statement to add inside this if so what we have right now will be the default for the else and the we check whether a variable is set before defaulting to what we have.

Don't worry about it's a small change.

john0isaac:
@pamelafox this will be very problematic IMHO
I think that if people want to change the version they can do that in the bicep files I can add some vars for it to be simpler than editing the if-else.
The whole purpose was to simplify the process when we set the version automatically.

Now if you switch to a different version and try to go back you will need to specify the other version.. which you might not know and you need to specify the specific model name. IMO, this complicates everything more if you have more than one deployment.

> I'm wondering if we can support them setting the deployment_name, model, and version, and just document that they need to use a different model name for OpenAI.com vs Azure OpenAI. What do you think?
I have a solution that might be completely different from what we are doing right now but it will fulfil what you are saying.

We check if `deployment_name, model, and version` is set and use it. if not we default to the gpt-3 model.

This means that people will have more flexibility to set the three vars but it also means that they need to save these three values somewhere to switch between different deployments. We will no longer support simple switching with USE_GPT4 or anything like that.

pamelafox (changes requested):
I don't think this way will give developers a way to say for example that they want to use "gpt-4-32k", right?

I'm thinking that we tell developers to set the model name, like

`azd env set AZURE_OPENAI_CHAT_GPT_MODEL gpt-4"

as well as the version:

`azd env set AZURE_OPENAI_CHAT_GPT_MODEL_VERSION 0613"

And also set the deployment name as currently described.

They would then need to set the model correctly based on the appropriate model name for the service. They'd be more likely to bungle something, but that would give them the flexibility they seem to be asking for.

I think we don't need the USE_GPT4 variable if we have them explicitly set those parameters.

In the case that a developer doesn't set the model name or version (empty(chatGptModelName)), then we'd set it the way it's currently set. That may require the use of a var.

Would that work? Not sure if there's a hole in the approach.

john0isaac:
> I don't think this way will give developers a way to say for example that they want to use "gpt-4-32k", right?

As I have mentioned yes, it won't allow for setting the model name.
> The current approach that I have pushed in the previous commit supports setting the deployment_name and version_name only not the model name.

> Would that work? Not sure if there's a hole in the approach.
YES, it will work

There are no holes in the approach of allowing developers to set the three vars the only considerations are:
- It will be hard to switch between different deployments. you will need to specify the exact three values as env vars for successful switching.
- If you make a typo in the `deployment_name` azd will create a new model with the new name (which I don't think people want)

If you are okay with these two points, we can proceed with the full flexibility approach.

pamelafox:
should I make a specific MD only PR to fix the links?

john0isaac:
I removed docs changes to stop it from triggering.
Yes, whenever I sync it checks out the latest repo with all markdown files problems.

I will add the docs again once I'm done.

infra/main.parameters.json
```diff
@@ -71,11 +71,29 @@
...
+    },
+    "chatGptDeploymentCapacity":{
+      "value": "${AZURE_OPENAI_CHATGPT_DEPLOYMENT_CAPACITY}"
+    },
+    "embeddingModelName":{
+      "value": "${AZURE_OPENAI_EMB_MODEL_NAME}"
     },
     "embeddingDeploymentName": {
-      "value": "${AZURE_OPENAI_EMB_DEPLOYMENT=embedding}"
+      "value": "${AZURE_OPENAI_EMB_DEPLOYMENT_NAME}"
```
pamelafox, line 90: I prefer your name, but we try to keep azd env vars backwards compatible when possible, so that devs can use their existing envs.

john0isaac: What do you mean?
I agree with you on backward compatibility but does that mean you want me to change these to the default stuff and the new names will follow the new standard to make them unique?

pamelafox: We currently use:
AZURE_OPENAI_CHATGPT_DEPLOYMENT
AZURE_OPENAI_EMB_DEPLOYMENT

So those should stay the same, and the new ones look good.

john0isaac: ok

john0isaac:
@tonybaloney if you approve please do not merge as we will need to add docs around how to do this first.

tonybaloney:
Run on existing deployment:

```default
ERROR: deployment failed: failing invoking action 'provision', error deploying infrastructure: deploying to subscription:

Deployment Error Details:
InvalidTemplateDeployment: The template deployment 'openai' is not valid according to the validation procedure. The tracking id is '98e29xxxe13'. See inner errors for details.
InvalidResourceProperties: The format and name of the model of account deployment must be specified.

TraceID: xx

ERROR: error executing step command 'provision': deployment failed: failing invoking action 'provision', error deploying infrastructure: deploying to subscription:

Deployment Error Details:
InvalidTemplateDeployment: The template deployment 'openai' is not valid according to the validation procedure. The tracking id is '9x'. See inner errors for details.
InvalidResourceProperties: The format and name of the model of account deployment must be specified.

TraceID: xxx
```

tonybaloney:
@john0isaac see last 2 commits, some of the new parameters weren't being used in the bicep and main.bicep was looking for values from the old parameter names. I quickly fixed them but please do another pass

john0isaac:
That was due to the back and force I copied the main bicep again to remove my old work and started working up there but forget to add them down in the resources sorry!
Thanks for fixing them.

tonybaloney:
Ok, so my deployment over an existing environment failed on openai bceause 

```javascript
{
  "code": "InvalidTemplateDeployment",
  "message": "The template deployment 'openai' is not valid according to the validation procedure. The tracking id is '98e29xcccc2371d7e13'. See inner errors for details.",
  "details": [
    {
      "code": "InvalidResourceProperties",
      "message": "The format and name of the model of account deployment must be specified."
    }
  ]
}
```

If I look at the deployment inputs, the following values were blank:

screenshot 2024-03-15 at 11 44 55

pamelafox:
I was worried that would be the case- that's why I often set defaults in main.parameters.json using the ENV_VAR=default syntax. If defaults dont make sense here, maybe you could do a empty check on the param and set a var? Not sure.

john0isaac:
testing that right now.
It's actually only setting the integer which is a strange behavior as it should do this for all parameters regardless of their data type.
image


The parameters that were set for you are from the previous deployment, not the current deployment.
Let me dig deeper into this.

john0isaac:
I did try with the empty first and it worked but I read in the documentation that this is how to set a default value.
https://learn.microsoft.com/en-us/azure/azure-resource-manager/bicep/parameters#:~:text=using%20parameters.-,Declaration,%3E%20%3Cparameter%2Ddata%2Dtype%3E%20%3D%20%3Cdefault%2Dvalue%3E,-A%20parameter%20can%27t

pamelafox:
azd itself is also doing some special stuff with the env vars in main.parameters.json - I'm not sure where that's documented, but the {ENV_VAR=value} in main.parameters.json is only azd-supported, I believe.

john0isaac:
if we use empty we will have to define the parameters twice as we can only use the name of a var, parameter, output only once.

john0isaac:
I will test with main.parameters and see, not sure why I preferred the main.bicep but there was a problem with the main.parameters

john0isaac:
What's weird is that now, even the chatModelName is not set.
image

image

john0isaac:
I remembered now.
If we use the `main.parameters.json` we won't be able to support `OpenAI deployment` without this manual step.

You will need to specify the model name with

```sh
azd env set AZURE_OPENAI_CHATGPT_MODEL_NAME gpt-3.5-turbo
```

As we will need to set everything in the main.parameters.json 
Which won't support checking:
image


I think we should submit an issue for azd team to look into this.

why the integer is working but not the string for the same configuration.


We have 3 different paths here:
1. Wait for azd team to work on fixing strings.
2. Use `!empty()` which means that we will define everything twice.
3. Use `main.parameters.json` and not support automatic switching to OpenAI.

Please, Let me know how you'd like to proceed.

tonybaloney:
> Use main.parameters.json and not support automatic switching to OpenAI.

MY strong preference. @pamelafox ?

pamelafox:
@john0isaac I would like to keep OpenAI support in .env's if we can find a way. I have several envs with it. Another thing we could look into is .bicepparam to see if that provides additional functionality beyond main.parameters.json. @jongio might have some ideas.

john0isaac:
I gave it a go again with empty again today.
Turns out that as long as we don't include the variables in the `main.parameters.json` it will be evaluated according to the content of the `main.bicep`

The `parameters` are empty because we are not setting them:
image

But what gets used in the deployment is the variable object which evaluates the parameters first and use the result of the evaluation.
image


When you set the value manually and do a new deployment to the old one.
It will create a new resource in the same deployment as I changed the name.
image

And now it has a DeploymentName because we set it manually.
image

And the resource is created:
image

That means that by specifying the parameters manually using azd env set it will do a new deployment with the new params.


Now, let's test on a new deployment using the old resources.

If I delete my `.azure` to deploy the resource to a new environment with the same name as the old one.
It works without having to do anything or creating extra resources.

@pamelafox, @tonybaloney this is a solution but the problem of redefining stuff still exists as we are now defining everything twice.

tonybaloney:
deployment succeeded but the app is in an error state

john0isaac:
Let me revert back any changes to app environment variables as I'm guessing that's what's messing it up.

tonybaloney:
It could be anything. I'll have a proper look

john0isaac:
I did a full deployment and it's working properly.
image


image


Don't know why your application is in an error state.
My suggestion is that you do `azd up` not `azd provision` to deploy the Python app again and pick up on the change in the code.
image

app/backend/app.py
```diff
@@ -220,7 +220,7 @@ async def setup_clients():
     AZURE_KEY_VAULT_NAME = os.getenv("AZURE_KEY_VAULT_NAME")
     # Shared by all OpenAI deployments
     OPENAI_HOST = os.getenv("OPENAI_HOST", "azure")
-    OPENAI_CHATGPT_MODEL = os.environ["AZURE_OPENAI_CHATGPT_MODEL"]
+    OPENAI_CHATGPT_MODEL = os.environ["AZURE_OPENAI_CHATGPT_MODEL_NAME"]
```
pamelafox, line 223: Please change this back to the previous env name.

john0isaac: Ok

tonybaloney:
I just had a dud-deploy. Redeployed and it worked.

john0isaac:
god

john0isaac:
I will work on the docs now.

john0isaac:
After following my docs this is the Azure OpenAI resources:
image

It successfully created the new GPT 4 model with the correct capacity.

And updated my application with the right configuration.
image

Following the instructions to revert this change and use my old GPT-3.5 deployment also works. It doesn't change Azure OpenAI models and it updates the Azure App service.
image

pamelafox:
The code changes look good now, so I am doing test deployments of:


- [ ] Previous environment with gpt 3.5
- [ ] Previous environment with gpt 4
- [ ] New environment with gpt 3.5 (aka no azd env set)
- [ ] New environment following your readme

pamelafox (approved):

pamelafox:
Looks good, thanks for tackling this! I'll merge Monday.

john0isaac:
Thanks for testing, I know it takes a big amount of time to do this! Appreciate it.

Issue #1393, opened 2024-03-08
Design for secure user document upload feature


pamelafox:
This is the proposed design for an optional feature that would allow users to upload documents, and for those documents to be queryable alongside the general documents. Users should only be able to search on their own documents.

# Architecture

We will create a new Blob container of the Azure DataLake Storage Gen2 variety, so that we can associate ACLs with the content inside it. That requires a new storage account in the Bicep, with the isHsnEnable parameter toggled, and corresponding environment variables for the account name and container name.

# Frontend

The /config route will respond with a boolean indicating whether the userUpload feature is enabled. If it's enabled *and* authentication is enabled, then the frontend will have an "Upload files" button.

The "upload files" button will upload a single file and sends a POST request with multi-part form data to the /upload route on the backend.

# App backend

The setup_clients function will set up the necessary DataLakeServiceClient per the environment variables.

The /upload route will:

* Parse the file from the payload
* Use the existing @authenticated decorator to get the auth_claims, extracting the user ID of the logged in user
* Check if a directory exists yet for that user ID using the Data Lake SDK
* If not, it will create that directory and [set the ACL](https://learn.microsoft.com/en-us/azure/storage/blobs/data-lake-storage-acl-python#set-the-acl-of-a-directory). 
* Once it exists, it will upload the file to that container, and also store the oid as a tag on the blob, for some redundancy.

# Ingestion function

If user upload feature is enabled, we will provision and deploy a new Python function with a [blob trigger](https://learn.microsoft.com/azure/azure-functions/functions-bindings-storage-blob-trigger?tabs=python-v2%2Cisolated-process%2Cnodejs-v4&pivots=programming-language-python)

The function will use the prepdocs library to ingest the target blob, per the environment variables of the function. The prepdocs library will either need to be copied as part of the prepackage step, or the function code itself will live inside scripts. Or we will reorganize the whole repo. :( We have to see what will work best and what will enable local development as well. We could even make prepdocs a pypi package, but that requires a long approval process and will make it harder for developers to quickly modify.

We have to decide whether we will put the function as a commented out service in azure.yaml, or if we use a postprovision script to deploy the function, as azd does not yet support optional deployments. We can at least provision the function in main.bicep, however.

# Why not use integrated vectorization for ingestion?

We would need a custom skill to associate the user oid with the document, which would itself be a new function (so we'd have to write a function regardless).
[...]

PR #1395, opened 2024-03-08
Optional upload documents feature

pamelafox:
## Purpose

This PR contains the changes required for the upload documents feature described in #1393 

High-level overview of changes:

* Moved prepdocs into app/backend
* Made a single venv in the root of the repo for both scripts and app/backend
* Added routes for /upload /delete_upload and /list_uploads
* Added logic to /content route for looking for docs in the user's ADLS2 folder
* Added frontend for upload UI

I notably did not end up making a function app as originally suggested. I realized that would significantly increase the architectural complexity of the app, given the additional function app deployment and the necessary for polling to let the user know when the document was ready to query. That does mean that if a document takes longer than the App Service timeout to process, then it will fail. If we find out that customer's users are facing that issue, and that large document handling is needed, we can discuss ways to break up the ingestion into tasks, possibly with a function or with a Celery queue backed by Redis.

```
[ ] Yes
[X] No
```

```
[ ] Yes
[X] No - optional feature, auth only
```

## Type of change

```
[ ] Bugfix
[X] Feature
[ ] Other... Please describe:
```

## Code quality checklist

- [ ] I added tests that prove my fix is effective or that my feature works: **Missing a test for upload, am discussing with Quart team how to do that.**
- [ ] I ran `python -m pytest --cov` to verify 100% coverage of added lines: **Missing coverage for upload**

None of that yet!! :( WIP.


app/frontend/src/components/UploadFile/UploadFile.tsx
```diff
@@ -0,0 +1,144 @@
...
+import { IIconProps, Callout, ActionButton, PrimaryButton, Label, IconButton, Text } from "@fluentui/react";
+import { FileUploadResponse, uploadFileApi } from "../../api";
+import { useMsal } from "@azure/msal-react";
+import { useLogin, getToken } from "../../authConfig";
+
+interface Props {
+    className?: string;
+}
+
+export const UploadFile: React.FC<Props> = ({ className }: Props) => {
```
pamelafox, line 12: This component is from the existing PR in this repo, so I need to give it another pass for style consistency and make sure it doesn't have any unneeded functionality.

javapro13: @pamelafox thanks for working on this. Do you have an ETA for the file upload to work?

pamelafox:
Update: Upload works locally, by adding prepdocslib to path. I am reconsidering whether we should use a function versus just doing it in app service (and moving prepdocslib to app/backend). Will write more tomorrow.

javapro13:
@pamelafox  and @zedhaque  - thanks a LOT for this feature. I am a newbie how do I install the version that allows file upload? I was able to deploy the original version but how do we get this functionality?

javapro13:
@zedhaque  got it thank you

pamelafox:
@zedhaque Thanks for the comments! The PR now implements deletion of files from both the storage and the index. We're currently using a single index since that is more performant on the RAG side. I certainly see advantages of storing in a different index, but given the complexity and performance downsides, we'll stick with a single index for now. I've added aclshash to the ID of the chunks in the index, to avoid naming collision issues.

Re file size: that might be something easy to enforce in either the JS or Python? I haven't looked into it, but we could put a quick check in the code if you know an easy way to check that.

Re limits per user: that would be easier to do if we had a User object in a database. I could see that happening once we have a conversation feature since that'd require a User object, but not in scope for this v1 of the feature.

app/frontend/src/components/UploadFile/UploadFile.tsx
pamelafox: This PR works now, so you can merge it in if you need the feature ASAP, but it has not yet been reviewed/approved, so it might change by the time it's merged into main. I'll be sending it for review today after I've written tests.

app/backend/app.py
```diff
@@ -205,15 +229,69 @@ def config():
...
+
+    user_oid = auth_claims["oid"]
+    file = request_files.getlist("file")[0]
+    user_blob_container_client: FileSystemClient = current_app.config[CONFIG_USER_BLOB_CONTAINER_CLIENT]
+    user_directory_client = user_blob_container_client.get_directory_client(user_oid)
+    await user_directory_client.set_access_control(owner=user_oid)
+    file_client = user_directory_client.get_file_client(file.filename)
+    file_io = file
+    file_io.name = file.filename
+    file_io = io.BufferedReader(file_io)
```
pamelafox, line 253: There might be a simpler way to get the appropriate file type to pass into our ingester, feel free to suggest if you know one.

app/backend/app.py
```diff
@@ -205,15 +229,69 @@ def config():
...
+    await ingester.remove_file(filename, user_oid)
+    return jsonify({"message": f"File {filename} deleted successfully"}), 200
+
+
+@bp.get("/list_uploaded")
+@authenticated
+async def list_uploaded(auth_claims: dict[str, Any]):
+    user_oid = auth_claims["oid"]
+    user_blob_container_client: FileSystemClient = current_app.config[CONFIG_USER_BLOB_CONTAINER_CLIENT]
+    all_paths = user_blob_container_client.get_paths(path=user_oid)
```
pamelafox, line 281: Annoying, you can't just get directory_client and call get_paths on it, this seems to be the only way.

app/backend/app.py
```diff
@@ -276,19 +355,24 @@ async def setup_clients():
...
 
     # Set up authentication helper
+    search_index = None
+    if AZURE_USE_AUTHENTICATION:
+        search_index_client = SearchIndexClient(
+            endpoint=f"https://{AZURE_SEARCH_SERVICE}.search.windows.net",
+            credential=search_credential,
+        )
+        search_index = await search_index_client.get_index(AZURE_SEARCH_INDEX)
+        await search_index_client.close()
```
pamelafox, line 373: This was never closed before :(

app/backend/app.py
```diff
@@ -351,7 +473,23 @@ async def setup_clients():
         query_speller=AZURE_SEARCH_QUERY_SPELLER,
     )
 
+    current_app.config[CONFIG_CHAT_APPROACH] = ChatReadRetrieveReadApproach(
```
pamelafox, line 476: Just moved this around as it was weirdly placed before. (It was Ask, then GPT4V, then Chat. Now its Chat,Ask,GPT4v)

app/frontend/src/pages/ask/Ask.tsx
```diff
@@ -49,16 +51,15 @@ export function Component(): JSX.Element {
     const client = useLogin ? useMsal().instance : undefined;
 
     const getConfig = async () => {
-        const token = client ? await getToken(client) : undefined;
-
-        configApi(token).then(config => {
+        configApi().then(config => {
```
pamelafox, line 54: The config API does not require a user token, so I removed it. If we did merge auth setup and config API, then it would require a token at that point.

app/start.ps1
```diff
@@ -16,23 +16,23 @@ if ($LASTEXITCODE -ne 0) {
 }
 
 
-Write-Host 'Creating python virtual environment "backend/backend_env"'
+Write-Host 'Creating python virtual environment ".venv"'
```
pamelafox, line 19: One venv to rule them all!

infra/core/storage/storage-account.bicep
```diff
@@ -16,6 +16,7 @@ param defaultToOAuthAuthentication bool = false
 param deleteRetentionPolicy object = {}
 @allowed([ 'AzureDnsZone', 'Standard' ])
 param dnsEndpointType string = 'Standard'
+param isHnsEnabled bool = false
```
pamelafox, line 19: I will add this to azure-dev Bicep

pamelafox:
Added the missing test, all passing now.

pamelafox:
@zedhaque Ah, I think you need:

```
// For computer vision access by the backend
module cognitiveServicesRoleBackend 'core/security/role.bicep' = if (useGPT4V) {
  scope: resourceGroup
  name: 'cognitiveservices-role-backend'
  params: {
    principalId: backend.outputs.identityPrincipalId
    roleDefinitionId: 'a97b65f3-24c7-4388-baec-2e87135dc908'
    principalType: 'ServicePrincipal'
  }
}
```

I'll add it. Not sure about the first error, is that the entire traceback?

app/backend/app.py
```diff
@@ -106,8 +118,20 @@ async def content_file(path: str):
...
     except ResourceNotFoundError:
-        logging.exception("Path not found: %s", path)
-        abort(404)
+        logging.info("Path not found in general Blob container: %s", path)
+        if current_app.config[CONFIG_USER_UPLOAD_ENABLED]:
+            try:
+                user_oid = auth_claims["oid"]
+                user_blob_container_client = current_app.config[CONFIG_USER_BLOB_CONTAINER_CLIENT]
+                user_directory_client: FileSystemClient = user_blob_container_client.get_directory_client(user_oid)
+                # Should we do an additional check on access control here?
```
mattgotteiner, line 127: Is this an ADLS Gen2 container? If so we can add the actual ACL to the directory? It doesn't get automatically added to child files though unfortunately.

https://learn.microsoft.com/en-us/azure/storage/blobs/data-lake-storage-acl-python

pamelafox: We set the acl upon uploading in the upload route, later:

await user_directory_client.set_access_control(owner=user_oid)

Are you saying it should also be done here, in the content route?

mattgotteiner: Got it - I just wanted to make sure that the ACL was set on the uploaded file.

mattgotteiner: OK now I understand - my comment is in the wrong place

app/backend/app.py
mattgotteiner: This seems reasonable - 1 upload to the container, then another pass through the ingester.

app/backend/app.py
mattgotteiner: correct, the reason why is that this call is paging potentially

pamelafox:
@zedhaque I think I've replicated in production, I'll work on it. Thanks so much for testing.

scripts/prepdocs.ps1
```diff
@@ -1,9 +1,9 @@
...
 
-$venvPythonPath = "./scripts/.venv/scripts/python.exe"
+$venvPythonPath = "./.venv/scripts/python.exe"
 if (Test-Path -Path "/usr") {
   # fallback to Linux venv path
-  $venvPythonPath = "./scripts/.venv/bin/python"
+  $venvPythonPath = "./.venv/bin/python"
 }
 
 Write-Host 'Running "prepdocs.py"'
```
jaspb, line 9: Missing path update in line 69 from `./scripts/prepdocs.py` to `./app/backend/prepdocs.py`

app/backend/app.py
mattgotteiner: theoretically we can use async with as well here

app/backend/prepdocslib/searchmanager.py
```diff
@@ -226,13 +226,13 @@ async def remove_content(self, path: Optional[str] = None, only_oid: Optional[st
...
                 filter = None if path is None else f"sourcefile eq '{os.path.basename(path)}'"
-                result = await search_client.search("", filter=filter, top=1000, include_total_count=True)
+                result = await search_client.search(search_text="", filter=filter, top=1000, include_total_count=True)
                 if await result.get_count() == 0:
                     break
                 documents_to_remove = []
                 async for document in result:
                     # If only_oid is set, only remove documents that have only this oid
-                    if not only_oid or document["oids"] == [only_oid]:
+                    if not only_oid or document.get("oids") == [only_oid]:
```
mattgotteiner, line 235: good catch here

app/frontend/src/api/api.ts
```diff
@@ -32,23 +37,58 @@ export async function askApi(request: ChatAppRequest, idToken: string | undefine
...
 }
 
 export function getCitationFilePath(citation: string): string {
     return `${BACKEND_URI}/content/${citation}`;
 }
+
+export async function uploadFileApi(request: FormData, idToken: string): Promise<SimpleAPIResponse> {
+    const response = await fetch("/upload", {
+        method: "POST",
+        headers: getHeaders(idToken),
```
mattgotteiner, line 55: needed to add "Content-Type: application/json" here?

app/frontend/src/api/api.ts
```diff
@@ -32,23 +37,58 @@ export async function askApi(request: ChatAppRequest, idToken: string | undefine
...
+
+    const dataResponse: SimpleAPIResponse = await response.json();
+    return dataResponse;
+}
+
+export async function deleteUploadedFileApi(filename: string, idToken: string): Promise<SimpleAPIResponse> {
+    const response = await fetch("/delete_uploaded", {
+        method: "POST",
+        headers: { ...getHeaders(idToken), "Content-Type": "application/json" },
+        body: JSON.stringify({ filename })
```
mattgotteiner, line 71: I'm actually not familiar with this syntax. does it expand to a object like this?
{ "filename": filename }

app/frontend/src/components/UploadFile/UploadFile.tsx
```diff
@@ -0,0 +1,146 @@
...
+    // State variables to manage the component behavior
+    const [isCalloutVisible, setIsCalloutVisible] = useState<boolean>(false);
+    const [isUploading, setIsUploading] = useState<boolean>(false);
+    const [isLoading, setIsLoading] = useState<boolean>(true);
+    const [uploadedFile, setUploadedFile] = useState<SimpleAPIResponse>();
+    const [uploadedFiles, setUploadedFiles] = useState<string[]>([]);
+
+    if (!useLogin) {
+        console.error("This component requires useLogin to be true");
```
mattgotteiner, line 21: should we raise an exception here? We would also have a conditional layout that never tries to render this component at all before we would hit this conditional correct?

app/frontend/src/components/UploadFile/UploadFile.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+    const handleRemoveFile = async (filename: string) => {
+        try {
+            const idToken = await getToken(client);
+            if (!idToken) {
+                throw new Error("No authentication token available");
+            }
+            await deleteUploadedFileApi(filename, idToken);
+            const newFiles = uploadedFiles.filter(file => file !== filename);
+            // To discuss: We could also poll the list endpoint here
```
mattgotteiner, line 53: IMO it's OK to immediately show the file deleted. We can file a follow up issue if we don't like this behavior later on?

app/frontend/src/components/UploadFile/UploadFile.tsx
```diff
@@ -0,0 +1,146 @@
...
+    };
+
+    // Handler for the form submission (file upload)
+    const handleUploadFile = async (e: ChangeEvent<HTMLInputElement>) => {
+        e.preventDefault();
+        if (!e.target.files || e.target.files.length === 0) {
+            return;
+        }
+        setIsUploading(true); // Start the loading state
+        const file: File = e.target.files[0];
```
mattgotteiner, line 67: What happens if the user drags multiple files in? Would we want to upload all of them?

mattgotteiner: or does the control only allow uploading a single file at a time?

app/frontend/src/pages/ask/Ask.tsx
mattgotteiner: if it's simpler to keep the separate, that's great. I was not under the impression that we had any sensitive info in the auth setup endpoint, so it's also possible I just added a auth token for no reason :)

infra/main.bicep
```diff
@@ -436,6 +450,33 @@ module storage 'core/storage/storage-account.bicep' = {
...
+    name: !empty(userStorageAccountName) ? userStorageAccountName : 'user${abbrs.storageStorageAccounts}${resourceToken}'
+    location: storageResourceGroupLocation
+    tags: tags
+    allowBlobPublicAccess: false
+    publicNetworkAccess: 'Enabled'
+    isHnsEnabled: true
+    sku: {
+      name: storageSkuName
+    }
+    deleteRetentionPolicy: {
```
mattgotteiner, line 467: is this an arbitrary choice? Or should we just remove the delete retention policy for now

infra/main.bicep
```diff
@@ -576,8 +638,8 @@ module searchReaderRoleBackend 'core/security/role.bicep' = if (useAuthenticatio
   }
 }
 
-// For computer vision access by the backend
-module cognitiveServicesRoleBackend 'core/security/role.bicep' = if (useGPT4V) {
+// For computer vision or document intelligence access by the backend
+module cognitiveServicesRoleBackend 'core/security/role.bicep' = if (useGPT4V || useUserUpload) {
```
mattgotteiner, line 642: Would they need this still if they used the local pdf parser?

scripts/auth_init.py
```diff
@@ -142,7 +142,7 @@ def create_client_app_payload(server_app_id: str, server_app_permission_setup_pa
             "redirectUris": ["http://localhost:50505/.auth/login/aad/callback"],
             "implicitGrantSettings": {"enableIdTokenIssuance": True},
         },
-        "spa": {"redirectUris": ["http://localhost:50505/redirect"]},
+        "spa": {"redirectUris": ["http://localhost:50505/redirect", "http://localhost:5173/redirect"]},
```
mattgotteiner, line 145: is 5173 a port used by codespaces, or was it a local choice?

pamelafox: Ah that's what vite dev uses! I often use "npm run dev" when working on the frontend in order to get hot reloading.

infra/main.bicep
pamelafox: No, but currently we *always* create the Document Intelligence resource, even when using local parsers. It doesn't cost money to create it, just to use it. So our Bicep doesnt use the USE_LOCAL* to decide anything about provisioning.

infra/main.bicep
pamelafox: Arbitrary, can delete.

app/frontend/src/components/UploadFile/UploadFile.tsx
pamelafox: I removed "multiple" from the input tag so that it'll only allow one.

app/frontend/src/api/api.ts
pamelafox: Yep, see here:
https://leanpub.com/understandinges6/read/#leanpub-auto-object-literal-syntax-extensions

app/frontend/src/api/api.ts
pamelafox: No, this is actually why I needed to refactor, as upload does not send JSON. The response is JSON but the content-type is about the request.

scripts/prepdocs.ps1
mattgotteiner: i hit this issue as well

pamelafox: Thanks, fixed!

app/frontend/src/components/UploadFile/UploadFile.tsx
pamelafox: I've decided it's easier to do the Right Thing than deal with possible confusion from an over-optimistic UI. Now I've got it so that it shows a status for deletion next to each file like "Deleting..." "File deleted!" and then finally the file goes away when the new list comes back. Also shows an error status if there's an issue deleting the file.

pamelafox:
@zedhaque There was a missing role for adding to the search index. I'm now able to upload and query data. However, I had an error when clicking the citation, in the /content route, so I am debugging that next.

pamelafox:
This is now working when deployed. Please re-test @zedhaque @mattgotteiner

superpoussin22:
@pamelafox  just got this error: Unexpected Application Error!
The UploadFile component requires useLogin to be true
do you have any idea ?
thanks in advance

pamelafox:
@superpoussin22 Does the UI display your user account as being logged in at that time? Are there any other errors in the console?

superpoussin22:
> @superpoussin22 Does the UI display your user account as being logged in at that time? Are there any other errors in the console?

Hi Pamela ,
if I take a look at the myfqdn/.auth/me ; I see that I have an access  token 

on the interface I have
Error: The UploadFile component requires useLogin to be true
    at Zn (https://datademocracy.azurewebsites.net/assets/index-c5c7b5cf.js:2:12822)
    at dv (https://datademocracy.azurewebsites.net/assets/vendor-f70a64b2.js:38:19518)
    at BL (https://datademocracy.azurewebsites.net/assets/vendor-f70a64b2.js:40:44014)
    at xL (https://datademocracy.azurewebsites.net/assets/vendor-f70a64b2.js:40:39750)
    at GB (https://datademocracy.azurewebsites.net/assets/vendor-f70a64b2.js:40:39678)
    at jd (https://datademocracy.azurewebsites.net/assets/vendor-f70a64b2.js:40:39532)
    at vS (https://datademocracy.azurewebsites.net/assets/vendor-f70a64b2.js:40:35900)
    at LL (https://datademocracy.azurewebsites.net/assets/vendor-f70a64b2.js:40:34851)
    at C (https://datademocracy.azurewebsites.net/assets/vendor-f70a64b2.js:25:1550)
    at MessagePort.L (https://datademocracy.azurewebsites.net/assets/vendor-f70a64b2.js:25:1910)

pamelafox:
@superpoussin22 Do you see your username in top right? Here's what I see:

Screenshot 2024-03-20 at 1 41 54 PM

I'm wondering if your environment variable has USE_AUTHENTICATION false?

superpoussin22:
> @superpoussin22 Do you see your username in top right? Here's what I see:
> 
> Screenshot 2024-03-20 at 1 41 54 PM
> 
> I'm wondering if your environment variable has USE_AUTHENTICATION false?

I dont have enough time to see the page before it display the error
but yes  it was AZURE_USE_AUTHENTICATION set to false, I change it to true and now it works, I have access to the interface so now I can start testing the upload  :)
very nice work @pamelafox

superpoussin22:
@pamelafox  for a user to be able to upload a file , I first need to create the directory in the blob storage , if not created I go an error 500 saying the path does not exist. I will take a look in the code to see if it can be automated , but perhaps you will be quicker than me

app/backend/app.py
```diff
@@ -102,12 +114,23 @@ async def content_file(path: str):
...
         blob = await blob_container_client.get_blob_client(path).download_blob()
     except ResourceNotFoundError:
-        logging.exception("Path not found: %s", path)
-        abort(404)
+        logging.info("Path not found in general Blob container: %s", path)
+        if current_app.config[CONFIG_USER_UPLOAD_ENABLED]:
+            try:
+                user_oid = auth_claims["oid"]
+                user_blob_container_client = current_app.config[CONFIG_USER_BLOB_CONTAINER_CLIENT]
+                user_directory_client: FileSystemClient = user_blob_container_client.get_directory_client(user_oid)
```
superpoussin22, line 126: try:
                    await user_directory_client.get_directory_properties()
                except ResourceNotFoundError:
                    # The directory does not exist, so create it
                    logging.info(f"Creating directory for user {user_oid}")
                    await user_directory_client.create_directory()

pamelafox: Huh, the reference for get_directory_client says that it need not exist:
https://learn.microsoft.com/en-us/python/api/azure-storage-file-datalake/azure.storage.filedatalake.filesystemclient?view=azure-python#azure-storage-filedatalake-filesystemclient-get-directory-client

But perhaps it's wrong or I misunderstood it. I'll delete my directory and test again, looks like I'll need that code. Thanks! @zedhaque I think you ran into this as well.

superpoussin22: %s not %S :)

pamelafox: Thank you! I brought in that code, added a test, and filed an issue on the SDK re the confusing documentation: https://github.com/Azure/azure-sdk-for-python/issues/34939

infra/main.bicep
pamelafox: Resolving this, please re-open if this is a concern.

app/frontend/src/api/api.ts
pamelafox: Resolving this, please re-open if there's still a question here.

app/frontend/src/pages/ask/Ask.tsx
pamelafox: I looked into merging them, but they are currently called at two different times, since config is used for the inner page and authSetup is used for the navigation header (different React components). They could potentially still be merged but it'd require some re-organization of the code. This seems to work alright for now.

pamelafox:
@zedhaque I've resolved the get_directory_client issue. As for AZURE_TENANT_ID and AZURE_AUTH_TENANT_ID:
* AZURE_TENANT_ID should be set on your web app in main.bicep
* AZURE_AUTH_TENANT_ID is also set in main.bicep, but you may not have set that. That's only needed for situations where someone provisions the app in one tenant but sets up auth in another (for permission reasons). We default to AZURE_TENANT_ID if not set.

pamelafox:
I've addressed the issues around directory_client not existing, for both the list_uploaded and upload endpoints, and added tests. Please re-review and let me know if you see any other issues.

pamelafox:
@zedhaque 

Hm, so the Bicep should have created this role for the app service app:

// For computer vision or document intelligence access by the backend
module cognitiveServicesRoleBackend 'core/security/role.bicep' = if (useGPT4V || useUserUpload) {
  scope: resourceGroup
  name: 'cognitiveservices-role-backend'
  params: {
    principalId: backend.outputs.identityPrincipalId
    roleDefinitionId: 'a97b65f3-24c7-4388-baec-2e87135dc908'
    principalType: 'ServicePrincipal'
  }
}

Can you check the roles of the app service app and confirm it has that role?

DSRC-ACS:
Hi,

I'm trying to upload a file from the UI, I've logged in successfully. But I'm facing the following error. Can you please help me regarding this.

image

DSRC-ACS:
Hi,

I'm trying to run the application locally. I'm successfully able to login but while uploading the document I'm getting the following error in the image. Can you please help regarding this.

image

Thankyou

pamelafox:
@zedhaque Your issue should now be resolved, with my recent change to make sure the RBAC roles are scoped to the current resource group.

@DSRC-ACS Did you set USE_USER_UPLOAD true? This is the code that should have constructed the container: https://github.com/Azure-Samples/azure-search-openai-demo/pull/1395/files#diff-5f1d71f916fd982ef5a7ab0b78e419d64d02a77c7422911db11911a74f3a8ee6R394

vrajroutu:
@pamelafox 
There's a problem with the code where the /Content route can't show citations for file types other than `.pdf.`

pamelafox:
@vrajroutu Are you saying that there's a regression in this branch? It's already the case in main that clicking on a docx, for example, will result in the document being downloaded. We should resolve that, but I think that's independent of this - i.e. its already a problem for non-user uploaded documents.

DSRC-ACS:
> @zedhaque Your issue should now be resolved, with my recent change to make sure the RBAC roles are scoped to the current resource group.
> 
> @DSRC-ACS Did you set USE_USER_UPLOAD true? This is the code that should have constructed the container: https://github.com/Azure-Samples/azure-search-openai-demo/pull/1395/files#diff-5f1d71f916fd982ef5a7ab0b78e419d64d02a77c7422911db11911a74f3a8ee6R394

Hi @pamelafox ,

Thanks, I was able to fix the issue.

Now, I'm facing a different issue. I'm able to upload and ingest successfully via http://localhost:50505 in my local. But when it's deployed to app service, I'm getting 403 forbidden error for the same endpoint.

I've also added the redirect uri of app service in the app registration and I'm able to login successfully. Can you please help me regarding this.

app/backend/app.py
```diff
@@ -205,15 +231,78 @@ def config():
...
+        # If no files were included in the request, return an error response
+        return jsonify({"message": "No file part in the request", "status": "failed"}), 400
+
+    user_oid = auth_claims["oid"]
+    file = request_files.getlist("file")[0]
+    user_blob_container_client: FileSystemClient = current_app.config[CONFIG_USER_BLOB_CONTAINER_CLIENT]
+    user_directory_client = user_blob_container_client.get_directory_client(user_oid)
+    try:
+        await user_directory_client.get_directory_properties()
+    except ResourceNotFoundError:
```
mattgotteiner, line 253: I think we can do this in a follow up pr - but should we have some "rollback" code that cleans up partially uploaded files in case 1 of these steps fails?

mattgotteiner (approved):

app/backend/app.py
pamelafox: Just thinking the main creation steps that could fail:

* create_directory: If that fails, nothing's been uploaded, so they'd need to retry in the browser.
* upload_data: Ditto
* add_file: If that fails, the document is in their directory, but not in the index. That should be fine since the deletion code also deletes from file client before deleting from index, so they should be able to delete it still, and should be able to list it. But they should also get an error in the browser that'd prompt them to retry upload, and the next upload should override the previous.

I just tested out failing, and realized the UI wasn't showing a good error, so I'm amending that now.

pamelafox: Failed upload:
Screenshot 2024-04-09 at 2 34 53 PM

Successful upload after:
Screenshot 2024-04-09 at 2 35 45 PM

MIBS-YN:
Hi @pamelafox ,
Just FYI: In Windows, the path to the python.exe folder in .venv is "${workspaceFolder}/.venv/Scripts/python" rather than "${workspaceFolder}/.venv/bin/python" in the line 13 of the "launch.json" file, I believe...

peeushgoyalcisa:
Hi @pamelafox Thanks for continuing to support this repository. Unrelated question - but I just ran into this other repo [PubSec-Info-Assistant](https://github.com/microsoft/PubSec-Info-Assistant/tree/main) which seems to be a derivative of this main repo but has upload screens etc built out already. Wondering if you or someone can comment on the differences? Which repo should be used when, for instance?

pamelafox:
@peeushgoyalcisa I was not aware of that repo, it looks like they've adapted it to use a fairly different architecture, based on the architecture diagram. We have docs about how this sample compares to other samples @ https://github.com/Azure-Samples/azure-search-openai-demo/blob/main/docs/other_samples.md but that won't ever cover all the samples. You could try posting in that repo's issue tracker to see if they're familiar with our repo here and can make a comparison.

beouk:
@pamelafox Quick try and I got 

> `Message: The request is invalid. Details: The property 'storageUrl' does not exist on type 'search.documentFields'.`

However I'm coming from your old [upload-docs branch](https://github.com/Azure-Samples/azure-search-openai-demo/pull/1395/commits/3a0155b20d56899b5c44b026efe7cdcc806294a2) to this so it might be that the index needs to be rebuilt and not related.

pamelafox:
@beouk Did you run enable_acls per the updated README?
https://github.com/Azure-Samples/azure-search-openai-demo/pull/1535/files#diff-75e3da8fd5af9aaa591ace5f5653b494b44dbc24366b287e735f675ee8f0c14b
That'll add the field for you.

kk0205:
> @DSRC-ACS 
>Now, I'm facing a different issue. I'm able to upload and ingest successfully via http://localhost:50505/ in my local. But when it's deployed to app service, I'm getting 403 forbidden error for the same endpoint.
> I've also added the redirect uri of app service in the app registration and I'm able to login successfully. Can you please help me regarding this. Any other configuration to be done?

Hello everyone!

I'm facing the 403 forbidden problem too  
is there any solution to solve it.

PR #1399, opened 2024-03-10
Set Azure OpenAI API version in global configuration

sogue:
## Purpose

Set Azure OpenAI API version in global configuration
[Older versions retire on 2 April 2024. ](https://learn.microsoft.com/en-us/azure/ai-services/openai/reference)

```
[ ] Yes
[x] No
```

```
[ ] Yes
[x] No
```

## Type of change

```
[ ] Bugfix
[x] Feature
[] Refactoring (no functional changes, no api changes)
[ ] Other... Please describe:
```

## Code quality checklist


sogue:

app/backend/app.py
```diff
@@ -308,9 +308,10 @@ async def setup_clients():
         else:
             endpoint = f"https://{AZURE_OPENAI_SERVICE}.openai.azure.com"
 
-        # Store on app.config for later use inside requests
+        api_version = os.environ["AZURE_OPENAI_API_VERSION"]
```
pamelafox, line 311: This will cause a key error in old environments, please use get() and provide a default value.

pamelafox (changes requested):
I think an env variable is a good idea, but we cant assume it exists.

app/backend/app.py
pamelafox: The latest. I just need to make sure it works with gpt-4-vision but I'll test that before merging.

app/backend/app.py
```diff
@@ -308,9 +308,10 @@ async def setup_clients():
         else:
             endpoint = f"https://{AZURE_OPENAI_SERVICE}.openai.azure.com"
 
-        # Store on app.config for later use inside requests
+        api_version = os.environ.get("AZURE_OPENAI_API_VERSION", "2024-03-01-preview")
```
pamelafox, line 311:
```suggestion
        api_version = os.getenv("AZURE_OPENAI_API_VERSION", "2024-03-01-preview")
```

slightly shorter wrapper func

infra/main.bicep
```diff
@@ -42,6 +42,7 @@ param appServiceSkuName string // Set in main.parameters.json
 param openAiHost string // Set in main.parameters.json
 param isAzureOpenAiHost bool = startsWith(openAiHost, 'azure')
 param azureOpenAiCustomUrl string = ''
+param azureOpenAiApiVersion string = ''
```
pamelafox, line 45: Could you also add to outputs at the bottom, just in case someone sets it by changing line 45? (vs using `azd env set`)

app/backend/app.py
```diff
@@ -308,9 +308,10 @@ async def setup_clients():
         else:
             endpoint = f"https://{AZURE_OPENAI_SERVICE}.openai.azure.com"
 
-        # Store on app.config for later use inside requests
+        api_version = os.getenv("AZURE_OPENAI_API_VERSION", "2024-03-01-preview")
```
pamelafox, line 311: Sorry, I just tested in another repo, and we actually need this:

```suggestion
        api_version = os.getenv("AZURE_OPENAI_API_VERSION") or "2024-02-15-preview"
```

To account for that variable being an empty string (versus None)

pamelafox (approved):
Just tested with normal gpt-35 requests and gpt-4-vision requests in an existing environment, new version seems to work fine. Will merge after my deploy is successful as well.

PR #1355, opened 2024-03-01
Don't disable text chunking when GPT4vision is enabled

pamelafox:
## Purpose

Fixes #1353 

This PR enables chunking when GPT-4-vision is enabled, and also makes sure that our code in the vision approach can gracefully handle chunks without an image (it'll just send the text, not th image). This enables developers to have indexes with a mix of document types, both PDFs and xlsx/docx (formats not yet supported by our vision code).

A further enhancement: if *none* of the search results found in chatreadretrievevision.py have associated images, we could send the request to the non-vision model, to save time/cost. That'd require passing the non-vision model information into the vision constructors, which should work fine. I can do that in this PR or a follow-up PR.

## Does this introduce a breaking change?

When developers merge from main and run the server, azd up, or azd deploy, will this produce an error?
If you're not sure, try it out on an old environment.

```
[ ] Yes
[X] No
```

## Does this require changes to learn.microsoft.com docs?

This repository is referenced by [this tutorial](https://learn.microsoft.com/azure/developer/python/get-started-app-chat-template)
which includes deployment, settings and usage instructions. If text or screenshot need to change in the tutorial,
check the box below and notify the tutorial author. A Microsoft employee can do this for you if you're an external contributor.

```
[ ] Yes
[X] No
```

## Type of change

```
[X] Bugfix
[ ] Feature
[ ] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[ ] Documentation content changes
[ ] Other... Please describe:
```

## Code quality checklist

See [CONTRIBUTING.md](../CONTRIBUTING.md#submit-pr) for more details.

- [X] The current tests all pass (`python -m pytest`).
- [X] I added tests that prove my fix is effective or that my feature works
- [X] I ran `python -m pytest --cov` to verify 100% coverage of added lines
- [X] I ran `python -m mypy` to check for type errors
- [X] I either used the pre-commit hooks or ran `ruff` and `black` manually on my code.

PR #1407, opened 2024-03-11
Refactoring of prepdocs for easier integration with user upload feature

pamelafox:
## Purpose

This PR includes only the changes to prepdocs that are needed for the user document upload feature in #1395. This way, we can merge these changes first, and then that PR will simply *move* the files into app/backend, instead of move and edit.

The notable changes are:

* print -> logger.info(). We use our own logger so that we can set its level to INFO. If we set global level to INFO, then we also get Azure Python SDK stuff that's quite noisy right now.
* Reorg of the functions in prepdocs.py to be more reusable by app/backend, with additional typing and validation. No more passing around of args!
* Additional option to remove_content to remove documents with a single oid assigned
* Changed the id function to include an acl hash if acls are defined. The goal is to avoid id collisions between user-uploaded files and company-uploaded files.

```
[ ] Yes
[X] No - Tested in a manual ingestion env and int vect env.
```

```
[ ] Yes
[X] No
```

## Type of change

```
[ ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## Code quality checklist

- [X] I added tests that prove my fix is effective or that my feature works: **Added tests for everything but filestrategy.py as that currently has no tests, requires a ton of mocking.**


pamelafox:
I think it'd be good to merge https://github.com/Azure-Samples/azure-search-openai-demo/pull/1355 before this one as they both touch same lines.

scripts/prepdocs.py
```diff
@@ -32,50 +32,159 @@
...
+async def setup_search_info(
+    search_service: str,
+    index_name: str,
+    azure_credential: AsyncTokenCredential,
+    search_key: Union[str, None] = None,
+    key_vault_name: Union[str, None] = None,
+    search_secret_name: Union[str, None] = None,
+) -> SearchInfo:
+    if key_vault_name and search_secret_name:
+        key_vault_client = SecretClient(
```
mattgotteiner, line 54: does this support async with?

scripts/prepdocs.py
```diff
@@ -388,28 +364,24 @@ async def main(strategy: Strategy, credential: AsyncTokenCredential, args: Any):
...
     parser.add_argument(
         "--useintvectorization",
         required=False,
         help="Required if --useintvectorization is specified. Enable Integrated vectorizer indexer support which is in preview)",
     )
     parser.add_argument("--verbose", "-v", action="store_true", help="Verbose output")
     args = parser.parse_args()
+
+    if args.verbose:
+        # set formatter to only show the message
```
mattgotteiner, line 381: should we comment to state this is for the SDK workaround as discussed?

mattgotteiner (approved):

scripts/prepdocs.py
pamelafox: yep! changing to that.

scripts/prepdocs.py
pamelafox: Good idea, adding

javapro13:
Hello @pamelafox I just installed the latest from https://github.com/Azure-Samples/azure-search-openai-demo/tree/main and do not see the + manage file upload button on the main page. Do I need to do anything additional for the users to be able to upload the files? This is a critical functionality for us

pamelafox:
@javapro13 Yes, you need to follow the steps here to enable it: https://github.com/Azure-Samples/azure-search-openai-demo/blob/main/docs/deploy_features.md#enabling-user-document-upload

PR #1417, opened 2024-03-13
Add minimum score criteria for AI search results

sogue:
## Purpose

A filtering mechanism based on minimum score criteria for documents. By setting a minimum value for both `search_score` and `reranker_score`. This aims to improve the relevance of documents sent to the AI, that only the most relevant information is sent for further processing.

```
[ ] Yes
[x] No
```

```
[X] Yes
[ ] No
```

## Type of change

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

## Code quality checklist

De

### DIFF
--- a/app/backend/approaches/approach.py
+++ b/app/backend/approaches/approach.py
@@ -165,10 +165,11 @@ class Approach(ABC):
                 )
 
             qualified_documents = [
-                doc for doc in documents
+                doc
+                for doc in documents
                 if (
-                    (doc.score or 0) >= (minimum_search_score or 0) and
-                    (doc.reranker_score or 0) >= (minimum_reranker_score or 0)
+                    (doc.score or 0) >= (minimum_search_score or 0)
+                    and (doc.reranker_score or 0) >= (minimum_reranker_score or 0)
                 )
             ]
 
--- a/app/backend/approaches/approach.py
+++ b/app/backend/approaches/approach.py
@@ -123,6 +123,8 @@ class Approach(ABC):
         vectors: List[VectorQuery],
         use_semantic_ranker: bool,
         use_semantic_captions: bool,
+        min_search_score: Optional[float],
+        min_reranker_score: Optional[float],
     ) -> List[Document]:
         # Use semantic ranker if requested and if retrieval mode is text or hybrid (vectors + text)
         if use_semantic_ranker and query_text:
@@ -157,11 +159,17 @@ class Approach(ABC):
                         oids=document.get("oids"),
                         groups=document.get("groups"),
                         captions=cast(List[QueryCaptionResult], document.get("@search.captions")),
-                        score=document.get("@search.score"),
-                        reranker_score=document.get("@search.reranker_score"),
+                        score=document.get("@search.score", 0.0),
+                        reranker_score=document.get("@search.reranker_score", 0.0),
                     )
                 )
-        return documents
+
+            qualified_documents = [
+                doc for doc in documents
+                if doc.search_score >= min_search_score and doc.reranker_score >= min_reranker_score
+            ]
+
+        return qualified_documents
 
     def get_sources_content(
         self, results: List[Document], use_semantic_captions: bool, use_image_citation: bool
--- a/app/backend/approaches/approach.py
+++ b/app/backend/approaches/approach.py
@@ -123,8 +123,8 @@ class Approach(ABC):
         vectors: List[VectorQuery],
         use_semantic_ranker: bool,
         use_semantic_captions: bool,
-        min_search_score: Optional[float],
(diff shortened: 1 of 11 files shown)


sogue:
@zedhaque No, enabling semantic ranker and semantic captions isn't required for this change to work. The filtering based on search_score operates independently of those features. But for reranker_score it is needed. If one of the scores don't meet the threshold, documents won't be sent to the AI for further processing. Instead, you could either retrieve more documents hoping some meet the criteria or inform the user that no relevant information was found, depending on your implementation.

app/frontend/src/pages/ask/Ask.tsx
```diff
@@ -259,6 +279,30 @@ export function Component(): JSX.Element {
...
+                    className={styles.chatSettingsSeparator}
+                    label="Min Search Score"
+                    min={0}
+                    max={1}
+                    step={0.1}
+                    defaultValue={minSearchScore}
+                    onChange={onMinSearchScoreChange}
+                    showValue
+                    snapToStep
+                />
```
pamelafox, line 292: I wonder if 0.1/0.2/0.3 is detailed enough, given it only goes from 0 to 1? More importantly, as far as I understand, this score varies based on whether the mode is vector, hybrid, or text:
https://learn.microsoft.com/en-us/azure/search/hybrid-search-ranking#scores-in-a-hybrid-search-results
So I think it'd have to be an `input type=number` or whatever the equivalent FluentUI component is.

Please correct me if I'm wrong though.

app/frontend/src/pages/ask/Ask.tsx
```diff
@@ -259,6 +279,30 @@ export function Component(): JSX.Element {
                     snapToStep
                 />
 
+                <Slider
+                    className={styles.chatSettingsSeparator}
+                    label="Min Search Score"
```
pamelafox, line 284:
```suggestion
                    label="Minimum search score"
```

app/frontend/src/pages/ask/Ask.tsx
```diff
@@ -259,6 +279,30 @@ export function Component(): JSX.Element {
...
+                    step={0.1}
+                    defaultValue={minSearchScore}
+                    onChange={onMinSearchScoreChange}
+                    showValue
+                    snapToStep
+                />
+
+                <Slider
+                    className={styles.chatSettingsSeparator}
+                    label="Min Reranker Score"
```
pamelafox, line 296:
```suggestion
                    label="Minimum reranker score"
```

...if we have room for "minimum", I think we probably do?

pamelafox (changes requested):
Thanks so much for the PR! We very much need this functionality. I have a comment about the score range, and there are also failing mypy checks, so please address those. Also will assign a reviewer from search team since this is a search specific change.

pamelafox:
This PR *would* require an update to the learn.microsoft.com tutorial by the way, as that tutorial includes a screenshot of developer settings and an explanation of each setting. FYI @diberry 
I will wait for this to be approved before sending a PR for the tutorial.

sogue:
> Thanks so much for the PR! We very much need this functionality. I have a comment about the score range, and there are also failing mypy checks, so please address those. Also will assign a reviewer from search team since this is a search specific change.

Thanks for the feedback! I've fixed the mypy issues and made the score range input flexible as suggested. The PR is updated and ready for the next review.

app/frontend/src/pages/ask/Ask.tsx
sogue: Yes, you're right. The score range varies significantly depending on the search method. I've updated the UI to replace the slider with a numeric input field.

sogue:
> Thanks so much for the PR! We very much need this functionality. I have a comment about the score range, and there are also failing mypy checks, so please address those. Also will assign a reviewer from search team since this is a search specific change.

Thanks for the feedback! I've fixed the mypy issues and made the score range input flexible as suggested. The PR is updated and ready for the next review.

app/backend/approaches/approach.py
```diff
@@ -161,7 +163,16 @@ async def search(
                         reranker_score=document.get("@search.reranker_score"),
                     )
                 )
-        return documents
+
+            qualified_documents = [
```
pamelafox, line 167: Could you add a unit test in our pytest test suite that checks this logic? You can test the Approach.search() function directly and use a mock of SearchClient.search, we have that mocked in several places in our tests. Let me know if you need further guidance. Thanks!

pamelafox (commented):
Thank you, those changes look good. I've asked @mattgotteiner to review.

mattgotteiner (approved):
Thanks!

pamelafox (approved):

PR #1428, opened 2024-03-15
Add more tests to the splitter

tonybaloney:
## Purpose

Verify that the splitter will always yield the same content as the input

Noticed that if you set the overlap percent to 0 you get a DBZ error

```
[ ] Yes
[ ] No
```

```
[ ] Yes
[ ] No
```

## Type of change

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## Code quality checklist


tonybaloney:
screenshot 2024-03-15 at 14 32 49
There is some logic here which isn't configurable to overlap sentences, so the diffs don't match perfectly, so I'm going to test that the split diffs aren't **missing** any content from the original text

pamelafox:
@tonybaloney What's current status of this PR? Shall we close? Seems to still be in draft.

PR #1502, opened 2024-04-08
Enforce RBAC only for non-free search service

pamelafox:
## Purpose

The goal of this PR is to improve the security of the provisioned search service, which is currently triggering alerts in my subscription with the reason "Role-based API authetication for Azure Cognitive Search must be enabled." (sic).

I was concerned that it would disable our ability to use the search service, since the parameter is named disableLocalAuth, but it seems to be misnamed, as local RBAC'd search service access works fine.

```
[ ] Yes
[X] No
```

```
[ ] Yes
[X] No
```

## Type of change

```
[X] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## Code quality checklist

- [ ] I added tests that prove my fix is effective or that my feature works - N/A, not possible
- [ ] I ran `python -m pytest --cov` to verify 100% coverage of added lines - N/A
- [ ] I ran `python -m mypy` to check for type errors - N/A


infra/core/search/search-services.bicep
```diff
@@ -47,7 +47,7 @@ resource search 'Microsoft.Search/searchServices@2021-04-01-preview' = {
   // The free tier does not support managed identity
   identity: searchIdentityProvider
   properties: {
-    authOptions: authOptions
+    authOptions: disableLocalAuth ? null : authOptions
```
pamelafox, line 50: It causes an error if authOptions is an empty object, which is the default in this file, and it didn't seem to work to set that to null in main.bicep (perhaps it gets set to an empty object as it considers null to be a non-setting?).

I can make this change in azure-dev if this PR is approved. cc @jongio

infra/main.bicep
```diff
@@ -418,11 +418,8 @@ module searchService 'core/search/search-services.bicep' = {
     name: !empty(searchServiceName) ? searchServiceName : 'gptkb-${resourceToken}'
     location: !empty(searchServiceLocation) ? searchServiceLocation : location
     tags: tags
-    authOptions: {
-      aadOrApiKey: {
-        aadAuthFailureMode: 'http401WithBearerChallenge'
```
pamelafox, line 423: Was this doing anything for us?? I didn't notice a change with local server, local prepdocs, or deployed app.

infra/core/search/search-services.bicep
mattgotteiner: the default is null not an empty object right? I do not think empty object is valid according to our swagger, but null is

infra/main.bicep
mattgotteiner: This explicitly sets that we can use both aad auth and api key auth. If a customer is using a pre-existing service that only supports api key auth they might notice a problem if this template isn't present because it would enable rbac

infra/core/search/search-services.bicep
pamelafox: The default at the top of this file is:

param authOptions object = {}

So that's what got set if I removed any specification of authOptions in main.bicep. However, the caused an error in the search service in combination with disableLocalAuth. Hence, my change here to explicitly set it to null when disableLocalAuth is set to true.

I also tried to set authOptions:null in main.bicep, to avoid having to change in Bicep, but I still got the error, so I think Bicep might see the "null" and decide that it's a falsy value that should be replaced with the default {} in this file (versus keeping it as null).

infra/main.bicep
pamelafox: Ah. I don't think we had customers doing that, since our app.py code only fetches a secret if there's a keyvault and a secret name for that keyvault (i.e. the changes we introduced for free tier support).

I could add these authOptions back only for free tier, but are they necessary? Looks like I need to test my free tier environment.

pamelafox: My free environment seems to also work fine with this change (local, prepdocs, and up).

infra/core/search/search-services.bicep
mattgotteiner: hmm, good to know

infra/main.bicep
mattgotteiner: ok, sounds like this works

mattgotteiner (approved):

PR #1510, opened 2024-04-09
Port to the Graph SDK for authentication scripts

pamelafox:
## Purpose

The msgraph-sdk is now publicly usable, as of October, so this PR ports our authentication setup scripts to use the graph SDK instead of raw HTTP calls.

```
[ ] Yes
[X] No
```

```
[ ] Yes
[X] No
```

## Type of change

```
[ ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## Code quality checklist

- [ ] I added tests that prove my fix is effective or that my feature works: N/A, we have no tests for this yet, and it would be very mock-heavy. 
- [ ] I ran `python -m pytest --cov` to verify 100% coverage of added lines: See above
- [X] I ran `python -m mypy` to check for type errors: msgraph is causing mypy errors so I had to add to overrides, and I filed https://github.com/microsoftgraph/msgraph-sdk-python/issues/655


pyproject.toml
```diff
@@ -26,6 +26,8 @@ python_version = 3.9
 
 [[tool.mypy.overrides]]
 module = [
-    "msal.*"
+    "msal.*",
+    "msgraph.*",
```
pamelafox, line 30: Filed https://github.com/microsoftgraph/msgraph-sdk-python/issues/655

scripts/auth_common.py
```diff
@@ -1,37 +1,16 @@
...
-    async with aiohttp.ClientSession(headers=auth_headers, timeout=aiohttp.ClientTimeout(total=TIMEOUT)) as session:
-        async with session.get(f"https://graph.microsoft.com/v1.0/applications(appId='{app_id}')") as response:
-            if response.status == 200:
-                response_json = await response.json()
-                return response_json["id"]
-
-    return None
-
-
-async def update_application(auth_headers: Dict[str, str], object_id: str, app_payload: object):
```
pamelafox, line 25: I removed this function entirely as it ended up a single line of code but can add it back if you prefer it to the (longer) line of code.

scripts/auth_init.py
```diff
@@ -86,89 +80,88 @@ def random_app_identifier():
...
-                    "id": "7b207263-0c4a-4127-a6fe-38ea8c8cd1a7",
-                    "adminConsentDisplayName": "Access Azure Search OpenAI Chat API",
-                    "adminConsentDescription": "Allows the app to access Azure Search OpenAI Chat API as the signed-in user.",
-                    "userConsentDisplayName": "Access Azure Search OpenAI Chat API",
-                    "userConsentDescription": "Allow the app to access Azure Search OpenAI Chat API on your behalf",
-                    "isEnabled": True,
-                    "value": "access_as_user",
-                    "type": "User",
-                }
+def server_app_initial(identifier: int) -> Application:
```
mattgotteiner, line 83: Do we even need this method anymore? 1 liners can be inlined IMO

mattgotteiner (approved):
Will we update authentication.py in the backend in a future PR?

scripts/auth_init.py
pamelafox: No, but I think in this case, it's nice and symmetrical with the other Application objects, so it's somewhat easier to read when it's inside a function like the others, I think. (It reads top to down)

pamelafox:
@mattgotteiner Yeah I only noticed afterwards that authentication.py has at least one graph call (transitiveMemberOf). (There's also key_url but I wasn't sure if that's graph SDK'able, didn't see a mention of SDK in linked doc). I could move transitiveMemberOf.

PR #1518, opened 2024-04-10
Disable key-based storage account access in Bicep

pamelafox:
## Purpose

This PR disables key-based access to the storage accounts (both normal and ADLS2) in the Bicep. This doesn't affect standard usage since we use DefaultAzureCredential by default. We do provide the option in prepdocs.py to supply a key, but developers have to come up with that key themselves, so if they really want, they can turn key access back on and use it.

This PR also removes some storage-related lines of seemingly deadcode.

```
[ ] Yes
[X] No
```

```
[ ] Yes
[X] No
```

## Type of change

```
[ ] Bugfix
[ ] Feature
[X] Other... Please describe: Security!
```

## Code quality checklist

- [ ] I added tests that prove my fix is effective or that my feature works: N/A
- [ ] I ran `python -m pytest --cov` to verify 100% coverage of added lines: N/A


app/backend/approaches/retrievethenread.py
```diff
@@ -9,10 +8,6 @@
 from core.authentication import AuthenticationHelper
 from core.messagebuilder import MessageBuilder
 
-# Replace these with your own values, either in environment variables or directly here
```
pamelafox, line 12: This was added in the GPT4-vision merge. I'm guessing it was added during development and never removed. We set up the storage accounts in app.py now.

mirojs:
@pamelafox Does it mean that all MS b2c tenant users (both new and old) will have to face more security challenges?

pamelafox:
@mirojs Are you currently using key-based access to storage? This template currently does not use storage keys by default.

mattgotteiner (approved):
we never used access via key in the backend, and the frontend never directly reaches out to the storage account, so I don't see how this causes any problems

mirojs:
@pamelafox No, I'm not. But all the b2c auth doesn't support RBAC, which leaves a vulnerable spot there in the app. Is my understanding correct?

pamelafox:
@mirojs Ah, you are correct that currently we use client secret when authenticating our Entra app registrations with the MSAL SDK. We would like to move away from that secret and are working with the Entra teams to figure out how we can do that. Stay tuned!

PR #1487, opened 2024-04-03
Add Tooltips to Developer Settings

john0isaac:
## Purpose

fix #1334

Thanks to @diberry, your tooltip works like a charm!

image

## Does this introduce a breaking change?

When developers merge from main and run the server, azd up, or azd deploy, will this produce an error?
If you're not sure, try it out on an old environment.

```
[ ] Yes
[X] No
```

## Does this require changes to learn.microsoft.com docs?

This repository is referenced by [this tutorial](https://learn.microsoft.com/azure/developer/python/get-started-app-chat-template)
which includes deployment, settings and usage instructions. If text or screenshot need to change in the tutorial,
check the box below and notify the tutorial author. A Microsoft employee can do this for you if you're an external contributor.

```
[ ] Yes
[X] No
```

## Type of change

```
[ ] Bugfix
[X] Feature
[ ] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[ ] Documentation content changes
[ ] Other... Please describe:
```

## Code quality checklist

See [CONTRIBUTING.md](https://github.com/Azure-Samples/azure-search-openai-demo/blob/main/CONTRIBUTING.md#submit-pr) for more details.

- [X] The current tests all pass (`python -m pytest`).
- [X] I added tests that prove my fix is effective or that my feature works
- [X] I ran `python -m pytest --cov` to verify 100% coverage of added lines
- [X] I ran `python -m mypy` to check for type errors
- [X] I either used the pre-commit hooks or ran `ruff` and `black` manually on my code.

PR #1522, opened 2024-04-11
Add clickable help icons for developer settings

### LINKED ISSUE #1334: Add tool tips to settings in Developer Settings

```
- [x] feature request
```

pamelafox:
## Purpose

Fixes #1334

This is an alternative to #1487 which instead adds clickable help icons. I was unable to get the tooltips to pop up on my machine (not sure if it's a Mac thing or a user error?) and clickable icons are generally more accessible, so I made a branch to try this possibility.
I had to move some of the fields over when the components did not support renderLabel, like SpinButton and ChoiceGroup, but the end result is very similar functionality-wise.

Note that we have a HUGE amount of repeated code between Chat.tsx and Ask.tsx. We could potentially make a DeveloperSettings component that took all the defaults/callbacks. I had to do a lot of repetitive updating across both.

Also realized that our VectorFields setting is weird- it's a list, but we never send multiple values. I think it should just be a string. But that's for another day.

Screenshots:

Screenshot 2024-04-11 at 12 46 38 PM
Screenshot 2024-04-11 at 12 46 32 PM

```
[ ] Yes
[X] No
```

```
[X] Yes - new screenshot needed, possibly new settings text
[ ] No
```

## Type of change

```
[ ] Bugfix
[X] Feature
[ ] Other... Please describe:
```

## Code quality checklist

- [X] I added tests that prove my fix is effective or that my feature works: No - This is a frontend only change. I could potentially add an E2E test but not clear if needed.
- [ ] I ran `python -m pytest --cov` to verify 100% coverage of added lines: No, since coverage doesnt include JS code


.github/workflows/python-test.yaml
```diff
@@ -63,7 +63,7 @@ jobs:
           id: e2e
           if: runner.os != 'Windows'
           run: |
-            playwright install --with-deps
+            playwright install chromium --with-deps
```
pamelafox, line 66: This should speed up CI, we only run in chromium right now, that's the default.

app/frontend/src/i18n/tooltips.ts
```diff
@@ -0,0 +1,29 @@
+// Keep values less than 20 words.
+// Don't add links to the tooltips.
+export const toolTipText = {
```
mattgotteiner, line 3: Thanks this looks a lot better now

mattgotteiner (approved):

PR #1524, opened 2024-04-12
Add python-dotenv to eliminate error in Github CodeSpaces

Tylersuard:
## Purpose

Currently, when I try to run this repo in CodeSpaces, I get an error saying 

` Error: there is no registered task type 'dotenv'. Did you miss installing an extension that provides a corresponding task provider? `

[ ] Yes
[X ] No

```
[ ] Yes
[X ] No
```

## Type of change

```
[x ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -9,6 +9,7 @@ playwright
 pytest-cov
 pytest-playwright
 pytest-snapshot
+python-dotenv
 pre-commit
 locust
 pip-tools


Tylersuard:
@pamelafox Just a small fix for an error I was getting when using this repo with codespaces.  I just added python-dotenv to the requirements.

pamelafox:
Hm, so I have seen that error as well occasionally, and I believe it occurs when your Azure Developer CLI extension has not yet loaded, as it's the extensions that's registered for that task type:

Screenshot 2024-04-20 at 2 02 26 PM

If this fix worked for you, I'm guessing it's because the extension loaded in the right order the second time around, so the error didn't pop out. I get all sorts of weird errors from slow loading extensions in Codespaces/Dev Containers, unfortunately, and I think this is one of them. In fact, I even reported the same error myself in an azd thread:
https://github.com/Azure/azure-dev/issues/2852#issuecomment-1915216058

Can you try reverting this change and see how frequently you're able to produce the error? You could also look at your Codespaces logs when it happens. There might be a specific log for extensions. Then you can put all the details in an issue here, or in https://github.com/Azure/azure-dev/, or in https://github.com/orgs/community/discussions/categories/codespaces

Going to close this since python-dotenv isn't a requirement.

MiguelElGallo:
Sadly this is back, opened #2033

PR #1535, opened 2024-04-13
Adds storageURL field to track file location

### LINKED ISSUE #1424: Prepdocs.py with ADLS Gen2 enabled re-uploads documents to the storage

```
- [ ] feature request
```

### Minimal steps to reproduce
Follow the instructions in `login_and_acl.md`, "Optional Scripts" section

### Expected/desired behavior
Documents are uploaded to directories as defined in `sampleacls.json`, then ingested into index with "sourcefile" value including directory where the blob is.

### OS and Version?
Ubuntu 22.04

### azd version?
azd version 1.6.1

After initial ADLS Gen2 setup and running `./scripts/asldgen2setup.ps1`, the storage container is populated with files in directories according to mapping in `sampleacls.json`.
When running `prepdocs.py` it re-uploads the file to the storage and prints "Uploading blob for whole file" in the output.
The index entries also have `sourcefile` value without any reference to the directory. This renders document-level security via directory ACL useless.

### LINKED ISSUE #1531: manageacls.py could accidentally grant permissions to wrong document in case of name clash
```
- [X] bug report -> please search issues before submitting
- [ ] feature request
```

### Minimal steps to reproduce

1. Upload document with name fileA.pdf using prepdocs.sh
2. Upload different document with name fileA.pdf using user-upload feature from webapp
3. Use manageacls.py to add ACL for fileA.pdf to user B. It'll search for all files named fileA.pdf, including ones that were user uploaded, since it doesn't have a way to distinguish user-uploaded from admin-uploaded.

### Suggested resolution

- [x] Add a field "sourcepath" or "blobpath" in searchmanager setup that will will contain the full path of the document: storageaccount/container/directory
- [x] Add that field to manageacls.py enable_acls as well 
- [x] Set that field from both user-upload and prepdocs ingestion path
- [x] Perhaps provide a script to auto-adjust existing paths?
- [x] Note in the login_and_acls doc that a longer path must be specified for manageacls. Though we could provide an argument or shortcut for saying "default".

pamelafox:
## Purpose

Fixes #1531 
Possibly fixes #1424

With the recent addition of the user upload feature, it is now possible for a file with the same filename to exist in two places (either the user's storage container or the global content container). I accounted for that in terms of searching and user deletion, but realized that manageacls.py didn't have a way to unambiguously identify what document's ACLs to update. 

So this PR adds a new field "storageUrl" that stores the full Blob storage URL (*not* with any SAS tokens however), and changes manageacls.py to expect a full URL to be passed in. It's slightly more work since you need to go into Blob storage and copy the URL from the portal, but much more clear what document you're trying to ACL.

I also improved logging in manageacls.py so that we can recommend the verbose flag always and provide more details about whether we've successfully found the document, found acls to remove, etc.

I also added a quick fix for #1424 to avoid uploading to BlobStorage in the case that a file has an existing blob storage URL (which is true for ADLS2 sources). I don't have an azd environment to test that out yet however, and we still have to improve GPT-4-vision code path so it uploads page images to the right place.

```
[ ] Yes
[X] No
```

```
[ ] Yes
[X] No
```

## Type of change

```
[X] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## Code quality checklist

- [x] The current tests all pass (`python -m pytest`)

### DIFF
--- a/app/backend/prepdocslib/searchmanager.py
+++ b/app/backend/prepdocslib/searchmanager.py
@@ -115,7 +115,7 @@ class SearchManager:
                     name="storageUrl",
                     type="Edm.String",
                     filterable=True,
-                    facetable=True,
+                    facetable=False,
                 ),
             ]
             if self.use_acls:
--- a/app/backend/prepdocslib/searchmanager.py
+++ b/app/backend/prepdocslib/searchmanager.py
@@ -111,6 +111,12 @@ class SearchManager:
                     filterable=True,
                     facetable=True,
                 ),
+                SimpleField(
+                    name="storageUrl",
+                    type="Edm.String",
+                    filterable=True,
+                    facetable=True,
+                ),
             ]
             if self.use_acls:
                 fields.append(
@@ -182,7 +188,9 @@ class SearchManager:
             else:
                 logger.info("Search index %s already exists", self.search_info.index_name)
 
-    async def update_content(self, sections: List[Section], image_embeddings: Optional[List[List[float]]] = None):
+    async def update_content(
+        self, sections: List[Section], image_embeddings: Optional[List[List[float]]] = None, url: Optional[str] = None
+    ):
         MAX_BATCH_SIZE = 1000
         section_batches = [sections[i : i + MAX_BATCH_SIZE] for i in range(0, len(sections), MAX_BATCH_SIZE)]
 
@@ -209,6 +217,9 @@ class SearchManager:
                     }
                     for section_index, section in enumerate(batch)
                 ]
+                if url:
+                    for document in documents:
+                        document["storageUrl"] = url
                 if self.embeddings:
                     embeddings = await self.embeddings.create_embeddings(
                         texts=[section.split_page.text for section in batch]
--- a/scripts/manageacl.py
+++ b/scripts/manageacl.py
@@ -170,7 +170,7 @@ class ManageAcl:
                         name="storageUrl",
                         type="Edm.String",
                         filterable=True,
-                        facetable=True,
+                        facetable=False,
                     ),
                 )
             await search_index_client.create_or_update_index(index_definition)
--- a/scripts/manageacl.py
+++ b/scripts/manageacl.py
@@ -14,6 +14,8 @@ from azure.search.documents.indexes.models import (
     SimpleField,
 )
 
+logger = logging.getLogger("manageacl")
+
 
 class ManageAcl:
     """
@@ -24,7 +26,7 @@ class ManageAcl:
         self,
         service_name: str,
         index_name: str,
-        document: str,
+        url: str,
         acl_action: str,
         acl_type: str,
         acl: str,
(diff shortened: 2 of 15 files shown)


app/backend/prepdocslib/searchmanager.py
```diff
@@ -111,6 +111,12 @@ async def create_index(self, vectorizers: Optional[List[VectorSearchVectorizer]]
                     filterable=True,
                     facetable=True,
                 ),
+                SimpleField(
+                    name="storageUrl",
+                    type="Edm.String",
+                    filterable=True,
+                    facetable=True,
```
mattgotteiner, line 118: Recommend making this not facetable.

scripts/manageacl.py
```diff
@@ -75,53 +78,75 @@ async def run(self):
                 await self.remove_all_acls(search_client)
             elif self.acl_action == "add":
                 await self.add_acl(search_client)
+            elif self.acl_action == "update_storage_urls":
```
mattgotteiner, line 81: Is this doc'd in the readme?

scripts/manageacl.py
```diff
@@ -139,9 +164,42 @@ async def enable_acls(self, endpoint: str):
...
                     )
                 )
-
+            if not any(field.name == "storageUrl" for field in index_definition.fields):
+                index_definition.fields.append(
+                    SimpleField(
+                        name="storageUrl",
+                        type="Edm.String",
+                        filterable=True,
+                        facetable=True,
```
mattgotteiner, line 173: Same here to remove facetable

scripts/manageacl.py
```diff
@@ -139,9 +164,42 @@ async def enable_acls(self, endpoint: str):
...
+    async def update_storage_urls(self, search_client: SearchClient):
+        filter = "storageUrl eq ''"
+        documents = await search_client.search("", filter=filter, select=["id", "storageUrl", "oids", "sourcefile"])
+        found_documents = []
+        documents_to_merge = []
+        async for document in documents:
+            found_documents.append(document)
+            if len(document["oids"]) == 1:
+                logger.warning(
+                    "Not updating storage URL of document %s as it has only one oid and may be user uploaded",
```
mattgotteiner, line 187: Interesting - we should stop people from doing this? Is there an override, there might be a legitimate use case here

mattgotteiner (approved):

scripts/manageacl.py
pamelafox: I documented it in deploy_features: https://github.com/Azure-Samples/azure-search-openai-demo/pull/1535/files#diff-75e3da8fd5af9aaa591ace5f5653b494b44dbc24366b287e735f675ee8f0c14bR145
Will also put it in the release notes.

scripts/manageacl.py
pamelafox: Great question! I figured that if they really want to do it, they can change the Python code themselves to do it. I thought about adding an override argument but realized that with the way manageacl.py is structured, any argument becomes a top level instance variable, so that made me less excited to add arguments. I'd be fine with adding --override as well, but I think the usage would be pretty low, since this is only for folks migrating from old repos anyway.

PR #1576, opened 2024-05-01
Allow public documents when authentication is enabled

mattgotteiner:
## Purpose

* Currently, when authentication is enforced there's no way to select "public" documents. Public documents are documents with neither oid or group ids associated with them.
* Add AZURE_ENABLE_GLOBAL_DOCUMENT_ACCESS env var which allows searching on these documents even if authentication is enabled.
* Add AZURE_ENABLE_UNAUTHENTICATED_ACCESS env var which allows searching on public documents even when a user is not logged in, and AZURE_ENFORCE_ACCESS_CONTROL is enabled. 

```
[ ] Yes
[X] No
```

```
[X] Yes
[ ] No
```

## Type of change

```
[ ] Bugfix
[X] Feature
[ ] Other... Please describe:
```

## Code quality checklist


mattgotteiner:
Update docs

mattgotteiner:
Manual e2e test

mattgotteiner:
> Manual e2e test

Local test succeeded. Manually tested the following:
1. Unset oids, groups - publicly accessible
2. Set oids, unset groups - not publicly accessible
3. Unset oids, set groups - not publicly accessible

mattgotteiner:
> > Manual e2e test
> 
> Local test succeeded. Manually tested the following:
> 
> 1. Unset oids, groups - publicly accessible
> 2. Set oids, unset groups - not publicly accessible
> 3. Unset oids, set groups - not publicly accessible

deployed test succeeded as well

mattgotteiner:
> Thanks @mattgotteiner :)
> 
> Just wondering, suppose below are my env settings:
> 
> AZURE_USE_AUTHENTICATION = True AZURE_ALLOW_PUBLIC_DOCUMENTS = True AZURE_ENFORCE_ACCESS_CONTROL = True
> 
> This would allow users to search for public documents without login? currently the chat text area field is disabled, and you need to login to continue. If I understand correctly the idea here it to allow authenticated users to chat with public documents, correct?

They still need to login (ENFORCE_ACCESS_CONTROL makes login require). The only purpose of this PR is to allow authenticated users to chat with public documents.

app/backend/core/authentication.py
```diff
@@ -242,6 +252,7 @@ async def get_auth_claims_if_enabled(self, headers: dict) -> dict[str, Any]:
     async def check_path_auth(self, path: str, auth_claims: dict[str, Any], search_client: SearchClient) -> bool:
         # Start with the standard security filter for all queries
         security_filter = self.build_security_filters(overrides={}, auth_claims=auth_claims)
+        print(security_filter)
```
pamelafox, line 255:
```suggestion
```

docs/login_and_acl.md
```diff
@@ -38,6 +38,7 @@ The easiest way to setup the two apps is to use the `azd` CLI. We've written scr
 1. Run `azd env set AZURE_USE_AUTHENTICATION true` to enable the login UI and App Service authentication.
 1. Ensure access control is enabled on your search index. If your index doesn't exist yet, run prepdocs with `AZURE_USE_AUTHENTICATION` set to `true`. If your index already exists, run `pwsh ./scripts/manageacl.ps1 --acl-action enable_acls`.
 1. (Optional) To require access control when using the app, run `azd env set AZURE_ENFORCE_ACCESS_CONTROL true`.
+1. (Optional) To allow searching of documents with no access control configured, even when access control is required, run `azd env set AZURE_ALLOW_PUBLIC_DOCUMENTS true`.
```
pamelafox, line 41:
```suggestion
1. (Optional) To allow authenticated users to search on documents that have no access controls assigned, even when access control is required, run `azd env set AZURE_ALLOW_PUBLIC_DOCUMENTS true`.
```

May be clearer?

pamelafox:
Hm, I wonder if public would be clearer if it was "global"? public may imply logged out users (which Zed has a use case for).

AZURE_DEFAULT_DOCUMENT_ACCESS_IS_GLOBAL ??

Not very snappy! I'll sleep on it.

It'd also be great if we could provide smoke tests that verify the expected behavior to increase confidence, but I haven't setup Entra authenticated smoke tests before.

app/backend/core/authentication.py
mattgotteiner: my print debugging is leaking out :(

mattgotteiner:
> Hm, I wonder if public would be clearer if it was "global"? public may imply logged out users (which Zed has a use case for).
> 
> AZURE_DEFAULT_DOCUMENT_ACCESS_IS_GLOBAL ??
> 
> Not very snappy! I'll sleep on it.
> 
> It'd also be great if we could provide smoke tests that verify the expected behavior to increase confidence, but I haven't setup Entra authenticated smoke tests before.

We don't need Entra authentication IMO - we just need to mock the search behavior or have a search instance for smoke tests

PR #1620, opened 2024-05-16
Removing unneeded key for free search service

pamelafox:
## Purpose

I had previously added support for using a key for search service because I thought the free tier didn't support managed identity. As it turns out, it does support authenticating with managed identity- it just doesn't support having its *own* identity (i.e. outbound identity). That's necessary for things like integrated vectorization for it to auth to services, but we already don't expect integrated vectorization to work well on free tier.

So this PR makes free tier the same in terms of using keyless auth.

```
[ ] Yes
[X] No
```

```
[ ] Yes
[X] No
```

## Type of change

```
[X] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## Code quality checklist


infra/main.bicep
```diff
@@ -730,8 +685,8 @@ module privateEndpoints 'private-endpoints.bicep' = if (usePrivateEndpoint) {
     tags: tags
     resourceToken: resourceToken
     privateEndpointConnections: privateEndpointConnections
-    applicationInsightsId: monitoring.outputs.applicationInsightsId
-    logAnalyticsWorkspaceId: monitoring.outputs.logAnalyticsWorkspaceId
+    applicationInsightsId: useApplicationInsights ? monitoring.outputs.applicationInsightsId : ''
```
pamelafox, line 688: This was missing a conditional since monitoring is an optional deployment gated on useApplicationInsights

docs/deploy_lowcost.md
```diff
@@ -2,7 +2,7 @@
 
 This AI RAG chat application is designed to be easily deployed using the Azure Developer CLI, which provisions the infrastructure according to the Bicep files in the `infra` folder. Those files describe each of the Azure resources needed, and configures their SKU (pricing tier) and other parameters. Many Azure services offer a free tier, but the infrastructure files in this project do *not* default to the free tier as there are often limitations in that tier.
 
-However, if your goal is to minimize costs while prototyping your application, follow the steps below _before_ running `azd up`. Once you've gone through these steps, return to the [deployment steps](../README.md#deploying).
+However, if your goal is to minimize costs while prototyping your application, follow the steps below *before* running `azd up`. Once you've gone through these steps, return to the [deployment steps](../README.md#deploying).
```
pamelafox, line 5: Fixed Markdown lint issues while I was here

pamelafox:
And yes, some of this code may come back if I figure out keyvault storage for Entra secrets, but it may be in a different form, so I don't want to leave it in just in case - I'll just bring back what's needed.

mattgotteiner (approved):
Great! I like that we can remove the key vault code

PR #1625, opened 2024-05-17
Fix samples browser

diberry:


achandmsft:
@pamelafox could you please merge this one when you see this? Wanted it to be picked by the early morning 5AM PT merge into Learn.

achandmsft (approved):
Looks good.

pamelafox (commented):
Dina made the same change in another repo, this ones a harder change as I think the links/images are broken due to relative URLs? Prefer testing with the other repo.

PR #1682, opened 2024-06-04
Upgrade to latest version of AI Chat Protocol

pamelafox:
## Purpose

The AI Chat Protocol has now been further codified in an SDK (https://github.com/microsoft/ai-chat-protocol). We are not yet porting to the SDK due to our need for authentication and some questions around the token provider option in the SDK. However, we'd still like for our HTTP API to comply to the Chat Protocol. There are two differences from our current API design:

* Instead of sending `"stream": True` to get a streaming response, we send to a different path, "/chat/stream". That is more compatible with API documentation generators like Swagger/TypeSpec, as they currently don't have a way of specifying a completely different response type based off a parameter inside a JSON, but they definitely can vary response type based on path. It's also potentially better for monitoring when it's clear from the path that you're making a streaming response. It's actually what @srbalakr originally had in this repo, and we only changed it to match the original proposed version of the protocol. You live, you learn!
* Instead of passing down the full `choices` list from the OpenAI response, we pass down only the first choice, specifically it's message or delta. That's because we never use the other choices in the frontend, and we also don't use the additional OpenAI metadata in the frontend. So it simplifies our frontend code and reduces the size of responses sent to the browser.

```
[ ] Yes
[X] No
```

```
[ ] Yes
[X] No
```

## Type of change

```
[ ] Bugfix
[X] Feature
[ ] Other... Please describe:
```

## Code quality checklist


pamelafox:
Note: this would probably require changes to ai-rag-chat-evaluator

mattgotteiner (approved):

pamelafox:
Related change to ai-rag-chat-evaluator: https://github.com/Azure-Samples/ai-rag-chat-evaluator/pull/92

PR #1687, opened 2024-06-05
Switch to PSRule for security testing

tonybaloney:
This PR replaces MSDO/template analyser with the PSRule GitHub Action.

1. template analyser recurses over all the bicep files in the folder and raises lots of false positives
2. PSRule is more configurable
3. We can test a specific set of parameters which works with AZD
4. template analyser doesn't work with conditional modules

Downsides:
1. The errors don't match to the source address of the bicep in the SARIF report 

The `main.test.bicep` file is the set of parameters that we want to test, in our case which is the private network, isolated network environment.


infra/main.test.bicep
```diff
@@ -0,0 +1,40 @@
+// This file is for doing static analysis and contains sensible defaults
+// for the template analyser to minimise false-positives and provide the best results.
```
pamelafox, line 2:
```suggestion
// for PSRule to minimise false-positives and provide the best results.
```

ps-rule.yaml
```diff
@@ -0,0 +1,3 @@
+# YAML: Set the AZURE_BICEP_FILE_EXPANSION configuration option to enable expansion
+configuration:
+  AZURE_BICEP_FILE_EXPANSION: true
```
pamelafox, line 3:
```suggestion
  AZURE_BICEP_FILE_EXPANSION: true

```

pamelafox (approved):
Neat, it'd be nice to have something in the docs, maybe productionizing.md, that told customers how to set up main.test.bicep for their own configuration.

PR #1746, opened 2024-06-19
Add support for specifying an Azure OpenAI Key

pamelafox:
## Purpose

This PR adds support for specifying an API key for Azure OpenAI. We typically do not support key-based access in this repository, but we would like to be able to use this repo with the Azure OpenAI Proxy (a service that advocates use in workshops/hackathons due to the difficulty of getting immediate Azure OpenAI access), which means we need a way of specifying a key.

I also changed the Bicep so that if you are specifying a custom URL for your Azure endpoint, then we do *not* create any Azure OpenAI instance at all. I'm not sure if that's always true, i.e. if developers using the load-balanced OpenAI instances want to skip creating one in this infra. Maybe @andredewes can comment.

Alternatively, I can add an explicit CREATE_AZURE_OPENAI env variable.

```
[ ] Yes
[X] No
```

```
[ ] Yes
[X] No
```

## Type of change

```
[ ] Bugfix
[X] Feature
[ ] Other... Please describe:
```

## Code quality checklist

- [x] I added tests that prove my fix is effective or that my feature works- current code is untested in this area, it seems

### DIFF
--- a/infra/main.bicep
+++ b/infra/main.bicep
@@ -43,9 +43,11 @@ param appServiceSkuName string // Set in main.parameters.json
 @allowed([ 'azure', 'openai', 'azure_custom' ])
 param openAiHost string // Set in main.parameters.json
 param isAzureOpenAiHost bool = startsWith(openAiHost, 'azure')
+param deployAzureOpenAi bool = openAiHost == 'azure'
 param azureOpenAiCustomUrl string = ''
 param azureOpenAiApiVersion string = ''
-
+@secure()
+param azureOpenAiApiKey string = ''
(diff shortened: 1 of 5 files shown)


infra/main.bicep
```diff
@@ -66,6 +68,7 @@ param openAiResourceGroupLocation string
 
 param openAiSkuName string = 'S0'
 
+@secure()
```
pamelafox, line 71: The secure() should resolve some security alerts, along with the removal of the output.

mattgotteiner (approved):
LGTM

Issue #1639, opened 2024-05-21
Search Results are Empty after Browsersession is inactive for a while


Recco23:
> Please provide us with the following information:
> ---------------------------------------------------------------

### This issue is for a: (mark with an `x`)
```
- [X] bug report -> please search issues before submitting
- [ ] feature request
- [ ] documentation issue or request
- [ ] regression (a behavior that used to work and stopped in a new release)
```

### Minimal steps to reproduce
> Use the app with login and document-level access control enabled. Interact with the app via chat. Then, leave the browser session open but idle for some time (3 hours should be sufficient, but a shorter period might also work).

### Any log messages given by the failure
> No

### Expected/desired behavior
> Browsersession should deliver Search Results as expected even after the Session idles for a longer time or a Message, that for example a Relog is needed must be displayed.

### OS and Version?
> Windows 7, 8 or 10. Linux (which distribution). macOS (Yosemite? El Capitan? Sierra?)
tested with Windows 10 and Windows 11 in Chrome, Edge and Firefox

### azd version?
> run `azd version` and copy paste here.
azd version 1.9.2
### Versions
>Latest from Main branch

### Mention any other details that might be useful
The problem occurs when you chat with the web app and then leave it idle for an extended period (for example, 3 hours). When you start chatting again, the search results are always empty, even if the answers are present in the documents and would normally be retrieved.
<img width="445" alt="image" src="https://github.com/Azure-Samples/azure-search-openai-demo/assets/5775591/69e57982-f920-433b-b859-2e04ad7298e3">
<img width="470" alt="image" src="https://github.com/Azure-Samples/azure-search-openai-demo/assets/5775591/a0df94db-10b9-4904-ad06-03642d381230">
After refreshing the browser session (by pressing F5, clicking the refresh button), the system works as intended again. Opening the chatbot in a new browser tab even makes the old, non-functional tab work again.
<img width="466" alt="image" src="https://github.com/Azure-Samples/azure-search-openai-demo/assets/5775591/14a46baf-10f6-4e87-995e-c2ca7642a74c">
<img width="476" alt="image" src="https://github.com/Azure-Samples/azure-search-openai-demo/assets/5775591/c982a657-29a3-42fa-a182-7d9afb6344ef">

> ---------------------------------------------------------------
> Thanks! We'll be in touch soon.

pamelafox:
cc @mattgotteiner 

[...]

PR #1778, opened 2024-07-01
Fix login state management

mattgotteiner:
## Purpose

* Issue #1639 was caused by a stale login token not being refreshed after the expiry period
* The login code only fetched the login token once when the page was loaded and did not properly reflect login state changes if the user logged in and out unless the page was refreshed.
* Update the React state management code to use a context to pass login state to all the components that need this to render properly.

```
[ ] Yes
[X] No
```

```
[ ] Yes
[X] No
```

## Type of change

```
[X] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## Code quality checklist

### DIFF
--- a/app/frontend/src/api/api.ts
+++ b/app/frontend/src/api/api.ts
@@ -1,11 +1,11 @@
 const BACKEND_URI = "";
 
 import { ChatAppResponse, ChatAppResponseOrError, ChatAppRequest, Config, SimpleAPIResponse } from "./models";
-import { useLogin, appServicesToken } from "../authConfig";
+import { useLogin, getToken, isUsingAppServicesLogin } from "../authConfig";
 
-export function getHeaders(idToken: string | undefined): Record<string, string> {
+export async function getHeaders(idToken: string | undefined): Promise<Record<string, string>> {
     // If using login and not using app services, add the id token of the logged in account as the authorization
-    if (useLogin && appServicesToken == null) {
+    if (useLogin && !isUsingAppServicesLogin) {
         if (idToken) {
             return { Authorization: `Bearer ${idToken}` };
         }
@@ -23,9 +23,10 @@ export async function configApi(): Promise<Config> {
 }
 
 export async function askApi(request: ChatAppRequest, idToken: string | undefined): Promise<ChatAppResponse> {
+    const headers = await getHeaders(idToken);
     const response = await fetch(`${BACKEND_URI}/ask`, {
         method: "POST",
-        headers: { ...getHeaders(idToken), "Content-Type": "application/json" },
+        headers: { ...headers, "Content-Type": "application/json" },
         body: JSON.stringify(request)
     });
 
@@ -42,9 +43,10 @@ export async function chatApi(request: ChatAppRequest, shouldStream: boolean, id
     if (shouldStream) {
         url += "/stream";
     }
+    const headers = await getHeaders(idToken);
     return await fetch(url, {
         method: "POST",
-        headers: { ...getHeaders(idToken), "Content-Type": "application/json" },
+        headers: { ...headers, "Content-Type": "application/json" },
         body: JSON.stringify(request)
     });
 }
@@ -80,7 +82,7 @@ export function getCitationFilePath(citation: string): string {
 export async function uploadFileApi(request: FormData, idToken: string): Promise<SimpleAPIResponse> {
     const response = await fetch("/upload", {
         method: "POST",
-        headers: getHeaders(idToken),
+        headers: await getHeaders(idToken),
         body: request
     });
 
@@ -93,9 +95,10 @@ export async function uploadFileApi(request: FormData, idToken: string): Promise
 }
 
 export async function deleteUploadedFileApi(filename: string, idToken: string): Promise<SimpleAPIResponse> {
+    const headers = await getHeaders(idToken);
     const response = await fetch("/delete_uploaded", {
         method: "POST",
-        headers: { ...getHeaders(idToken), "Content-Type": "application/json" },
+        headers: { ...headers, "Content-Type": "application/json" },
         body: JSON.stringify({ filename })
     });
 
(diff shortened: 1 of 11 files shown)


mattgotteiner:
Manual deploy w/ app services login test

app/frontend/src/api/api.ts
```diff
@@ -23,9 +23,10 @@ export async function configApi(): Promise<Config> {
 }
 
 export async function askApi(request: ChatAppRequest, idToken: string | undefined): Promise<ChatAppResponse> {
+    const headers = await getHeaders(idToken);
     const response = await fetch(`${BACKEND_URI}/ask`, {
         method: "POST",
-        headers: { ...getHeaders(idToken), "Content-Type": "application/json" },
+        headers: { ...headers, "Content-Type": "application/json" },
```
pamelafox, line 29: Was this split into 2 lines for debuggability, or because we can't use await with the ...? Just curious.

mattgotteiner: I assumed we couldn't use the ... with await, but I can put it back to ...(await getHeaders(idToken)) if necessary

app/frontend/src/authConfig.ts
```diff
@@ -156,3 +157,31 @@ export const getToken = (client: IPublicClientApplication): Promise<string | und
             return undefined;
         });
 };
+
+export const getUsername = async (client: IPublicClientApplication): Promise<string | null> => {
```
pamelafox, line 161: I wonder if itd be helpful to add docstrings like "Gets the username from either MSAL.JS-based authentication or App Service built-in auth" to help developers understand why we have all these functions. That could also be a top-level comment (with links to relevant doc overviews), but for intellisense, it can be nice to have function docstrings too.

app/frontend/src/components/LoginButton/LoginButton.tsx
```diff
@@ -2,13 +2,23 @@ import { DefaultButton } from "@fluentui/react";
...
-    const isLoggedIn = (activeAccount || appServicesToken) != null;
+    const [username, setUsername] = useState("");
+
+    useEffect(() => {
+        const fetchUsername = async () => {
+            setUsername((await getUsername(instance)) ?? "");
+        };
+
+        fetchUsername();
+    }, []);
```
pamelafox, line 21: As I understand it, useEffect happens anytime the state changes. So this would trigger whenever username state changed? I guess since username is the only state, then [] is effectively the same as ["username"] for the triggers?

app/frontend/src/authConfig.ts
```diff
@@ -133,14 +133,15 @@ export const appServicesLogout = () => {
 
 // Determine if the user is logged in
 // The user may have logged in either using the app services login or the on-page login
-export const isLoggedIn = (client: IPublicClientApplication | undefined): boolean => {
-    return client?.getActiveAccount() != null || appServicesToken != null;
+export const checkLoggedIn = async (client: IPublicClientApplication | undefined): Promise<boolean> => {
+    return client?.getActiveAccount() != null || (await getAppServicesToken()) != null;
```
pamelafox, line 137: I wonder if it would help developers debugging if this was a full if/else with a console.log for "Logged in with Built-in Auth on App Service", that sort of thing. I've started adding those sorts of logs in similar scenarios, i.e. whenever an env var affects an if/else, to help developers debug their setups.

app/frontend/src/components/LoginButton/LoginButton.tsx
mattgotteiner: calling useEffect like this only invokes it once in the constructor as far as I can tell. It has no dependencies so it only executes once. The reason for this one is that getUsername may need a fresh token and that's an async operation

app/frontend/src/pages/chat/Chat.tsx
```diff
@@ -136,6 +137,7 @@ const Chat = () => {
     };
 
     const client = useLogin ? useMsal().instance : undefined;
+    const { loggedIn } = useLogin ? useContext(LoginContext) : { loggedIn: false };
```
pamelafox, line 140: Would it make sense for LoginContext to have a function to return the { loggedin: false}? I notice you do that in a few files, and that it matches what an actual LoginContext has, so it might make sense for the dummy context to be in that file as well. Just an idea.

app/frontend/src/loginContext.tsx
```diff
@@ -0,0 +1,5 @@
+import { createContext } from "react";
```
pamelafox, line 1: Can you add a docstring to this file to explain how to use it? I'm actually new to React contexts (in my days, we didnt have contexts..) so an explanation would be helpful.

pamelafox (approved):
I have no blocking requests, all of my comments are for my edification / code clarity. Thanks for fixing!

app/frontend/src/pages/chat/Chat.tsx
mattgotteiner: Good idea!

app/frontend/src/api/api.ts
pamelafox: I don't mind either way, I think splitting for debuggability can be nice anyway.

mattgotteiner: OK - I'll leave this as-is

app/frontend/src/authConfig.ts
mattgotteiner: Did not add the console.log but took your advice on ifs

app/frontend/src/authConfig.ts
```diff
@@ -92,29 +93,65 @@ export const getRedirectUri = () => {
     return window.location.origin + authSetup.msalConfig.auth.redirectUri;
 };
 
-// Get an access token if a user logged in using app services authentication
-// Returns null if the app doesn't support app services authentication
+// Cache the app services token if it's available
+declare global {
+    var cachedAppServicesToken: AppServicesToken | null;
+}
+globalThis.cachedAppServicesToken = null;
```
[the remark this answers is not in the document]
pamelafox: Ah its a global. Huh. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis

pamelafox (approved):

pamelafox:
prettier failed!

app/frontend/src/authConfig.ts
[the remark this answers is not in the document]
mattgotteiner: I can add a comment here!

Issue #1780, opened 2024-07-02
CORS blocking redirect on app services logout


mattgotteiner:
> Please provide us with the following information:
> ---------------------------------------------------------------

### This issue is for a: (mark with an `x`)
```
- [X] bug report -> please search issues before submitting
- [ ] feature request
- [ ] documentation issue or request
- [ ] regression (a behavior that used to work and stopped in a new release)
```

### Minimal steps to reproduce
* Deploy app on azure with app services authentication enabled
* Observe that when you log out you hit a CORS error because you cannot be redirected back to the login page automatically. You stay on a blank screen

### Any log messages given by the failure
Redirect blocked by CORS policy

### Expected/desired behavior
The redirect is allowed

### Mention any other details that might be useful

> ---------------------------------------------------------------
> Thanks! We'll be in touch soon.

pelonchasva:
any updates on this issue? I just faced it while enabling authentication for the app, login works fine, but if I logout I end up in a white screen, developer tools on chrome says that it is a CORS issue.

PR #1782, opened 2024-07-03
Fix CORS setup

mattgotteiner:
## Purpose

* Issue #1780 indicated something was wrong with the CORS configuration
* ALLOWED_ORIGIN was not set properly so the app backend couldn't respond to cross-origin requests that are allowed
* Fix setting of ALLOWED_ORIGIN in bicep

```
[ ] Yes
[X] No
```

```
[ ] Yes
[X] No
```

## Type of change

```
[X] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

### DIFF
--- a/app/backend/app.py
+++ b/app/backend/app.py
@@ -677,6 +677,8 @@ def create_app():
     logging.basicConfig(level=os.getenv("APP_LOG_LEVEL", default_level))
 
     if allowed_origin := os.getenv("ALLOWED_ORIGIN"):
-        app.logger.info("CORS enabled for %s", allowed_origin)
-        cors(app, allow_origin=allowed_origin, allow_methods=["GET", "POST"])
+        allowed_origin = allowed_origin.split(";")
+        if len(allowed_origin) > 0:
+            app.logger.info("CORS enabled for %s", allowed_origin)
+            cors(app, allow_origin=allowed_origin, allow_methods=["GET", "POST"])
     return app
--- a/infra/main.bicep
+++ b/infra/main.bicep
@@ -191,6 +191,16 @@ var tags = { 'azd-env-name': environmentName }
 var tenantIdForAuth = !empty(authTenantId) ? authTenantId : tenantId
 var authenticationIssuerUri = '${environment().authentication.loginEndpoint}${tenantIdForAuth}/v2.0'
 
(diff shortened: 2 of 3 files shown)


pamelafox:
mypy eek

infra/main.bicep
```diff
@@ -283,7 +293,7 @@ module backend 'core/host/appservice.bicep' = {
     managedIdentity: true
     virtualNetworkSubnetId: isolation.outputs.appSubnetId
     publicNetworkAccess: publicNetworkAccess
-    allowedOrigins: [ allowedOrigin ]
+    allowedOrigins: allowedOrigins
```
pamelafox, line 296: Does this always end up allowing origins even if they don't enable user auth? I dont see a conditional that makes it empty in that case. But maybe my eyes have glazed over.

pamelafox: Or maybe we always allowed them anyway?

mattgotteiner: allowedOrigin should be empty if they don't enable user auth. I will double check this, good catch

mattgotteiner: So this is an interesting case
1. We were never enabling the portal origins before due to a bug
2. So unless they explicitly specified an allowed origin, we never enabled cors

My thought is that we want to only add in the portal / login origins if auth is enabled, otherwise just pick the origins they added. I'll make this adjustment

app/backend/app.py
```diff
@@ -677,6 +677,8 @@ def create_app():
     logging.basicConfig(level=os.getenv("APP_LOG_LEVEL", default_level))
 
     if allowed_origin := os.getenv("ALLOWED_ORIGIN"):
-        app.logger.info("CORS enabled for %s", allowed_origin)
-        cors(app, allow_origin=allowed_origin, allow_methods=["GET", "POST"])
+        allowed_origin = allowed_origin.split(";")
+        if len(allowed_origin) > 0:
+            app.logger.info("CORS enabled for %s", allowed_origin)
+            cors(app, allow_origin=allowed_origin, allow_methods=["GET", "POST"])
```
mattgotteiner, line 683: We do need to allow credentials

mattgotteiner:
Pulled into #2205

PR #1817, opened 2024-07-15
Use AVM for log analytics and application insights.

tonybaloney:
Provisions the same resources but using the AVM resources. Replaces the 2 custom bicep files that existed for provisioning the resources with the custom private network flag.


infra/core/monitor/monitoring.bicep
```diff
@@ -7,33 +7,43 @@ param tags object = {}
...
-module logAnalytics 'loganalytics.bicep' = {
+module logAnalytics 'br/public:avm/res/operational-insights/workspace:0.4.0' = {
   name: 'loganalytics'
   params: {
     name: logAnalyticsName
     location: location
     tags: tags
+    skuName: 'PerGB2018'
+    dataRetention: 30
     publicNetworkAccessForIngestion: publicNetworkAccess
```
pamelafox, line 18: I'm surprised that the avm has a parameter named publicNetworkAccessForIngestion ? How is that working?

tonybaloney: It's a parameter of the log analytics workspace and the application insights account anyway https://github.com/Azure/bicep-registry-modules/blob/main/avm/res/operational-insights/workspace/main.bicep#L192-L193

pamelafox: Ohh, I thought it was one we made up ("ingestion" threw me off). Got it, thanks

pamelafox:
Which envs have you tested with, both public and vnet? I want to figure out what still needs testing.

tonybaloney:
After doing a diff on the deployment, the only thing this changes is `forceCmkForQuery` to true (previously unspecified) for the log analytics workspace which improves security and should have no impact on most users.

pamelafox (approved):

pamelafox:
Okay, thanks for diff'ing! @mattgotteiner This will be the first move to AVM but the hope is to move all modules to AVM, to reduce Bicep modules in our code and inherit best practices by default from AVM.

PR #1826, opened 2024-07-17
Replace storage accounts with AVM modules

tonybaloney:
## Purpose

Replaces AZD Core Storage account module with the AVM one.


tonybaloney:
```default
Deployment Error Details:
AccountPropertyCannotBeUpdated: The property 'isHnsEnabled' was specified in the input, but it cannot be updated as it is read-only. For more information, see - https://aka.ms/storageaccountupdate
```

I think this is because AZD core hard-coded the HNS value to false whereas AVM uses `null` by default.

tonybaloney:
Related to https://github.com/Azure/azure-rest-api-specs/issues/18441

tonybaloney:
AVM tracking issue https://github.com/Azure/bicep-registry-modules/issues/1522

We can't make this migration until this is fixed. Anyone who has a storage account deployed already will get a failure.

pamelafox:
https://github.com/Azure/bicep-registry-modules/pull/3317 seems fixed now, so can this PR be revisited?

PR #1861, opened 2024-07-24
Add NSG Support

mattgotteiner:
## Purpose

* Add support for Azure Network Security Groups

```
[ ] Yes
[X] No
```

```
[ ] Yes
[X] No
```

## Type of change

```
[ ] Bugfix
[X] Feature
[ ] Other... Please describe:
```

### DIFF
--- a/infra/network-isolation.bicep
+++ b/infra/network-isolation.bicep
@@ -3,6 +3,9 @@ metadata description = 'Sets up private networking for all resources, using VNet
 @description('The name of the VNet to create')
 param vnetName string
 
+@description('The name of the NSG to create')
+param nsgName string
+
 @description('The location to create the VNet and private endpoints')
 param location string = resourceGroup().location
 
@@ -18,6 +21,15 @@ resource appServicePlan 'Microsoft.Web/serverfarms@2022-03-01' existing = {
   name: appServicePlanName
 }
 
+module nsg './core/networking/nsg.bicep' = if (usePrivateEndpoint) {
+  name: 'nsg'
+  params: {
+    name: nsgName
+    location: location
+    tags: tags
+  }
+}
+
 module vnet './core/networking/vnet.bicep' = if (usePrivateEndpoint) {
   name: 'vnet'
   params: {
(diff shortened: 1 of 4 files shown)


mattgotteiner:
Manual deployment check

infra/network-isolation.bicep
```diff
@@ -47,6 +54,9 @@ module vnet './core/networking/vnet.bicep' = if (usePrivateEndpoint) {
           addressPrefix: '10.0.3.0/24'
           privateEndpointNetworkPolicies: 'Enabled'
           privateLinkServiceNetworkPolicies: 'Enabled'
+          networkSecurityGroup: {
+            id: usePrivateEndpoint ? nsg.outputs.id : ''
```
pamelafox, line 58: Hm, I think Sarah suggested to me that there should be a different NSG for each subnet. Did you consider that approach?

infra/network-isolation.bicep
```diff
@@ -18,6 +21,24 @@ resource appServicePlan 'Microsoft.Web/serverfarms@2022-03-01' existing = {
   name: appServicePlanName
 }
 
+module backendNsg './core/networking/nsg.bicep' = if (usePrivateEndpoint) {
```
pamelafox, line 24: Can we use AVM's module here?
https://github.com/Azure/bicep-registry-modules/tree/main/avm/res/network/network-security-group

pamelafox: Example- https://github.com/Azure/bicep-registry-modules/tree/main/avm/res/network/network-security-group#example-3-waf-aligned

pamelafox: You'd have to define the securityrules as a variable in this file, in that case.

pamelafox: But we can also defer the AVM til later, if you prefer, so that we can get this in sooner.

mattgotteiner: Are there AVM docs? Didn't realize we have them now

mattgotteiner:
https://github.com/microsoft/Oryx/issues/1585

There doesn't seem to be a way to allow connectivity to Oryx through an NSG without hardcoding IP addresses that may change based on the region you are deployed in

PR #1872, opened 2024-07-29
Feat: Template Introduction

alice-wondering:
## Purpose

Introduce templates to standardise and support switching between different models at runtime. Added a templates directory to store structured templates for ChatGPT 3.5 and MistralAI-Instruct (from Hugging Face). 

Each template consists of the configuration for the model parameters written in yaml, and a Jinja template for prompting models. The backend for Ask and Chat tabs have been modified to use the Prompty objects available in the Promptflow package to load the templates. This feature addresses the issue of incompatible prompting for OpenAI services and the Hugging Face inference API through universalized, structured templates.

```
[ ] Yes
[x] No
```

```
[ ] Yes
[x] No
```

## Type of change

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

## Code quality checklist


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -8,32 +8,26 @@
 When developers merge from main and run the server, azd up, or azd deploy, will this produce an error?
 If you're not sure, try it out on an old environment.
 
-```
-[ ] Yes
-[ ] No
-```
+- [ ] Yes
```
pamelafox, line 11: These are intentionally in blocks, since they're not checklists.

.github/workflows/python-test.yaml
```diff
@@ -58,7 +58,7 @@ jobs:
           run: black . --check --verbose
         - name: Run Python tests
           if: runner.os != 'Windows'
-          run: python3 -m pytest -s -vv --cov --cov-fail-under=86
+          run: python3 -m pytest -s -vv --cov --cov-fail-under=80
```
pamelafox, line 61: We shouldn't be reducing the coverage threshold

pamelafox:
Thanks for the PR, I have some high level questions since this is a large change:

1) Are the Prompty files actually used? I don't see a call to them, and I don't see the Prompty runtime included

2) How is token-counting being handled for the Huggingface models? We are using a package (that I maintain) for OpenAI token counting, but I haven't got support for Huggingface in there

3) Generally, we'd also like support for Azure Marketplace models, like Phi3, Mistral from the college, and those can be inferenced using the Azure Inference SDK. However, I see you're using the Huggingface SDK. Any reason not to use the Azure Inference SDK?

pamelafox:
I see you've closed it, is that because it's a WIP or not intended to be merged in the main repo?

phoevos:
Hey @pamelafox, thanks for your swift reply but the PR was accidentally opened against the main repo (hence the totally sound engineering-wise changes made strictly for our convenience like reducing the coverage requirement :smile: ). Appreciate you taking the time to check it out though! Quickly going through your questions:

> Are the Prompty files actually used? I don't see a call to them, and I don't see the Prompty runtime included

Currently, they're only meant to be used to handle the template rendering logic. The Prompty runtime (and Promptflow) only supports Azure-hosted models atm, therefore we can't use their executors to run the flows. The hope is that support for arbitrary model targets will be added in the future and that's why we introduce Prompty over rendering the templates directly with Jinja.

> How is token-counting being handled for the Huggingface models?

That's a great question, I'll raise it internally, but my understanding is that it's not.

> Generally, we'd also like support for Azure Marketplace models, like Phi3, Mistral from the college, and those can be inferenced using the Azure Inference SDK. However, I see you're using the Huggingface SDK. Any reason not to use the Azure Inference SDK?

I'd also like to support that and it'd be lovely indeed, but this is a school project and that was our project statement! Our hope was that we would start with Hugging Face, allowing for easy (and inexpensive) experimentation, and add support for Azure-hosted models down the line. Due to subscription and budget constraints, deploying models on Azure as serverless or online endpoints respectively, is not an option for us so it seems unlikely that we'll be moving forward with these. Right now we're focusing on providing a straightforward way for people to try different LLMs, exposing a list of supported models and adding templates to handle the configuration and particularities of talking to each one of them.

pamelafox:
Cool! Well, it's fun to see how you adapted the repo to use different models. We're still figuring out the approach to work with a bigger range of models, as it's definitely a big ask, but there's multiple ways to do it. I may make a PR soon to move to Prompty with the lightweight Prompty runtime, and use similar files to what you used here.

PR #1874, opened 2024-07-31
Fix logging

pamelafox:
## Purpose

Makes logs work in prod! See https://github.com/Kludex/uvicorn-worker/pull/18 for details

Screenshot 2024-07-31 at 7 32 32 AM

```
[ ] Yes
[X] No
```

```
[ ] Yes
[Z]X ] No
```

## Type of change

```
[X] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## Code quality checklist


mattgotteiner (approved):

TaylorN15:
@pamelafox - should the logging functions in other areas be updated too? e.g. [here](https://github.com/Azure-Samples/azure-search-openai-demo/blob/7b55ad2b4cb4911141a69963a230ee79fd2f2b41/app/backend/error.py#L24) and [here](https://github.com/Azure-Samples/azure-search-openai-demo/blob/7b55ad2b4cb4911141a69963a230ee79fd2f2b41/app/backend/decorators.py#L29)

pamelafox:
@TaylorN15 Good question - I think I may have attempted to change those to use the app logger, but gotten an error due to them happening outside of the app context (like during the streamed response for the error handling function). Fortunately since they're error-level, we *should* see them show up in the logs regardless.

A better fix might be to drop gunicorn and only use uvicorn. I've done that for a few other containerized apps, but am hoping that App Service will add an env variable like WEB_CONCURRENCY that makes it easy to change the worker count based on the number of cores.

TaylorN15:
Ah yes, you're right. You'll get an error as it's outside of the app context. 

Honestly, I'm considering moving to FastAPI as I've been having some issues with our prod deployment and it appears to be related to gunicorn, but it's unclear.

pamelafox:
We can also run Quart apps with uvicorn. For either Quart or FastAPI, we still need to set the number of workers somehow to take advantage of the host machine's cores, but we could hardcode that for the default sku, or just write a clever shell script.

We could also just port this app to FastAPI (I know a few devs have!), I just find that some things are a bit harder in FastAPI than in Quart (static files, middleware, etc), per my blog at https://blog.pamelafox.org/2024/07/should-you-use-quart-or-fastapi-for-ai.html

PR #1879, opened 2024-08-01
Better handling of auth errors

pamelafox:
## Purpose

Right now, if the getToken function receives an error from MSAL, it'll get logged in the console, the API request will be made, and the user will see a JS error in the answer box.

This PR changes it so that you see an alert, and then you see a 403 in the answer box.

Screenshot 2024-08-01 at 9 33 01 AM
Screenshot 2024-08-01 at 9 46 48 AM

```
[ ] Yes
[X] No
```

```
[ ] Yes
[X] No
```

## Type of change

```
[X] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## Code quality checklist


app/frontend/src/api/api.ts
```diff
@@ -30,9 +30,12 @@ export async function askApi(request: ChatAppRequest, idToken: string | undefine
         body: JSON.stringify(request)
     });
 
-    const parsedResponse: ChatAppResponseOrError = await response.json();
     if (response.status > 299 || !response.ok) {
-        throw Error(parsedResponse.error || "Unknown error");
+        throw Error(`Request failed with status ${response.status}`);
```
pamelafox, line 34: We don't always send back JSON, so it's better to throw earlier if the status code isn't OK.

app/frontend/src/authConfig.ts
```diff
@@ -207,6 +207,7 @@ export const getToken = async (client: IPublicClientApplication): Promise<string
         .then(r => r.accessToken)
         .catch(error => {
             console.log(error);
+            window.alert(error);
```
pamelafox, line 210: I don't know if there are situations where this will be annoying? Do we ever have errors that we're okay with ignoring?

app/frontend/src/pages/chat/Chat.tsx
```diff
@@ -183,16 +183,20 @@ const Chat = () => {
             };
 
             const response = await chatApi(request, shouldStream, token);
+
             if (!response.body) {
                 throw Error("No response body");
             }
+            if (response.status > 299 || !response.ok) {
```
pamelafox, line 189: Same change here, throw a non-200 error earlier before looking for error in the JSON

app/frontend/src/authConfig.ts
mattgotteiner: i would remove this personally

mattgotteiner (approved):

PR #1880, opened 2024-08-01
Rename Azure OpenAI key environment variable

pamelafox:
## Purpose

Since adding AZURE_OPENAI_API_KEY as an optional env variable several months ago, many developers have had errors when running and deploying the app, due to that variable existing from other tutorials/templates. This PR changes the variable to a much more unique name.
Fortunately, I never documented it in this repo, since that change was to support our use of the Azure OpenAI Proxy in workshops, so I do not believe many developers are using it for production. I will note in release notes as a backwards compatible change.

```
[X] Yes - for any developer who'd been relying on it. Should be minimal given its 2 months old and undocumented.
[ ] No
```

```
[ ] Yes
[X] No - But I do need to update https://github.com/Azure-Samples/ai-app-templates-workshop once merged
```

## Type of change

```
[X] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## Code quality checklist

### DIFF
--- a/app/backend/app.py
+++ b/app/backend/app.py
@@ -541,8 +541,8 @@ async def setup_clients():
             if not AZURE_OPENAI_SERVICE:
                 raise ValueError("AZURE_OPENAI_SERVICE must be set when OPENAI_HOST is azure")
             endpoint = f"https://{AZURE_OPENAI_SERVICE}.openai.azure.com"
-        if api_key := os.getenv("AZURE_OPENAI_API_KEY"):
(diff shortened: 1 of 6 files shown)


mattgotteiner (approved):
I think we just need a doc update, otherwise this all looks good!

pamelafox:
Our docs don't describe this option, only the docs at https://github.com/Azure-Samples/ai-app-templates-workshop describe it. I could add this option to the docs, but generally, we discourage key usage, what do you think?

PR #1894, opened 2024-08-07
On Demand Azure Speech Generation

### LINKED ISSUE #1892: Feature: Generate text to speech on demand
It seems that TTS is run for all answers automatically, and users are given the option to play the stream. Most of my users wouldn't use the feature, so we would be unnecessarily calling the Speech Service API in many cases. I'd like to be able to call this on demand only. I.e. a user clicks the play button on a message, and _then_ it calls the API to synthesize the speech.

john0isaac:
## Purpose

fix #1892 

```
[ ] Yes
[x] No
```

```
[ ] Yes
[x] No
```

## Type of change

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

## Code quality checklist


john0isaac:
TODO
- [ ] Cache the Generated Audio after the first time it's generated to not incur more charges.
- [ ] Handel cases where the Azure Speech API request function returns `null`.

john0isaac:
@TaylorN15, can you look at this and let me know if it's working as you expect?

@pamelafox done, it's not the prettiest implementation but it works.
I couldn't figure out a way to update the URLs and **Caching** them on click other than passing all of the URLs, index, and setter to the AzureSpeechUrls function to check then make a request and use setter to update the value.

This is the initial implementation for on-demand speech request https://github.com/Azure-Samples/azure-search-openai-demo/pull/1894/commits/c08ce23765fe6ec06ec72631f0ad155705e4b7b6 but every time you click on it, it will make a new request to Azure Speech Service.

app/frontend/src/components/Answer/SpeechOutputAzure.tsx
```diff
@@ -1,44 +1,64 @@
 import { useState } from "react";
 
 import { IconButton } from "@fluentui/react";
+import { getSpeechApi } from "../../api";
+import { update } from "@react-spring/web";
```
john0isaac, line 5:
```suggestion
```

TaylorN15:
Why is `speechUrls` an array? And why pass it down from `Chat.tsx`? Is that because you were just trying not change too much code?

TaylorN15:
Also it seems you can play multiple concurrently, this doesn't happen with the web speech API as it seems to handle it automatically. I think it would be best to have some sort of state to manage if audio is playing on any answer, and pause/stop on others if another one is played.

john0isaac:
> Why is `speechUrls` an array? And why pass it down from `Chat.tsx`? Is that because you were just trying not change too much code?

nope like I said on the issue the AzureSpeechOutput is a component it can't keep track of it's state anything done inside it is not saved.
To save it's state you need to keep track of it in the Parent Page like we keep an array of answers to get the Thought Process etc.


> Also it seems you can play multiple concurrently, this doesn't happen with the web speech API as it seems to handle it automatically. I think it would be best to have some sort of state to manage if audio is playing on any answer, and pause/stop on others if another one is played.

This is out of the scope of this PR it was never being handled correctly for Azure Speech Service the web speech Api is not handling it automatically I implemented the logic to handle this to not play multiple audios concurrently.
But happy to fix that too.

TaylorN15:
I have gotten this working in my implementation, but quite a few changes from what you have so it's hard to compare.

```
// Chat.tsx
    const [activeAudioAnswer, setActiveAudioAnswer] = useState<number | null>(null);
    const [isAudioPlaying, setIsAudioPlaying] = useState<boolean>(false);
    const [cachedAudioUrls, setCachedAudioUrls] = useState<{ [key: number]: string }>({});
    const ttsAudio = useRef(new Audio()).current;

    const onToggleAudioPlay = (index: number, speechUrl: string) => {
        if (isAudioPlaying && activeAudioAnswer === index) {
            stopAudio();  // Stop if it's currently playing the same answer
        } else {
            if (isAudioPlaying) {
                stopAudio();  // Stop any currently playing audio
            }
            playAudio(index, speechUrl);
        }
    };

    const playAudio = (index: number, speechUrl: string) => {
        ttsAudio.src = speechUrl;
        ttsAudio.play().then(() => {
            setActiveAudioAnswer(index);
            setIsAudioPlaying(true);
        }).catch((error) => {
            handleError(error, "Error playing audio:");
        });

        ttsAudio.onended = () => {
            setIsAudioPlaying(false);
            setActiveAudioAnswer(null);
        };
    };

    const stopAudio = () => {
        ttsAudio.pause();
        ttsAudio.currentTime = 0;
        setActiveAudioAnswer(null);
        setIsAudioPlaying(false);
    };

    const cacheAudioUrl = (index: number, url: string) => {
        setCachedAudioUrls(prevUrls => ({
            ...prevUrls,
            [index]: url
        }));
    };
...
<Answer
    index={index}
    isAudioPlaying={index === activeAudioAnswer && isAudioPlaying}
    onToggleAudioPlay={onToggleAudioPlay}
    cachedAudioUrl={cachedAudioUrls[index] || null}
    onCacheAudioUrl={cacheAudioUrl}
...
/>
```

```
// Answer.tsx
const [isAudioLoading, setIsAudioLoading] = useState<boolean>(false);

    const toggleAudioPlay = async () => {
        if (cachedAudioUrl) {
            onToggleAudioPlay(index, cachedAudioUrl);
        } else {
            setIsAudioLoading(true);
            const plainText = removeHtmlTags(sanitizedAnswerHtml);
            const token = client ? await getToken(client) : undefined;
            const speechUrl = await getSpeechApi(token, plainText);

            if (!speechUrl) {
                handleError("Error", "An error occurred while generating speech");
                setIsAudioLoading(false);
                return;
            }

            onCacheAudioUrl(index, speechUrl);
            onToggleAudioPlay(index, speechUrl);
            setIsAudioLoading(false);
        }
    };
```

Then I pass the state of the audio loading/playing into the buttons so that it shows a play/stop depending on state.

john0isaac:
@TaylorN15 welp, I didn't want to do this I found a workaround.
I implemented it differently to maintain the modularity of the implementation.
Thanks for your code though it helped, can you give it a go now and let me know if there is anything else.

TaylorN15:
> @TaylorN15 welp, I didn't want to do this I found a workaround.
> 
> I implemented it differently to maintain the modularity of the implementation.
> 
> Thanks for your code though it helped, can you give it a go now and let me know if there is anything else.

Looks good. I don't have time to deploy and test, like I said my code is quite different.

app/frontend/src/pages/ask/Ask.tsx
```diff
@@ -48,13 +48,15 @@ export function Component(): JSX.Element {
...
+    const audio = useRef(new Audio()).current;
+    const [isPlaying, setIsPlaying] = useState(false);
 
     const lastQuestionRef = useRef<string>("");
 
     const [isLoading, setIsLoading] = useState<boolean>(false);
     const [error, setError] = useState<unknown>();
     const [answer, setAnswer] = useState<ChatAppResponse>();
-    const [speechUrl, setSpeechUrl] = useState<string | null>(null);
+    const [speechUrl, setSpeechUrl] = useState<(string | null)[]>([]);
```
pamelafox, line 59: Since this is a list, can you change to speechUrls, setSpeechUrls, etc?

app/frontend/src/components/Answer/Answer.tsx
```diff
@@ -23,11 +29,16 @@ interface Props {
     showFollowupQuestions?: boolean;
     showSpeechOutputBrowser?: boolean;
     showSpeechOutputAzure?: boolean;
-    speechUrl: string | null;
 }
 
 export const Answer = ({
     answer,
+    index,
```
pamelafox, line 36: Hm so it feels a little awkward that we have to send so many speech-specific properties to Answer. Ideally it could feel a bit more modular.
Would it be better if there was a SpeechConfig constructed in Chat.tsx/Ask.tsx, and that was passed to Answer and then to the component?

pamelafox:
Hm I just had a bug where I got an answer, heard the speech, then cleared chat and asked anopther question, and it played the original speech. I'm trying to replicate it to figure out what leads to it.

john0isaac:
When you clear the chat all of the generated voice urls are set to null so, i'm not sure about that.
Maybe it was cached in the audio var?

app/frontend/src/pages/ask/Ask.tsx
john0isaac: Even though it's a list, it consists of only one url, wasn't sure about naming it like that but I'm happy to make the change.

app/frontend/src/components/Answer/Answer.tsx
john0isaac: Sure let me see what I can do.

app/frontend/src/pages/ask/Ask.tsx
pamelafox: Ah, is it only a list so that you can update it from the child component?
Is this the standard approach in React these days for child->parent updates?

john0isaac: Not sure if it's the standard or not.
But the Chat.tsx requires it to be a list as there are a list of messages,
while the Ask.tsx doesn't require it to be a list as there is only one message.

The child component needs it be of the same type as it's managed using a state to update the Parent (i.e. speechUrls in Chat.tsx) state from the Child (i.e. SpeechAzure)
So, I opted into creating it as a list for both, keep in mind that this is needed to always have a list of generated urls to not regenerate them every time.

All of this weird logic can be removed if the update was done in the Chat.tsx but to keep it modular all of these stuff are required to be passed around and have the same types.


If you have a better idea / prefer something else please let me know.

app/frontend/src/components/Answer/Answer.tsx
john0isaac: Done.
It looks way much better now, thanks!

app/frontend/src/pages/ask/Ask.tsx
john0isaac: I recommend you take another look on it, I'm not as reluctant as I was about now that I added the SpeechConfig thingy

john0isaac:
> Hm I just had a bug where I got an answer, heard the speech, then cleared chat and asked anopther question, and it played the original speech. I'm trying to replicate it to figure out what leads to it.

I could not replicate this issue.

app/frontend/src/pages/ask/Ask.tsx
```diff
@@ -48,13 +48,23 @@ export function Component(): JSX.Element {
...
+    const audio = useRef(new Audio()).current;
+    const [isPlaying, setIsPlaying] = useState(false);
 
     const lastQuestionRef = useRef<string>("");
 
     const [isLoading, setIsLoading] = useState<boolean>(false);
     const [error, setError] = useState<unknown>();
     const [answer, setAnswer] = useState<ChatAppResponse>();
-    const [speechUrl, setSpeechUrl] = useState<string | null>(null);
+    const [speechUrls, setSpeechUrls] = useState<(string | null)[]>([]);
```
pamelafox, line 59:
```suggestion
    // For the Ask tab, this array will hold a maximum of one URL
    const [speechUrls, setSpeechUrls] = useState<(string | null)[]>([]);
```

app/frontend/src/pages/chat/Chat.tsx
```diff
@@ -67,8 +67,8 @@ const Chat = () => {
 
     const [selectedAnswer, setSelectedAnswer] = useState<number>(0);
     const [answers, setAnswers] = useState<[user: string, response: ChatAppResponse][]>([]);
-    const [streamedAnswers, setStreamedAnswers] = useState<[user: string, response: ChatAppResponse][]>([]);
     const [speechUrls, setSpeechUrls] = useState<(string | null)[]>([]);
+    const [streamedAnswers, setStreamedAnswers] = useState<[user: string, response: ChatAppResponse][]>([]);
```
pamelafox, line 71: Did this line have to move?

john0isaac: nope, I think I deleted it and re-added it in the wrong place. let me revert this change.

john0isaac, line 71:
```suggestion
    const [streamedAnswers, setStreamedAnswers] = useState<[user: string, response: ChatAppResponse][]>([]);
    const [speechUrls, setSpeechUrls] = useState<(string | null)[]>([]);
```

pamelafox:
I haven't replicated the issue either. I think the new code architecture looks good. 

I just pushed a change which preloads the "sync" icon. That's because I was briefly seeing the no-image-indicator briefly before the sync icon came in, and I found that confusing.
I preload it by adding an invisible disabled button with the same icon, and verified it passes accessibility checks.

pamelafox (approved):

pamelafox:
Thanks, merged!

PR #1955, opened 2024-09-05
ignore resourceToken and migrationToken as being secure params

tonybaloney:
## Purpose

Closes 3 security issues

- `resourceToken` is an AZD unique name 
- `migrationToken` is null


pamelafox (approved):

ps-rule.yaml
```diff
@@ -1,3 +1,6 @@
 # YAML: Set the AZURE_BICEP_FILE_EXPANSION configuration option to enable expansion
 configuration:
   AZURE_BICEP_FILE_EXPANSION: true
+  AZURE_DEPLOYMENT_NONSENSITIVE_PARAMETER_NAMES:
+    - resourceToken
+    - migrationToken
```
pamelafox, line 6:
```suggestion
    - migrationToken

```

PR #1964, opened 2024-09-11
Better support for deploying to non-home tenant

pamelafox:
## Purpose

I've been deploying to my non-home tenant this week, and it's been tricky since DefaultAzureCredential doesn't necessarily know which tenant to use. Its chain is:
1) AzureCliCredential - which can have a default tenant override, if explicitly set
2) AzureDeveloperCliCredential - doesn't have a default tenant override, but will obey tenant_id or AZURE_TENANT_ID in global env

So sometimes my DAC will work, if I've got my az or env setup right, but often it will break.
This change makes our code more explicitly pass in the tenant_id for local use, which should be set by the azd up process.
However, its slightly backward incompatible as we were bringing tenantId into main.parameters.json, which was overriding what would otherwise be a correctly output tenantId and setting it to an empty string. So some developers may need to re-run azd provision, or go into their .env and set AZURE_TENANT_ID="" there, to get local dev working again.

```
[X] Yes - See above
[ ] No
```

```
[ ] Yes
[X] No
```

## Type of change

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## Code quality checklist


pamelafox:
CCing everyone who I've discussed DefaultAzureCredential woes with :)

app/backend/app.py
```diff
@@ -436,11 +440,13 @@ async def setup_clients():
...
 
-    # Use the current user identity to authenticate with Azure OpenAI, AI Search and Blob Storage (no secrets needed,
-    # just use 'az login' locally, and managed identity when deployed on Azure). If you need to use keys, use separate AzureKeyCredential instances with the
-    # keys for each service
-    # If you encounter a blocking error during a DefaultAzureCredential resolution, you can exclude the problematic credential by using a parameter (ex. exclude_shared_token_cache_credential=True)
-    azure_credential = DefaultAzureCredential(exclude_shared_token_cache_credential=True)
+    # Use the current user identity for keyless authentication to Azure services.
+    # This assumes you use 'azd auth login' locally, and managed identity when deployed on Azure.
+    # The managed identity is setup in the infra/ folder.
+    if os.getenv("WEBSITE_HOSTNAME"):
```
tonybaloney, line 446:
```suggestion
    if os.getenv("WEBSITE_HOSTNAME"):  # Environment variable set on Azure Web Apps
```

app/backend/prepdocs.py
```diff
@@ -381,11 +381,12 @@ async def main(strategy: Strategy, setup_index: bool = True):
...
-        AzureDeveloperCliCredential()
-        if args.tenantid is None
-        else AzureDeveloperCliCredential(tenant_id=args.tenantid, process_timeout=60)
-    )
+    if args.tenantid:
+        logger.info("Connecting to Azure services using the azd credential for tenant %s", args.tenantid)
+        azd_credential = AzureDeveloperCliCredential(tenant_id=args.tenantid, process_timeout=60)
+    else:
+        logger.info("Connecting to Azure services using the azd credential for home tenant")
+        azd_credential = AzureDeveloperCliCredential()
```
tonybaloney, line 389: Why does this one not have the same timeout?

pamelafox: I don't know! I thought about removing it for consistency but then thought I shouldnt change things I dont know the origin for. Maybe @mattgotteiner recalls a reason?

.github/workflows/azure-dev-validation.yaml
```diff
@@ -22,7 +22,9 @@ jobs:
       - name: Build Bicep for linting
         uses: azure/CLI@v2
         with:
-          inlineScript: az config set bicep.use_binary_from_path=false && az bicep build -f infra/main.bicep --stdout
+          inlineScript: |
+            export DOTNET_SYSTEM_GLOBALIZATION_INVARIANT=1
```
pamelafox, line 26: Fix for AZ CLI regression-
https://github.com/Azure/azure-cli/issues/29828

app/backend/prepdocs.py
pamelafox: Maybe we should put the timeout in both, just to give the credential more time to come back locally.

mattgotteiner (approved):
It's unfortunate DefaultAzureCredential didn't work here

app/backend/prepdocs.py
mattgotteiner: I don't think adding the timeout will have much issue. Sure, we can add it to both

pamelafox:
@mattgotteiner  It works only if you're also logged into Azure CLI and have explicitly set your tenant's subscription to the one that was used in the app. So I think it's safer to go with the more explicit credential. We'll see if we get more/less issues with this sort of approach.

PR #1979, opened 2024-09-17
Run format on main.bicep (#1968)

everton-o:


pamelafox:
This looks like an accidental PR, closing.

PR #1986, opened 2024-09-23
Refactor scripts to avoid anti-patterns, redundancy

pamelafox:
## Purpose

This PR refactors our scripts in several key ways:

* Removes the anti-pattern of exporting all of `azd env get-values` to the current environment. Instead, we only load the current environment into the currently active Python environment, or if we're only accessing a few variables in a shell script, we use `azd env get-value` to fetch the value.
* Uses os.getenv instead of argparse arguments for all cases where the "argument" is actually set in the azd environment. This affected prepdocs.py/sh/ps1 the most, which had a growing unmaintenable list of arguments. Now, only non-azd arguments can be specified.
* Removed sh/ps1 scripts where the wrapper doesn't seem necessary, for manageacls.py and adlsgen2.py. The developer just has to run them inside the Python env with the right packages, and they'll work.

```
[ ] Yes
[X] No - In theory, no, but more testing needed.
```

```
[ ] Yes
[X] No
```

## Type of change

```
[ ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## Code quality checklist


app/backend/prepdocs.py
```diff
@@ -154,10 +156,9 @@ def setup_file_processors(
     local_html_parser: bool = False,
     search_images: bool = False,
 ):
-    html_parser: Parser
-    pdf_parser: Parser
-    doc_int_parser: DocumentAnalysisParser
+    sentence_text_splitter = SentenceTextSplitter(has_image_embeddings=search_images)
```
pamelafox, line 159: I changed this code around a bit as I found some issues once I wrote tests that touched it.

app/backend/app.py
```diff
@@ -714,9 +714,10 @@ def create_app():
     # Log levels should be one of https://docs.python.org/3/library/logging.html#logging-levels
     # Set root level to WARNING to avoid seeing overly verbose logs from SDKS
     logging.basicConfig(level=logging.WARNING)
-    # Set the app logger level to INFO by default
-    default_level = "INFO"
-    app.logger.setLevel(os.getenv("APP_LOG_LEVEL", default_level))
+    # Set our own logger levels to INFO by default
+    app_level = os.getenv("APP_LOG_LEVEL", "INFO")
+    app.logger.setLevel(os.getenv("APP_LOG_LEVEL", app_level))
+    logging.getLogger("scripts").setLevel(app_level)
```
pamelafox, line 720: Since prepdocs uses the "scripts" logger, I wanted to make sure we also see its logs when using user upload feature.
Though I wonder if I should give it a different name, like "ragapp".

app/backend/load_azd_env.py
```diff
@@ -0,0 +1,23 @@
+import json
+import logging
+import subprocess
+
+from dotenv import load_dotenv
+
+logger = logging.getLogger("scripts")
+
+
+def load_azd_env():
```
pamelafox, line 10: This file actually shows up in two places, both here and in scripts folder, for convenience.

app/backend/main.py
```diff
@@ -1,3 +1,12 @@
+import os
+
 from app import create_app
+from load_azd_env import load_azd_env
+
+# WEBSITE_HOSTNAME is always set by App Service, RUNNING_IN_PRODUCTION is set in main.bicep
+RUNNING_ON_AZURE = os.getenv("WEBSITE_HOSTNAME") is not None or os.getenv("RUNNING_IN_PRODUCTION") is not None
```
pamelafox, line 7: This code also exists in two places- 
1) here, so that we load the environment when starting up the app locally
2) in app.py, so that we can decide what kind of credential to use

app/backend/prepdocs.py
```diff
@@ -218,111 +238,19 @@ async def main(strategy: Strategy, setup_index: bool = True):
 if __name__ == "__main__":
     parser = argparse.ArgumentParser(
         description="Prepare documents by extracting content from PDFs, splitting content into sections, uploading to blob storage, and indexing in a search index.",
-        epilog="Example: prepdocs.py '.\\data\*' --storageaccount myaccount --container mycontainer --searchservice mysearch --index myindex -v",
+        epilog="Example: prepdocs.py '.\\data\*' -v",
```
pamelafox, line 241: I only kept the arguments that did *not* come directly from azd env

app/backend/prepdocs.py
```diff
@@ -167,31 +168,50 @@ def setup_file_processors(
...
+                ".xlsx": FileProcessor(doc_int_parser, sentence_text_splitter),
+                ".png": FileProcessor(doc_int_parser, sentence_text_splitter),
+                ".jpg": FileProcessor(doc_int_parser, sentence_text_splitter),
+                ".jpeg": FileProcessor(doc_int_parser, sentence_text_splitter),
+                ".tiff": FileProcessor(doc_int_parser, sentence_text_splitter),
+                ".bmp": FileProcessor(doc_int_parser, sentence_text_splitter),
+                ".heic": FileProcessor(doc_int_parser, sentence_text_splitter),
+            }
+        )
+    return file_processors
```
pamelafox, line 214: ^^ all of that code change was due to issues found via tests/mypy

app/backend/prepdocs.py
```diff
@@ -333,42 +261,32 @@ async def main(strategy: Strategy, setup_index: bool = True):
...
     parser.add_argument(
-        "--visionendpoint",
+        "--documentintelligencekey",
         required=False,
-        help="Optional, required if --searchimages is specified. Endpoint of Azure AI Vision service to use when embedding images.",
+        help="Optional. Use this Azure Document Intelligence account key instead of the current user identity to login (use az login to set current user for Azure)",
     )
     parser.add_argument(
-        "--useintvectorization",
+        "--searchserviceassignedid",
```
pamelafox, line 285: I cannot find ANY evidence of us telling folks how to use this, or even actually using it. Do we even need this arg??

app/frontend/src/pages/chat/Chat.tsx
```diff
@@ -131,7 +131,7 @@ const Chat = () => {
                 if (event["context"] && event["context"]["data_points"]) {
                     event["message"] = event["delta"];
                     askResponse = event as ChatAppResponse;
-                } else if (event["delta"]["content"]) {
+                } else if (event["delta"] && event["delta"]["content"]) {
```
pamelafox, line 134: Unrelated error that I found while presenting. This error meant we arent currently rendering errors correctly, as {error: } responses dont contain a delta key.

infra/main.parameters.json
```diff
@@ -116,6 +116,9 @@
     "embeddingDimensions": {
       "value": "${AZURE_OPENAI_EMB_DIMENSIONS}"
     },
+    "gpt4vDeploymentCapacity":{
+      "value": "${AZURE_OPENAI_GPT4V_DEPLOYMENT_CAPACITY=10}"
```
pamelafox, line 120: Also unrelated, I needed to be able to set capacity to get the vision endpoint deployed to our restrictive tenants. I can move to other branch if desired.

samples/document-security/README.md
```diff
@@ -23,62 +23,66 @@ The [azure-search-openai-demo](/) project can set up a full RAG chat app on Azur
 ## Table of Contents
 
 - [Requirements](#requirements)
-- [Setting up Microsoft Entra ID Apps](#setting-up-entra-id-apps)
+- [Setting up Microsoft Entra applications](#setting-up-microsoft-entra-applications)
```
pamelafox, line 26: This file is just a copy paste of login_and_acl.md

app/backend/app.py
mattgotteiner: I think scripts is an acceptable name for now - we can change it later if it's not working

app/backend/prepdocs.py
```diff
@@ -167,31 +168,50 @@ def setup_file_processors(
...
-    sentence_text_splitter = SentenceTextSplitter(has_image_embeddings=search_images)
-    return {
-        ".pdf": FileProcessor(pdf_parser, sentence_text_splitter),
-        ".html": FileProcessor(html_parser, sentence_text_splitter),
+    else:
+        logger.warning("No HTML parser available")
+
+    # These file formats can always be parsed:
+    file_processors = {
         ".json": FileProcessor(JsonParser(), SimpleTextSplitter()),
```
mattgotteiner, line 190: why did we remove all these file processors?

mattgotteiner: oh, it's below

app/backend/prepdocs.py
mattgotteiner: yay mypy

app/backend/prepdocs.py
mattgotteiner: yeah, let's create a follow-up pr to remove this

app/backend/prepdocs.py
```diff
@@ -378,12 +296,17 @@ async def main(strategy: Strategy, setup_index: bool = True):
...
+    use_gptvision = os.getenv("USE_GPT4V", "").lower() == "true"
+    use_acls = os.getenv("AZURE_ADLS_GEN2_STORAGE_ACCOUNT") is not None
+    dont_use_vectors = os.getenv("USE_VECTORS", "").lower() == "false"
 
-    # Use the current user identity to connect to Azure services unless a key is explicitly set for any of them
-    if args.tenantid:
-        logger.info("Connecting to Azure services using the azd credential for tenant %s", args.tenantid)
-        azd_credential = AzureDeveloperCliCredential(tenant_id=args.tenantid, process_timeout=60)
+    # Use the current user identity to connect to Azure services. See infra/main.bicep for role assignments.
+    if tenant_id := os.getenv("AZURE_TENANT_ID"):
```
mattgotteiner, line 307: I like the idea of using the env vars instead of args - great improvement!

samples/document-security/README.md
mattgotteiner: can we eventually remove this? OK to leave it for now

mattgotteiner (approved):
Thanks, this is amazing

samples/document-security/README.md
pamelafox: I assume you mean, can we we eventually remove the duplicate copy of login_and_acl.md?

Arun wants the ACL feature to show up as a completely separate sample in the Microsoft Learn Samples browser, which requires the copy. If it's annoying to maintain, we could discuss removing it, and it wouldn't show up in the samples browser anymore.

Issue #1863, opened 2024-07-26
Feature request: Chat history persisted to Cosmos DB


hfaouaz-rcg:
> Please provide us with the following information:
> ---------------------------------------------------------------

### This issue is for a: (mark with an `x`)
```
- [ ] bug report -> please search issues before submitting
- [X] feature request
- [ ] documentation issue or request
- [ ] regression (a behavior that used to work and stopped in a new release)
```

### Minimal steps to reproduce
>

### Any log messages given by the failure
>

### Expected/desired behavior
>

To Capture and maintain question/answers in cosmos.    I was able to refactor current codebase and chatGPT code base to capture questions in cosmos. still some work needs to be done, but I am wondering if this feature is on the roadmap.  Once I have a complete working solution, I will create a feature enhancement PR.

Build cosmos biceps  <- done
provide api for history capture/generate/list <- done
capture question <-done
capture answer 
session base/conversation base 

### OS and Version?
> Windows 7, 8 or 10. Linux (which distribution). macOS (Yosemite? El Capitan? Sierra?)

### azd version?
> run `azd version` and copy paste here.

### Versions
>

### Mention any other details that might be useful

> ---------------------------------------------------------------
> Thanks! We'll be in touch soon.

DuboisABB:
This seems like a duplicate of #489. There is also a stale PR that implemented this (although probably not based on sample-app-aoai-chatGPT):  #259
I'm also interested in this functionality. Do you have a public fork where I could follow your development? I might be able to contribute too, as I had also planned on doing this.

hfaouaz-rcg:
> This seems like a duplicate of #489. There is also a stale PR that implemented this (although probably not based on sample-app-aoai-chatGPT): #259 I'm also interested in this functionality. Do you have a public fork where I could follow your development? I might be able to contribute too, as I had also planned on doing this.

I am currently developing it for a client.  Let me see if I can create a fork minus clients changes. would love the help.

cloudhunnter:
I also have a client asking for this kind of feature. @hfaouaz-rcg did you made any steps towards a working solution? What kind of services do you implement for this? i was thinking about cosmos db, but im fairly new to databases tbh.

DuboisABB:
This request is pretty much the same: #1822

idrisfl:
I'm also interested in this kind of feature for one of my clients. We are looking to implement a feedback loop on the responses, which would be stored in a Cosmos DB.

[...]

Issue #977, opened 2023-11-21
Chat history feature


UmarMohamedUsman:
### This issue is for a: (mark with an `x`)
```
- [ ] bug report -> please search issues before submitting
- [x] feature request
- [ ] documentation issue or request
- [ ] regression (a behavior that used to work and stopped in a new release)
```
### Expected/desired behavior
Currently chat history is not persisted. In an enterprise environment its critical to provide the ability to review/audit chat history.  
Chat with your data repo has this feature: https://github.com/microsoft/sample-app-aoai-chatGPT

apiontek:
It looks like there was a [PR 259](https://github.com/Azure-Samples/azure-search-openai-demo/pull/259) related to this, but several months back. I haven't tried it myself but I wonder if the approach is sound or if there's a reason it hasn't been included.

michbalazs:
This would be really helpful. Any update ?

pamelafox:
We now have a chat history feature that stores it in IndexedDB:
https://github.com/Azure-Samples/azure-search-openai-demo/blob/main/docs/deploy_features.md#enabling-chat-history

I'll close this issue, but if you want Cosmos DB persistence, please subscribe/star/comment on this issue:
https://github.com/Azure-Samples/azure-search-openai-demo/issues/977

PR #259, opened 2023-06-02
Adding Conversation History, "Vanilla" chat page, and tweaked local dev settings for hot swap

bdleavitt:
## Purpose
Deploying module deploys CosmosDB instance for storing conversation history. Uses Azure app service's built-in authentication to separate user conversations from each othe.r 

## Does this introduce a breaking change?

```
[ ] Yes
[x] No
```

## Pull Request Type
What kind of change does this Pull Request introduce?

```
[ ] Bugfix
[x] Feature
[ ] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[ ] Documentation content changes
[ ] Other... Please describe:
```

## How to Test
Initialize repo `azd init`
deploy infra `azd up`
Enable auth on app service 

* Test the code
Open website. Navigate to "ChatGPT" tab. 

## What to Check
Verify that the following are valid
* Enter a prompt and get a response on the ChatGPT page. 
* Click on the "Conversation List" to make sure the conversation is saved. 
* Try to delete the conversation and verify it drops out of the list.

PR #1988, opened 2024-09-25
Add chat history feature

fujita-h:
This commit adds the following new files:
- HistoryItem: A new component for displaying chat history items.
- HistoryPanel: A new component for displaying the chat history panel.
- HistoryButton: A new component for opening the chat history.
- HistoryProviders: A new module containing history provider classes.

These changes include functionality to support future enhancements to the chat history feature.

The chat history feature is only enabled when you explicitly set the environment variable `USE_CHAT_HISTORY`.

The history function uses the browser's IndexedDB. I understand that there is also an approach to use CosmosDB, but this requires enabling authentication to identify users, so I decided to start with IndexedDB, which does not depend on authentication.

To make it easier to implement other data stores, I made it easier to implement additional data stores with `HistoryProviders`.

Related: #977 #1863 #259 

```
[ ] Yes
[x] No
```

```
[ ] Yes
[x] No
```

## Type of change

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

## Code quality checklist


fujita-h:

pamelafox:
Ooo, very interesting approach with the IndexedDB! It sounds like you've designed it very thoughtfully, I'll take a look soon.

app/frontend/src/pages/chat/Chat.tsx
```diff
@@ -194,7 +203,8 @@ const Chat = () => {
                     }
                 },
                 // AI Chat Protocol: Client must pass on any session state received from the server
-                session_state: answers.length ? answers[answers.length - 1][1].session_state : null
+                // Set new UUID as session state for the first request
+                session_state: answers.length ? answers[answers.length - 1][1].session_state : crypto.randomUUID()
```
pamelafox, line 207: Do you think we'd also use this randomUUID for apps with logged in users, or would we use Entra OIDs for them? Am wondering if it'd be better to have a helper function to generate this ID.
And if the history storage was in CosmosDB, would we still generate the UUID on the client? Want to make sure this is forward-compatible if possible.

app/frontend/package.json
```diff
@@ -22,6 +22,7 @@
     "i18next": "^23.12.2",
     "i18next-browser-languagedetector": "^8.0.0",
     "i18next-http-backend": "^2.5.2",
+    "idb": "^8.0.0",
```
pamelafox, line 25: This seems like a reasonable addition according to packagephobia and snyk:

https://packagephobia.com/result?p=idb
https://snyk.io/advisor/npm-package/idb

app/backend/config.py
```diff
@@ -22,3 +22,4 @@
 CONFIG_SPEECH_SERVICE_LOCATION = "speech_service_location"
 CONFIG_SPEECH_SERVICE_TOKEN = "speech_service_token"
 CONFIG_SPEECH_SERVICE_VOICE = "speech_service_voice"
+CONFIG_CHAT_HISTORY_ENABLED = "chat_history_enabled"
```
pamelafox, line 25: Given that a "chat history" feature means different things to different people, I think it might help for the config and env variable to be more specific.

For example, for the env var:

USE_CHAT_HISTORY_BROWSER

In future, we might have:
USE_CHAT_HISTORY_COSMOS

That'd be similar to how we did speech output.

app/frontend/src/locales/es/translation.json
```diff
@@ -6,6 +6,19 @@
     "login": "Iniciar sesión",
     "logout": "Cerrar sesión",
     "clearChat": "Borrar chat",
+    "history": {
+        "chatHistory": "Historial de chat",
```
pamelafox, line 10: Did you use AI for the translations? i.e. do we need a human to review them?

app/frontend/src/api/models.ts
```diff
@@ -89,6 +89,7 @@ export type Config = {
     showSpeechInput: boolean;
     showSpeechOutputBrowser: boolean;
     showSpeechOutputAzure: boolean;
+    showChatHistory: boolean;
```
pamelafox, line 92:
```suggestion
    showChatHistoryBrowser: boolean;
```

pamelafox: Per my other comment, I think we should be clear about what's enabled.

app/frontend/src/pages/chat/Chat.tsx
fujita-h: Thank you for pointing that out.
You're right, it seems we should think about this carefully.

We need to assign an ID to each "series of chats" to identify them, regardless of whether the user is logged in or not. We'll call this `sessionId`.

## In the case of IndexedDB
- On the client side (React), `sessionId` does not necessarily have to be a UUID, as long as it is a unique key within the client side.
- There is also no need to exchange `sessionId` with the server side using `session_state` (the current implementation of storing `sessionId` in `session_state` is in anticipation of expansion to CosmosDB).

## In the case of CosmosDB
- It is expected that the design will set the user's Entra OID as the partition key and `sessionId` as the unique key.
- In CosmosDB, `conversationId` does not need to be a UUID, as long as it is unique within the partition.
- As mentioned above, `sessionId` does not need to be a UUID on the client side (React).
- `sessionId` is a key in CosmosDB. Therefore, it is preferable to generate it on the server side rather than on the client side.
- As the name `session_state` suggests, it is natural that the `sessionId` stored here is generated on the server side.
- When chat messages are sent and received, the `sessionId` needs to be exchanged between the server side and the client side. `session_state` can be used for this.

## Fix Plan
The plan is to summarize these points and fix them in the following plan.
- If the request from the client contains `session_state`, it will be used as `sessionId`. If not, a new one will be generated on the server and added to `session_state`.
- `sessionId` does not necessarily need to be UUID, but unless there is a specific reason, there is no major problem with UUID, so we will basically generate it with `uuid.uuid4()`.
- Considering the possibility that UUID may be inconvenient when other storage methods are added in the future, we will create a helper function that generates `sessionId` according to the storage method used.

app/frontend/src/locales/es/translation.json
pamelafox: Okay, will do!

fujita-h:
I have corrected the points you pointed out, so please review again.
The environment variable to enable this feature has been changed to `USE_CHAT_HISTORY_BROWSER`.

fujita-h:
Thank you for pointing that out. I had forgotten about bicep.
`ENABLE_LANGUAGE_PICKER` and `USE_SPEECH_INPUT_BROWSER` are outside the scope of this PR, so I have omitted them.

app/frontend/src/locales/fr/translation.json
```diff
@@ -6,6 +6,19 @@
     "login": "Connexion",
     "logout": "Déconnexion",
     "clearChat": "Effacer le chat",
+    "history": {
+        "chatHistory": "Historique de chat",
```
manekinekko, line 10:
```suggestion
        "chatHistory": "Historique du chat",
```

app/frontend/src/locales/fr/translation.json
```diff
@@ -6,6 +6,19 @@
...
+        "openChatHistory": "Ouvrir l'historique de chat",
```
manekinekko, line 11:
```suggestion
        "openChatHistory": "Ouvrir l'historique du chat",
```

app/frontend/src/locales/fr/translation.json
```diff
@@ -6,6 +6,19 @@
...
+        "noHistory": "Pas d'historique de chat",
+        "deleteModalTitle": "Supprimer l'historique de chat",
+        "deleteModalDescription": "Cette action est irréversible. Supprimer cet historique de chat ?",
```
manekinekko, line 14:
```suggestion
        "deleteModalDescription": "Cette action est irréversible. Supprimer l'historique de ce chat ?",
```

app/frontend/src/locales/fr/translation.json
```diff
@@ -6,6 +6,19 @@
     "login": "Connexion",
     "logout": "Déconnexion",
     "clearChat": "Effacer le chat",
+    "history": {
+        "chatHistory": "Historique de chat",
+        "openChatHistory": "Ouvrir l'historique de chat",
+        "noHistory": "Pas d'historique de chat",
```
manekinekko, line 12:
```suggestion
        "noHistory": "Pas d'historique dans ce chat",
```

app/frontend/src/locales/fr/translation.json
```diff
@@ -6,6 +6,19 @@
...
+        "deleteModalTitle": "Supprimer l'historique de chat",
```
manekinekko, line 13:
```suggestion
        "deleteModalTitle": "Supprimer l'historique du chat",
```

app/frontend/src/locales/fr/translation.json
pamelafox: Merci, Wassim!

app/frontend/src/locales/es/translation.json
pamelafox: I've asked a colleague to review the es translations. Wassim has reviewed fr.

pamelafox (approved):
I'm approving the change, but will await a review of the es translations and will check with @mattgotteiner before merging.

mattgotteiner (approved):
approved pending translation review

app/frontend/src/locales/es/translation.json
miguelmsft: I approve of the Spanish translation 👍

pamelafox: Thanks miguel! Resolving this thread.

pamelafox (approved):

ibrahimnasir0:
@fujita-h Where we will be able to see this chat history  ?
Can you guide on that

fujita-h:
@ibrahimnasir0 Just set the environment variables as follows:
`azd env set USE_CHAT_HISTORY_BROWSER true`

ibrahimnasir0:
@fujita-h Thanks for the help! By the way, would you mind sharing your LinkedIn profile? I would love to connect and discuss the implementation further

ibrahimnasir0:
Hi @fujita-h 
I'm working on integrating Cosmos DB with this updated project to save chat history along with the user OID (since login and Entra ID are enabled). Could you provide some guidance on how to proceed with this?

PR #1992, opened 2024-09-26
Changes for easier working with local models

pamelafox:
## Purpose

This PR makes it easier for developers who want to use local models in development, either for cost reasons or comparison reasons. I tested with llama3.1:8b from Ollama, which works even for the Chat tab, as well as phi3.5 which works only for the Ask tab (no function calling).

The primary change is that I enabled the use of the GPT tokenizers even for non-GPT models. This means that the token counting will be inaccurate for local models, but that isn't practically an issue unless you have very long conversations.

```
[ ] Yes
[X] No
```

```
[ ] Yes
[X] No
```

## Type of change

```
[ ] Bugfix
[X] Feature
[ ] Other... Please describe:
```

## Code quality checklist

### DIFF
--- a/app/backend/approaches/chatreadretrieveread.py
+++ b/app/backend/approaches/chatreadretrieveread.py
@@ -51,7 +51,7 @@ class ChatReadRetrieveReadApproach(ChatApproach):
         self.content_field = content_field
         self.query_language = query_language
         self.query_speller = query_speller
-        self.chatgpt_token_limit = get_token_limit(chatgpt_model)
+        self.chatgpt_token_limit = get_token_limit(chatgpt_model, default_to_minimum=self.ALLOW_NON_GPT_MODELS)
 
     @property
     def system_message_chat_conversation(self):
@@ -133,6 +133,7 @@ class ChatReadRetrieveReadApproach(ChatApproach):
             past_messages=messages[:-1],
             new_user_content=user_query_request,
             max_tokens=self.chatgpt_token_limit - query_response_token_limit,
+            fallback_to_default=self.ALLOW_NON_GPT_MODELS,
         )
 
         chat_completion: ChatCompletion = await self.openai_client.chat.completions.create(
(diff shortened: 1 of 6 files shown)


app/backend/approaches/chatreadretrieveread.py
```diff
@@ -51,7 +51,7 @@ def __init__(
         self.content_field = content_field
         self.query_language = query_language
         self.query_speller = query_speller
-        self.chatgpt_token_limit = get_token_limit(chatgpt_model)
+        self.chatgpt_token_limit = get_token_limit(chatgpt_model, default_to_minimum=self.ALLOW_NON_GPT_MODELS)
```
mattgotteiner, line 54: We might want to update the docs to say that we can't estimate tokens in this case

mattgotteiner (approved):

app/backend/approaches/chatreadretrieveread.py
mattgotteiner: One question - does it even try to get the token limit if default_to_minimum is True? It seems to be always true even if we aren't using a non-gpt model

mattgotteiner: https://github.com/pamelafox/openai-messages-token-helper/blob/eb59b9a202505720bc2f1db7dadf1b78500eb114/src/openai_messages_token_helper/model_helper.py#L42-L58

mattgotteiner: looks like it does

pamelafox: default_to_minimum is False unless set otherwise, and you'd get an error with a non-recognized model:
https://github.com/pamelafox/openai-messages-token-helper/blob/eb59b9a202505720bc2f1db7dadf1b78500eb114/src/openai_messages_token_helper/model_helper.py#L42

pamelafox: Updated docs!

Issue #2017, opened 2024-10-04
Markdown URLs with non-file extensions are incorrectly recognized as citations


EMjetrot:
### This issue is for a: (mark with an `x`)
```
- [x] bug report -> please search issues before submitting
- [ ] feature request
- [ ] documentation issue or request
- [ ] regression (a behavior that used to work and stopped in a new release)
```

### Minimal steps to reproduce
>
I use a mixture of PDF and markdown files as data sources, but noticed that links in the markdown text get interpreted as citations. Here is an example of some markdown with a link inside (ignore the danish language):

`I kraft af dit ansættelsesforhold i styrelsen indsamler, opbevarer og behandler Koncern HR en række personlige oplysninger om dig. Du kan læse meget mere herom lige her [Sådan behandler Koncern HR persondata om dig](https://somefile.pdf).`

Since the title [Sådan behandler Koncern HR persondata om dig] of the link uses square brackets like citations, the title wrongly ends up being listed among the sources while the link doesn't get formatted correctly in the answer.

<img width="1011" alt="before_change" src="https://github.com/user-attachments/assets/49285957-3b52-4ddf-9c09-27da44f2131b">

### Any log messages given by the failure
>
None
### Expected/desired behavior
>
If we assume that a citation is always a file, we can modify the code to check whether the citation candidate contains a valid file extension. Only candidates with a recognized file extension will be treated as citations, while any other text enclosed in square brackets will remain unchanged.

Here is the code from line 32 in ./app/frontend/src/components/Answer/AnswerParser.tsx before the change:

```
const fragments: string[] = parts.map((part, index) => {
        if (index % 2 === 0) {
            return part;
        } else {
            let citationIndex: number;
            if (citations.indexOf(part) !== -1) {
                citationIndex = citations.indexOf(part) + 1;
            } else {
                citations.push(part);
                citationIndex = citations.length;
            }

            const path = getCitationFilePath(part);

            return renderToStaticMarkup(
                <a className="supContainer" title={part} onClick={() => onCitationClicked(path)}>
                    <sup>{citationIndex}</sup>
                </a>
            );
        }
    });
```

And here is the changed code where citation candidates are checked for common file extensions:

```
const fragments: string[] = parts.map((part, index) => {

        if (index % 2 === 0) {
            // This is text outside square brackets (regular text)
            return part;
        } else {
            // This is text inside square brackets (citation or markdown URL)
            // Check if the citation ends with a valid file extension
[...]

Issue #735, opened 2023-10-05
Relevant text from the source document is being removed from the answer and used as a citation if it is in square brackets


elhele:
> Please provide us with the following information:
> ---------------------------------------------------------------

### This issue is for a: (mark with an `x`)
```
- [x] bug report -> please search issues before submitting
- [ ] feature request
- [ ] documentation issue or request
- [ ] regression (a behavior that used to work and stopped in a new release)
```

### Minimal steps to reproduce
> Load a document that contains text in square brackets, e.g. "[km / h ]"

### Any log messages given by the failure
> There are no errors in the logs, because it works as it supposed to work according to the code, if I understand it correctly: "Do not include any text inside [] or <<>> in the search query terms. ".

### Expected/desired behavior
> The relevant text is inside the answer and not the citations.

### OS and Version?
>  macOS (Sonoma, Version 14.0)

### azd version?
> azd version 1.3.1 

### Versions
> Repo version from 25.09 with fixed werkzeug-requirements

### Mention any other details that might be useful
Please, see the screenshots

> ---------------------------------------------------------------
<img width="370" alt="Bildschirmfoto 2023-10-05 um 09 53 09" src="https://github.com/Azure-Samples/azure-search-openai-demo/assets/13983474/2c18e503-e315-4ade-8cc2-1250fd24de12">

> Thanks! We'll be in touch soon.
<img width="456" alt="Bildschirmfoto 2023-10-05 um 09 53 44" src="https://github.com/Azure-Samples/azure-search-openai-demo/assets/13983474/c8561f54-05b6-4358-b3be-5e7d6a954f47">

igforce:
Here is a fix that works for me.

Change the chatreadretrieveread.py prompt to include `[[info1.txt]][[info2.pdf]]` instead of just` [info1.txt][info2.pdf]` or any kind of additional structure that is less likely format than regular response.

Then change the AnswerParser.tsx `const parts = parsedAnswer.split(/\[\[([^\]]+)\]\]/g);` to now format with the structure you defined in the prompt. In my case it's '[[]]'

elhele:
@igforce  I also tried it like this, but somehow I had problems with "Suggest follow-up questions" functionality afterwards. It could also be the case that I broke it somehow else though. I'll test it again. Thank you!
I though, may be it also creates problems with this part and these indices should be adjusted:

`if (isStreaming){
        let lastIndex = parsedAnswer.length;
[...]

PR #2049, opened 2024-10-18
Bugfix: Citation Parsing Issue (Version 1)

bnodir:
## Purpose
Fix for:
- #2017
- #735

【**_Before_**】
image

【**_After_**】
image
→ As shown in this example, if the answers include file names with square brackets, they are still treated as citations, as below. I am wondering how to distinguish between the two: actual citations with URIs and other ones. 

_Citation: 1. role_library.pdf#page=28 **2. somefile.pdf 3. sometxt.txt**_

```
[ ] Yes
[x] No
```

```
[ ] Yes
[x] No
```

## Type of change

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## Code quality checklist


pamelafox:
Re the discussion in that thread, we talked about a different approach--
 "Double square brackets would be a lot more simple fix and I realize now that the fix I proposed doesn't work as intended if the URL in a link in a markdown file contains one of the file extensions we check for."

I think your proposal suffers from that issue right? So my suggestion was double/triple brackets.

A third approach: We could always send back a list of the possible citations in the context, and require that a valid citation matches one of those. They're already there now, I believe, in the data_points, but they could also be in a separate list like citations: ["bla.pdf", ...]

pamelafox:
Closing this PR since I just merged v2. Thanks so much for the fix(es)!

PR #2056, opened 2024-10-19
Bugfix: Citation Parsing Issue (Version 2)

bnodir:
## Purpose

Fix for:
- #2017
- #735

In this PR, I implemented the following approach:
> From https://github.com/Azure-Samples/azure-search-openai-demo/pull/2049#issuecomment-2421965701
A third approach: We could always send back a list of the possible citations in the context, and require that a valid citation matches one of those. They're already there now, I believe, in the data_points, but they could also be in a separate list like citations: ["bla.pdf", ...]

【_**Before**_】
image

【_**After**_】
image

【_**Test strings**_】
_◇Testing [km/h], [speed.pdf], [Menu]-[Sub-menu]-[Command], report.txt [summary.pdf], [[document.docx]], [[[file.docx]]], [note1]], [[[note2]], [cm], [m], [kg], [g], [L], [ml], [°C], [°F].◆_
[console-export_[Streaming]=ON.log](https://github.com/user-attachments/files/17444304/console-export_.Streaming.ON.log)
[console-export_[Streaming]=OFF.log](https://github.com/user-attachments/files/17444302/console-export_.Streaming.OFF.log)

```
[ ] Yes
[x] No
```

```
[ ] Yes
[x] No
```

## Type of change

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## Code quality checklist


app/frontend/src/components/Answer/AnswerParser.tsx
```diff
@@ -1,16 +1,44 @@
...
+    if (Array.isArray(contextDataPoints)) {
+        dataPointsArray = contextDataPoints;
+    } else if (contextDataPoints && Array.isArray(contextDataPoints.text)) {
+        dataPointsArray = contextDataPoints.text;
+    } else {
+        return false;
+    }
+
+    // Check if the citation is included in any of the strings within the text array
+    const isValidCitation = dataPointsArray.some(dataPoint => dataPoint.includes(citationCandidate));
```
pamelafox, line 27: I think checking startsWith() would also work, right? Given the current response:

```
        "data_points": {
            "text": [
                "Benefit_Options.pdf#page=2:  With Northwind Health Plus, you can choose from a variety of in-network providers, ...
```

Or would startsWith not cover a situation?

bnodir: @pamelafox - Thank you, `startsWith()` worked as well:
image

(I also noticed another issue: as seen in the screenshot, "Capture the full web page with scroll" in Edge does not seem to work well with the question input field.)

bnodir: Here is the result with` startsWith()`:
image

pamelafox:
@bnodir I've had a reported regression for customers with whitespace in their filenames. This PR used /^[^\s]+\.[a-zA-Z0-9]+/ for the regex, could we loosen that to allow internal whitespace?

bnodir:
@pamelafox - Sorry for the regression. I was able to reproduce it on my side. Here is my suggestion, what do you think:

- **_Reproduce:_**
With `const regex = /^[^\s]+\.[a-zA-Z0-9]+/;`

image

- **_Fix for regression:_**
Adjust the regex to allow internal whitespace in the filenames by using `const regex = /^[^\s]+(\s[^\s]+)*\.[a-zA-Z0-9]+(#.*)?$/;`

image
image
→　There is still a problem if the PDF filename has whitespace in it and we don't have the fragment identifier (like #page=8) at the end of it. It seems to be a potential issue (the PDF file can be opened by clicking the citation, but it won't jump to the specific page) to me. Could you please check it?

nickmachairas:
Pardon my jumping in. This is definitely an issue for my new deployments, ever since `AnswerParser.tsx` was significantly changed. I tried `const regex = /^[^\s]+(\s[^\s]+)*\.[a-zA-Z0-9]+(#.*)?$/;` and it is very inconsistent, with or without whitespaces and/or pages at the end.

For example, `[GSDG Ex. B - 6Cle Rx. Klepper.pdf#page=54]` and `[GSDG Ex. B - 6Cle Rx. Klepper.pdf]` are both parsed as expected and converted to links, however, `[6-4-14_1455x_RE STAGNEW Rex CAT 16.pdf]` is not.

What's even more baffling is that the same sources that were parsed to links sometimes will not on a new chat session and vise versa.

bnodir:
I apologize for the inconvenience caused by the recent changes. I’ll work on resolving this issue to the best of my ability over the weekend, though I can’t guarantee a complete solution. If needed, I may also consider requesting a reversion of the changes from this PR. Thank you for your patience.

nickmachairas:
No worries, please let me know if you'd like me to test any changes. Thanks

bnodir:
@nickmachairas - I came up with const `regex = /.+\.\w{1,}(?:#\S*)?$/` and tested it on my end, confirming that it works. Could you please verify it as well?

[Tested Strings.txt](https://github.com/user-attachments/files/17532726/Tested.Strings.txt)

[[Streaming=ON]console-export-2024-10-27_12-59-18.txt](https://github.com/user-attachments/files/17532721/Streaming.ON.console-export-2024-10-27_12-59-18.txt)

image

[[Streaming=OFF]console-export-2024-10-27_13-4-16.txt](https://github.com/user-attachments/files/17532722/Streaming.OFF.console-export-2024-10-27_13-4-16.txt)
image

```
// Function to validate citation format and check if dataPoint includes citationCandidate
function isCitationValid(contextDataPoints: any, citationCandidate: string): boolean {
    const regex = /.+\.\w{1,}(?:#\S*)?$/;
    console.log("🌐 Checking citation format for:", citationCandidate);
    if (!regex.test(citationCandidate)) {
        console.log("✖ Citation format is invalid:", citationCandidate);
        return false;
    }
    console.log("✔ Citation format is valid.");

    // Check if contextDataPoints is an object with a text property that is an array
    let dataPointsArray: string[];
    if (Array.isArray(contextDataPoints)) {
        dataPointsArray = contextDataPoints;
        console.log("📂 contextDataPoints is an array:", dataPointsArray);
    } else if (contextDataPoints && Array.isArray(contextDataPoints.text)) {
        dataPointsArray = contextDataPoints.text;
        console.log("📂 contextDataPoints.text is an array:", dataPointsArray);
    } else {
        console.log("✖ contextDataPoints does not contain a valid array.");
        return false;
    }

    // Check if any data point includes the citationCandidate
    const isValidCitation = dataPointsArray.some(dataPoint => {
        console.log("🔍 Checking dataPoint:", dataPoint);
        return dataPoint.startsWith(citationCandidate);
    });

    console.log(isValidCitation ? "✔ Citation candidate found in data points." : "✖ Citation candidate not found in data points.");
    return isValidCitation;
}
```

PR #2063, opened 2024-10-21
Feature: Store chat history in Cosmos DB

fujita-h:
## Purpose

This PR extends the chat history feature added in #1988 to store chat history server-side (Cosmos DB) when user authentication is enabled.

```
[ ] Yes
[x] No
```

```
[ ] Yes
[x] No
```

## Type of change

```
[ ] Bugfix
[x] Feature
[ ] Other... Please describe:
```

## Code quality checklist


pamelafox:
Woot! I know this is a very popular request, thanks for the PR, I'll review it this week. In the meantime, if anyone subscribed tries it out as well, please report on the PR with your feedback as well. I'll also ask Cosmos DB team to take a look.

app/backend/app.py
```diff
@@ -397,6 +407,129 @@ async def list_uploaded(auth_claims: dict[str, Any]):
     return jsonify(files), 200
 
 
+@bp.post("/chat_history")
```
pamelafox, line 410: Do you think the chat history endpoints could be in a different file/blueprint? That would make it to bring in additional providers (I've been asked about Azure SQL, for example) without bloating app.py.
I know we only have a single file/blueprint for routes now though, so it's possible that I haven't organized it in such a way that it's feasible.

app/frontend/src/api/models.ts
```diff
@@ -103,3 +104,21 @@ export interface SpeechConfig {
     isPlaying: boolean;
     setIsPlaying: (isPlaying: boolean) => void;
 }
+
+export type HistoryListApiResponse = {
+    items: {
+        id: string;
+        entra_id: string;
+        title?: string;
+        _ts: number;
```
pamelafox, line 113: What/why is _ts? It's a funny variable name, so it could benefit from either an inline comment or a rename.

app/frontend/src/api/models.ts
```diff
@@ -103,3 +104,21 @@ export interface SpeechConfig {
     isPlaying: boolean;
     setIsPlaying: (isPlaying: boolean) => void;
 }
+
+export type HistoryListApiResponse = {
+    items: {
+        id: string;
+        entra_id: string;
```
pamelafox, line 111: This is oid, right? Maybe entra_oid would be slightly clearer (versus an Entra group id)

infra/core/security/documentdb-sql-role.bicep
```diff
@@ -0,0 +1,27 @@
...
+param principalId string = ''
+
+@description('Id of the role definition to assign to the targeted principal in the context of the account.')
+param roleDefinitionId string
+
+resource databaseAccount 'Microsoft.DocumentDB/databaseAccounts@2023-04-15' existing = {
+  name: databaseAccountName
+}
+
+resource sqlRoleAssignment 'Microsoft.DocumentDB/databaseAccounts/sqlRoleAssignments@2023-04-15' = {
```
pamelafox, line 17: I assume this wasn't available in the avm module?

app/backend/app.py
fujita-h: Certainly, `app.py` is getting too large, so I think it's good to separate the blueprint. I'll try to restructure it to separate it.

app/frontend/src/api/models.ts
fujita-h: `_ts` is a timestamp record that CosmosDB assigns by default.
We can define and assign a timestamp record defined in our application, but which is better?

https://learn.microsoft.com/en-us/rest/api/cosmos-db/documents

app/frontend/src/api/models.ts
fujita-h: This is a typo. I will fix it.

infra/core/security/documentdb-sql-role.bicep
fujita-h: This SQL Role setting can be created together with the AVM that creates the CosmosDB resource, but because a circular reference occurred, I had no choice but to externalize the SQL Role. Specifically, the following situation arose:

- To create CosmosDB, we need the principalId of the AppService to grant a role that can read and write to the database, so the AppService needs to be deployed first.

- The AppService module needs to store the CosmosDB access destination in an environment variable, so CosmosDB needed to be deployed first.

This required the module that sets the SQL Role to be separated, but I couldn't find an AVM that only sets the SQL Role, so I defined it separately.

app/frontend/src/api/models.ts
fujita-h: Fixed in 0189dfc

app/backend/app.py
fujita-h: Configure separate blueprint in 44bfc40

app/frontend/src/api/models.ts
fujita-h: Updated in 11c4d9f

infra/core/security/documentdb-sql-role.bicep
pamelafox: Okay, yep, that makes sense, thanks for clarifying!

pamelafox:
I'm reviewing now, have made a few cosmetic changes and added a test. I'll add a few more tests tomorrow and test my deploy. Have also asked a Cosmos DB colleague to review.

pamelafox:
Update: I found an issue with the authentication decorator being used for the /items/id get and delete routes and have fixed it. It's working well for me locally now.
I'll add additional pytests tomorrow.

app/backend/chat_history/cosmosdb.py
```diff
@@ -0,0 +1,198 @@
...
+                )
+
+        # If there are no page, StopAsyncIteration is raised
+        except StopAsyncIteration:
+            items = []
+            continuation_token = None
+
+        return jsonify({"items": items, "continuation_token": continuation_token}), 200
+
+    except Exception as error:
```
TheovanKraay, line 102: Consider using more specific cosmos exceptions, e.g. https://learn.microsoft.com/en-us/python/api/azure-cosmos/azure.cosmos.exceptions.cosmoshttpresponseerror?view=azure-python.

app/backend/chat_history/cosmosdb.py
```diff
@@ -0,0 +1,198 @@
...
+            parameters=[dict(name="@entra_oid", value=entra_oid)],
+            max_item_count=count,
+        )
+
+        # set the continuation token for the next page
+        pager = res.by_page(continuation_token)
+
+        # Get the first page, and the continuation token
+        try:
+            page = await pager.__anext__()
```
TheovanKraay, line 81: Consider iterating over the by_page result directly and avoiding explicit calls to await pager.__anext__(), i.e. process each page as soon as it’s available without awaiting when there are no more pages.

app/backend/chat_history/cosmosdb.py
```diff
@@ -0,0 +1,198 @@
...
+chat_history_cosmosdb_bp = Blueprint("chat_history_cosmos", __name__, static_folder="static")
+
+
+@chat_history_cosmosdb_bp.post("/chat_history")
+@authenticated
+async def post_chat_history(auth_claims: Dict[str, Any]):
+    if not current_app.config[CONFIG_CHAT_HISTORY_COSMOS_ENABLED]:
+        return jsonify({"error": "Chat history not enabled"}), 405
+
+    container: ContainerProxy = current_app.config[CONFIG_COSMOS_HISTORY_CONTAINER]
```
TheovanKraay, line 27: This is called in every function and I assume it never changes. Could this be cached or defined globally?

app/backend/chat_history/cosmosdb.py
```diff
@@ -0,0 +1,198 @@
...
+        id = request_json.get("id")
+        answers = request_json.get("answers")
+        title = answers[0][0][:50] + "..." if len(answers[0][0]) > 50 else answers[0][0]
+        timestamp = int(time.time() * 1000)
+
+        await container.upsert_item(
+            {"id": id, "entra_oid": entra_oid, "title": title, "answers": answers, "timestamp": timestamp}
+        )
+
+        return jsonify({}), 201
```
xinlian12, line 46: wonder whether the status code here should be based on the response from `container.upsert_item` - if the item already already exists, then it will be 200 instead of 201

app/backend/chat_history/cosmosdb.py
pamelafox: `current_app.config` is basically our global dict, it's how we access objects that were setup at the beginning of the app start. I don't think there's a performance hit, since it should be O(1) retrieval.

app/backend/chat_history/cosmosdb.py
pamelafox: That's a nice idea! I can't figure out from the SDK return types what in the response would indicate that however, as all the function signatures just say that they return a dict: https://learn.microsoft.com/en-us/python/api/azure-cosmos/azure.cosmos.container.containerproxy?view=azure-python#azure-cosmos-container-containerproxy-upsert-item
And this example doesn't do anything with the dict: https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/cosmos/azure-cosmos/samples/document_management.py#L149-L156
So is there some listing of what key in the dict would indicate it already existing?

app/backend/chat_history/cosmosdb.py
pamelafox: I think you mean to just do:
async for page in pager:
  async for item in page:

As we use similar code for other Azure Python SDKs elsewhere in this repo. That wouldn't give us the continuation token, right, as that would exhaust all the pages?

pamelafox:
@fujita-h I added pytest for 97% coverage of CosmosDB. While adding the tests, I made a few changes:
* 400 instead of 405 for CosmosDB feature disabled (I think 405 is meant to specifically indicate that a different HTTP method would have worked, but none would work in this case)
* 204 instead of 200 for successful item deletion, as that seems to be the web convention
* Removed the check in the "/item_id" paths for "if not item_id" as I don't think there's any possible way to reach that code path- if no item_id is specified, it won't match the route and will 404 instead.

app/backend/chat_history/cosmosdb.py
TheovanKraay: Yes, but you should still be able to get the continuation token. Something like (untested):

```python
        items = []
        async for page in container.query_items(
            query="SELECT c.id, c.entra_oid, c.title, c.timestamp FROM c WHERE c.entra_oid = @entra_oid ORDER BY c.timestamp DESC",
            parameters=[{"name": "@entra_oid", "value": entra_oid}],
            max_item_count=count,
        ).by_page(continuation_token):
            async for item in page:
                items.append(
                    {
                        "id": item.get("id"),
                        "entra_oid": item.get("entra_oid"),
                        "title": item.get("title", "untitled"),
                        "timestamp": item.get("timestamp"),
                    }
                )
            # Update continuation token after processing the page
            continuation_token = page.continuation_token if hasattr(page, "continuation_token") else None

            # Break if no continuation token (i.e., last page)
            if not continuation_token:
                break
```

Just a suggestion. I think its fine as it is :-)

fujita-h:
@pamelafox Thank you for adding the test, I have read and understood your corrections.

app/backend/chat_history/cosmosdb.py
pamelafox: I've taken a look at the dicts for a first-time ID versus a second-time ID, and I can't find any key that'd indicate whether the item already existed:
https://www.diffchecker.com/Fs0ECaUa/
So I don't think we should change the status code currently, as I'm not sure what condition to use, but happy to change in the future if you have suggestions.

pamelafox (approved):
I am approving this PR, as I've both tested manually and with pytest unit tests, and the CosmosDB team has also reviewed it, with no blocking issues. I'll check if @mattgotteiner wants to take a look before we merge.
Thanks @fujita-h for the great work!

pamelafox:
I looked at the Bicep validation error but I think that is an unrelated issue, likely due to a newer version of Bicep that's stricter.

mattgotteiner (approved):
Thanks for your contributions

PR #2021, opened 2024-10-05
Feature: include category 

bnodir:
## Purpose

Adding "Include category" functionality in developer settings, with "All" as a default option. By combining data categorization with the `prepdocs.py` scripts, it lets you filter Azure AI Search based on categories. This functionality allows category-based inquiries and informs the user about the categories present in the search index.

This PR is somehow related to the following past issues:

- #2006 
- #1715
- #1333
- #872 
- #647 
- #493 

## Changes
- Modified approach.py and models.ts to include category logic
- Added translations for category ("All") in en, es, fr, and ja locales
- Updated Ask.tsx and Chat.tsx to handle category
- Updated data ingestion documentation

## Quick insight

_**Here is an example of how categories appear when ingested into the search index:**_

1. "category": "**_BenefitOptions_**"
2. "category": "**_RoleLibrary_**"
3. "category": "**_EmployeeHandbook_**"

image

_**Let's add category keys like this:**_
image

_**Then, in the default "All" category, we get answers from all three categories, like this:**_
image

_**When specifying a particular category, we get generated answers from that category's data:**_
image
image

## Does this introduce a breaking change?

When developers merge from main and run the server, azd up, or azd deploy, will this produce an error?
If you're not sure, try it out on an old environment.

```
[ ] Yes
[x] No
```

## Does this require changes to learn.microsoft.com docs?

This repository is referenced by [this tutorial](https://learn.microsoft.com/azure/developer/python/get-started-app-chat-template)
which includes deployment, settings and usage instructions. If text or screenshot need to change in the tutorial,
check the box below and notify the tutorial author. A Microsoft employee can do this for you if you're an external contributor.

```
[x] Yes
[ ] No
```

## Type of change

```
[ ] Bugfix
[x] Feature
[ ] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[ ] Documentation content changes
[ ] Other... Please describe:
```

## Code quality checklist

See [CONTRIBUTING.md](https://github.com/Azure-Samples/azure-search-openai-demo/blob/main/CONTRIBUTING.md#submit-pr) for more details.

- [ ] The current tests all pass (`python -m pytest`).
- [ ] I added tests that prove my fix is effective or that my feature works
- [ ] I ran `python -m pytest --cov` to verify 100% coverage of added lines
- [ ] I ran `python -m mypy` to check for type errors
- [x] I either used the pre-commit hooks or ran `ruff` and `black` manually on my code.

PR #2111, opened 2024-11-03
Refactor settings into a single component across Chat/Ask

bnodir:
## Purpose

We had a refactoring conversation below. In this PR, I tried that.

> https://github.com/Azure-Samples/azure-search-openai-demo/pull/2021#discussion_r1806283583
> I wish that I had refactored these settings to be in a single component, as now we have duplicated code each time we add a new setting. This is fine for this PR, but eventually I want to try a single Settings.tsx across Chat/Ask.

```
[ ] Yes
[x] No
```

```
[ ] Yes
[x] No
```

## Type of change

```
[ ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## Code quality checklist


app/frontend/src/pages/chat/Chat.tsx
```diff
@@ -257,60 +254,63 @@ const Chat = () => {
...
-    };
-
-    const onUseOidSecurityFilterChange = (_ev?: React.FormEvent<HTMLElement | HTMLInputElement>, checked?: boolean) => {
-        setUseOidSecurityFilter(!!checked);
-    };
-
-    const onUseGroupsSecurityFilterChange = (_ev?: React.FormEvent<HTMLElement | HTMLInputElement>, checked?: boolean) => {
-        setUseGroupsSecurityFilter(!!checked);
+    const handleSettingsChange = (field: string, value: any) => {
+        switch (field) {
```
pamelafox, line 258: I guess there's no way to avoid having this function in both places, aye? I could imagine a function that tried to be very clever and transform the field name into a function name, but that might be too clever and make TypeScript mad.

pamelafox:
Is this a WIP as you're still testing it? You could add playwright tests to click each of the settings and check that the subsequent /chat or /ask network request had the expected options. Am not sure how much that'd slow the tests down to test each of them - you could do just one test that touched all the options, so it'd be one for chat and one for ask.

app/frontend/src/pages/chat/Chat.tsx
bnodir: Yes, I agree with keeping the code as is, avoiding unnecessary complexity.

bnodir:
I have manually tested the changes, not with Playwright, and they were working as expected. Therefore, I removed WIP from the title.

Regarding document updates, at least `data_ingestion.md` needs to be edited with the Settings.tsx file path. Should I wait for #2118 to be merged?

pamelafox:
The techcommunity link breakage is expected due to an ongoing upgrade of that site.
I am just adding a bit more to the playwright test on settings and will push that soon.
No changes needed from you @bnodir, as I also made the docs change.

pamelafox (approved):

Issue #2065, opened 2024-10-22
feature: enhance adlsgen2setupadv.py to scandirs for files and add them


cforce:
with "--scandirs" scan all directories recursively and add the files  to the same directory, without the need to list in files section

new code 
```
  if scandirs:
      await self.scan_and_upload_directories(directories, filesystem_client)
```

                        
all:
```
import argparse
import asyncio
import json
import logging
import os
from typing import Any, Optional

import aiohttp
from azure.core.credentials_async import AsyncTokenCredential
from azure.identity.aio import AzureDeveloperCliCredential
from azure.storage.filedatalake.aio import (
    DataLakeDirectoryClient,
    DataLakeServiceClient,
)

from load_azd_env import load_azd_env

logger = logging.getLogger("scripts")


class AdlsGen2Setup:
    """
    Sets up a Data Lake Storage Gen 2 account with sample data and access control
    """

    def __init__(
        self,
        data_directory: str,
        storage_account_name: str,
        filesystem_name: str,
        security_enabled_groups: bool,
        data_access_control_format: dict[str, Any],
        credentials: AsyncTokenCredential,
    ):
        """
        Initializes the command

        Parameters
        ----------
        data_directory
            Directory where sample files are located
        storage_account_name
            Name of the Data Lake Storage Gen 2 account to use
        filesystem_name
            Name of the container / filesystem in the Data Lake Storage Gen 2 account to use
        security_enabled_groups
            When creating groups in Microsoft Entra, whether or not to make them security enabled
        data_access_control_format
            File describing how to create groups, upload files with access control. See the sampleacls.json for the format of this file
        """
        self.data_directory = data_directory
        self.storage_account_name = storage_account_name
        self.filesystem_name = filesystem_name
        self.credentials = credentials
        self.security_enabled_groups = security_enabled_groups
        self.data_access_control_format = data_access_control_format
        self.graph_headers: Optional[dict[str, str]] = None

    async def run(self, scandirs: bool = False):
        async with self.create_service_client() as service_client:
            logger.info(f"Ensuring {self.filesystem_name} exists...")
            async with service_client.get_file_system_client(self.filesystem_name) as filesystem_client:
                if not await filesystem_client.exists():
                    await filesystem_client.create_file_system()

                logger.info("Creating groups...")
                groups: dict[str, str] = {}
                for group in self.data_access_control_format["groups"]:
                    group_id = await self.create_or_get_group(group)
                    groups[group] = group_id

[...]

PR #2113, opened 2024-11-04
scandir and md5 for adlsgen2setup.py

cforce:
see https://github.com/Azure-Samples/azure-search-openai-demo/issues/2065

### DIFF
--- a/scripts/adlsgen2setup.py
+++ b/scripts/adlsgen2setup.py
@@ -1,8 +1,10 @@
 import argparse
 import asyncio
+import datetime
 import json
 import logging
 import os
+import hashlib
 from typing import Any, Optional
 
 import aiohttp
@@ -56,7 +58,7 @@ class AdlsGen2Setup:
         self.data_access_control_format = data_access_control_format
         self.graph_headers: Optional[dict[str, str]] = None
 
-    async def run(self):
+    async def run(self, scandirs: bool = False):
         async with self.create_service_client() as service_client:
             logger.info(f"Ensuring {self.filesystem_name} exists...")
             async with service_client.get_file_system_client(self.filesystem_name) as filesystem_client:
@@ -80,15 +82,17 @@ class AdlsGen2Setup:
                         )
                         directories[directory] = directory_client
 
+                    logger.info("Uploading scanned files...")
+                    if scandirs and directory != "/":
+                        await self.scan_and_upload_directories(directories, filesystem_client)
+
                     logger.info("Uploading files...")
                     for file, file_info in self.data_access_control_format["files"].items():
                         directory = file_info["directory"]
                         if directory not in directories:
                             logger.error(f"File {file} has unknown directory {directory}, exiting...")
                             return
-                        await self.upload_file(
-                            directory_client=directories[directory], file_path=os.path.join(self.data_directory, file)
-                        )
+                        await self.upload_file(directory_client=directories[directory], file_path=os.path.join(self.data_directory, file))
 
                     logger.info("Setting access control...")
                     for directory, access_control in self.data_access_control_format["directories"].items():
@@ -100,8 +104,7 @@ class AdlsGen2Setup:
                                         f"Directory {directory} has unknown group {group_name} in access control list, exiting"
                                     )
                                     return
-                                await directory_client.update_access_control_recursive(
-                                    acl=f"group:{groups[group_name]}:r-x"
+                                await directory_client.update_access_control_recursive(acl=f"group:{groups[group_name]}:r-x"
                                 )
                         if "oids" in access_control:
                             for oid in access_control["oids"]:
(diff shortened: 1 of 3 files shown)


scripts/adlsgen2setup.py
```diff
@@ -110,15 +113,110 @@ async def run(self):
...
+                    
+        #Return filepath_list        
+        return filepath_list
+
+    async def scan_and_upload_directories(self, directories: dict[str, DataLakeDirectoryClient], filesystem_client):
+        logger.info("Scanning and uploading files from directories recursively...")
+        for directory, directory_client in directories.items():
+            directory_path = os.path.join(self.data_directory, directory)
+            
+             # Überprüfen, ob 'scandir' existiert und auf False gesetzt ist
```
mattgotteiner, line 147: German comment?

scripts/adlsgen2setup.py
```diff
@@ -110,15 +113,110 @@ async def run(self):
...
+        #This for loop uses the os.walk() function to walk through the files and directories
+        #and records the filepaths of the files to a list
+        for root, dirs, files in os.walk(src_filepath):
+            
+            #iterate through the files currently obtained by os.walk() and
+            #create the filepath string for that file and add it to the filepath_list list
+            for file in files:
+                #Checks to see if the root is '.' and changes it to the correct current
+                #working directory by calling os.getcwd(). Otherwise root_path will just be the root variable value.
+                if root == '.':
```
mattgotteiner, line 128: Do we only need to set this once? or does it change for every file

scripts/adlsgen2setup.py
```diff
@@ -110,15 +113,110 @@ async def run(self):
...
+        logger.info("Scanning and uploading files from directories recursively...")
+        for directory, directory_client in directories.items():
+            directory_path = os.path.join(self.data_directory, directory)
+            
+             # Überprüfen, ob 'scandir' existiert und auf False gesetzt ist
+            if not self.data_access_control_format["directories"][directory].get("scandir", True):
+                logger.info(f"Skipping directory {directory} as 'scandir' is set to False")
+                continue
+            
+            groups = self.data_access_control_format["directories"][directory].get("groups", [])
```
mattgotteiner, line 152: groups is not used?

scripts/adlsgen2setup.py
```diff
@@ -110,15 +113,110 @@ async def run(self):
...
         )
 
-    async def upload_file(self, directory_client: DataLakeDirectoryClient, file_path: str):
+    async def calc_md5(self, path: str) -> str:
+        with open(path, "rb") as file:
+            return hashlib.md5(file.read()).hexdigest()
+
+    async def check_md5(self, path: str, md5_hash: str) -> bool:
+        # if filename ends in .md5 skip
+        if path.endswith(".md5"):
```
pamelafox, line 178: We actually want to move the storage of the MD5 into the Blob storage metadata for our main prepdocs strategy, as the local md5 is problematic when you switch environments. Would that be possible with ADSL2? I assume it also has the ability to store metadata, given its a type of Blob storage?
What do you think of that approach, remote MD5?

scripts/adlsgen2setup.py
cforce: Ja! will translate that ;)

scripts/adlsgen2setup.py
cforce: yes, good point

scripts/adlsgen2setup.py
cforce: I’ve already implemented that and will include it with some additional metadata in the next commit. If iread the file directly from the source (local storage or blob/data lake storage) and create a new checksum on the fly, keep it in heap . Then, i can compare it with the persistent checksum on the target storage (datalake)and update the blob only if diff and as well the checksum in the meta.

I’ve also started on preodocs.sh to make use of the persist MD5 on the backend once it’s "injected" from the data lake using the datalake strategy which seems to be unstable for exceptions . (exits the whole loop whi is even more painful if you don't have a resume. md5 will not dupe work ok, but it still  will need to iterate again over all the offset until where the ingestion died abnormally.  For external systems i use queries with timestamps to find a proper point to resume but i have no idea if and how current datalake file walker creates any order  or if is completely random.
Maybe there is a way to walk/iterate  along a query based on a metadata field (i have added) like lastchange date in metadadata? This way i could storage the last i9nkestion date seen in another blob and use it as resume

I’m unsure where best to store this: 
Two options i had in mind

a) In the index, where it would be associated with each chunk, meaning we’d need to locate chunks where there are more than zero results for the fingerprint, updating multiple rows accordingly.

b) In /content (blob storage) for inline rendering of the source document as a citation in the browser. This has the benefit of only requiring one update per change, though we’d still need to locate and remove all matching chunks in the index by primary key for consistency.

Option (a) seems more logical, doesn’t it?

Btw i remember i had trouble with md5 dupes for bigger data amount , so maybe its better to use sha digests?

And even more important: how would above efforts fit into integrated-vectorization?
Will that work with whatever fingerprint being takeover into the index. I assume it requires a new skill, doesn't it?

cforce: Blob storage has embedded md5 already as it seems -

`
from azure.storage.blob import BlobServiceClient

# Your connection string
connection_string = "your_connection_string"

# Your blob container name and blob name
container_name = "your_container_name"
blob_name = "your_blob_name"

# Initialize a BlobServiceClient
blob_service_client = BlobServiceClient.from_connection_string(connection_string)

# Get a reference to the blob
blob_client = blob_service_client.get_blob_client(container=container_name, blob=blob_name)

# Get the properties of the blob
blob_properties = blob_client.get_blob_properties()

# Retrieve the MD5 hash value from the properties
md5_hash = blob_properties.content_settings.content_md5

print(f"MD5 Hash of the blob: {md5_hash}")
`

If that md5 will have same value as a local computed md5 from file is the question
Also as long  as we use as blob name some filename (instead of md5) we still need to rely on that and can't handle rename scenarios. Also we can't  easily with O(1) find out if we have any blob up there for the md5 of the file in our hands. If we would use md5 as blob name it would make things in code much slimmer, but debugging and using the azure portal ui file browser for debugging a bit more cumberstone

pamelafox: Blob service does have its own md5, but it only computes it for small files, so we would need to compute our own hash. Using sha256 also seems fine if we can store that in the blob metadata.

cforce:
@pamelafox @mattgotteiner  still wishes for this PR?

PR #2123, opened 2024-11-06
feat: make-rag-optional-but-default

jeannotdamoiseaux:
## Purpose

Added functionality to optionally disable RAG from the developer settings. The default option remains using all documents in the knowledge base.

This change also addresses hallucination issues with sources observed when RAG was disabled, caused by the master prompt instruction to include sources in the answer. The fix involves injecting the source-related part of the master prompt only when RAG is enabled. Additionally, the supporting content button is now displayed (in the chat) or enabled (in the analysis panel) only when supporting content is available.

```
[ ] Yes
[ X ] No
```

```
[ ] Yes
[ X ] No
```

## Type of change

```
[ ] Bugfix
[ X ] Feature
[ ] Other... Please describe:
```

## Code quality checklist

Note: Many tests are failing with the message "... value does not match the expected value in snapshot tests/snapshots/test_app/...". These failures need to be addressed before merging.


jeannotdamoiseaux:
@pamelafox, could you help me identify where the errors in the test scripts might be coming from? When I deploy the app, everything seems to work as expected on my end.

app/frontend/src/locales/da/translation.json
```diff
@@ -79,7 +79,8 @@
         "retrieveCount": "Hent dette antal søgeresultater:",
         "includeCategory": "Inkludér kategori",
         "includeCategoryOptions": {
-            "all": "Alle"
+            "all": "Alle",
+            "none": "Ingen"
```
pamelafox, line 83: @EMjetrot Can you review this string- "Ingen" for "None" , as in "no categories"?

app/frontend/src/locales/ja/translation.json
```diff
@@ -83,7 +83,8 @@
         "retrieveCount": "ここで指定する検索結果数を取得：",
         "includeCategory": "カテゴリを指定",
         "includeCategoryOptions": {
-            "all": "全て"
+            "all": "全て",
+            "none": "无"
```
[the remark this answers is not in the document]
jeannotdamoiseaux: Thanks for the review! I’ve made the changes.

app/backend/approaches/chatreadretrieveread.py
```diff
@@ -58,7 +58,7 @@ def system_message_chat_conversation(self):
         return """Assistant helps the company employees with their healthcare plan questions, and questions about the employee handbook. Be brief in your answers.
         Answer ONLY with the facts listed in the list of sources below. If there isn't enough information below, say you don't know. Do not generate answers that don't use the sources below. If asking a clarifying question to the user would help, ask the question.
```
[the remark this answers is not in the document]
jeannotdamoiseaux: This is actually the intended behavior. This option provides flexibility to use either the knowledge base exclusively or extend responses with the general knowledge of the LLM.

app/backend/approaches/chatreadretrieveread.py
```diff
@@ -58,7 +58,7 @@ def system_message_chat_conversation(self):
...
         If the question is not in English, answer in the language used in the question.
-        Each source has a name followed by colon and the actual information, always include the source name for each fact you use in the response. Use square brackets to reference the source, for example [info1.txt]. Don't combine sources, list each source separately, for example [info1.txt][info2.pdf].
+        {sources_reference_content}
```
pamelafox, line 61: We've considered moving the prompts into Jinja files. I'm wondering if we should make that change first, as this could read a bit nicer, like
```
{% if should_reference_sources $}
Each source...
{% endif %}
```

Our current method of storing the templates inside multi-line strings is not easy to work with, so I've been hoping to move them at least into separate files, and just hadn't decided about .txt versus Jinja versus prompty files. Jinja is a happy medium. That could be done by you or I in a separate PR. Thoughts?

app/backend/approaches/chatreadretrieveread.py
```diff
@@ -96,106 +96,89 @@ async def run_until_final_call(
...
-                        "type": "object",
-                        "properties": {
-                            "search_query": {
-                                "type": "string",
-                                "description": "Query string to retrieve documents from azure search eg: 'Health care plan'",
-                            }
+        sources_content = []
+        extra_info = {"thoughts": [], 'data_points': []}
+
+        if include_category != "__NONE__":
```
pamelafox, line 110: I think `include_category != "__NONE__"` condition should be executed only once, stored in a variable, and be under a comment that says that __NONE__ is a value passed down by the frontend categories picker. 
I'm nervous about special magical values like __NONE__ so want to make it clear where it came from and minimize places to use it incorrectly.
It could also be a class attribute on Approach, like NO_CATEGORIES

app/backend/approaches/chatreadretrieveread.py
```diff
@@ -96,106 +96,89 @@ async def run_until_final_call(
...
+            tools: List[ChatCompletionToolParam] = [
```
pamelafox, line 111: Hm. It'd be nice to make this change in a way that doesn't cause this large indentation change, since that can make it harder for developers to merge in new changes..but I think that's not possible, right? Just musing out loud.

app/backend/approaches/chatreadretrieveread.py
```diff
@@ -96,106 +96,89 @@ async def run_until_final_call(
...
+                minimum_reranker_score,
+            )
 
-        data_points = {"text": sources_content}
+            sources_content = self.get_sources_content(results, use_semantic_captions, use_image_citation=False)
+            if sources_content:
+                extra_info["data_points"] = {"text": sources_content}
 
-        extra_info = {
-            "data_points": data_points,
```
pamelafox, line 197: You seem to have lost the data_points from thought process, please bring them back.

app/frontend/src/components/Settings/Settings.tsx
```diff
@@ -198,7 +198,8 @@ export const Settings = ({
                 onChange={(_ev?: React.FormEvent<HTMLElement | HTMLInputElement>, option?: IDropdownOption) => onChange("includeCategory", option?.key || "")}
                 aria-labelledby={includeCategoryId}
                 options={[
-                    { key: "", text: t("labels.includeCategoryOptions.all") }
+                    { key: "", text: t("labels.includeCategoryOptions.all") },
+                    { key: "__NONE__", text: t("labels.includeCategoryOptions.none") }
```
pamelafox, line 202: I'm not 100% sure that this will be intuitive to folks as the way to go into "non-RAG" mode. Before looking at the PR, I was expecting a top-level checkbox like "Use sources for answers". But it's also nice to avoid adding even more things to settings. Do any other folks have thoughts?

pamelafox:
@jeannotdamoiseaux The snapshot tests are failing due to the addition of a newline from the template change. They can be updated with `pytest --snapshot-update` and the differences reviewed in GitHub.

app/backend/approaches/chatreadretrieveread.py
jeannotdamoiseaux: Agreed. I'll create a separate PR for this first.

app/backend/approaches/chatreadretrieveread.py
jeannotdamoiseaux: This behavior is intentional, as there are no data points when RAG is disabled. However, the solution could be more elegant than simply overwriting the `extra_info` variable.

app/frontend/src/components/Settings/Settings.tsx
jeannotdamoiseaux: Given my use case, this approach is preferable. We will move this setting out of the developer settings and make it available to all users. This way, they will have a single dropdown to select a relevant subset of the documents in the index.

jeannotdamoiseaux:
> @jeannotdamoiseaux The snapshot tests are failing due to the addition of a newline from the template change. They can be updated with `pytest --snapshot-update` and the differences reviewed in GitHub.

I'm new to this type of testing and not entirely sure what steps to follow here. Could you clarify what exactly needs to be done?

app/frontend/src/components/Settings/Settings.tsx
jeannotdamoiseaux: No, in this repository, it will remain a developer setting. However, in our clone, we plan to migrate the dropdown to the UI, making it accessible to all users.

By the way, I’m curious to know which organization you're working with, as we also need to comply with regulations like DPIAs as a governmental entity.

pamelafox: Good point by @EMjetrot, this feature is significant enough that there should be an option to enable/disable, as many organizations won't want to enable a general purpose chat. Given that, and concerns about unintuitive nature of the "None" dropdown, I'd vote for either a separate developer setting or a setting in the user space. (We don't have an obvious place for it, probably next to User uploads, but it's getting crowded. Perhaps you have ideas)

jeannotdamoiseaux: My preference would be to remove the text next to the buttons, move the 'Upload File' button to the left of the chat input container, change the 'Remove Chat' button to 'New Chat,' and place it in the top-left (similar to the ChatGPT interface). The RAG switch could either be placed to the left of the question input container or in the top-right.

pamelafox: * Re "upload file": I'd be slightly concerned that folks would think the file was *only* for that particular question, since that's often a feature of chat interfaces, versus persistent. 
* "Remove chat" -> "New chat" seems like a good change, given that "Remove chat" doesn't actually remove from chat history for folks using that. That's also the phrasing used by GitHub Copilot Chat.
* Placing in top-left to match ChatGPT could work. I don't feel strongly either way.
* I do like when there's text next to icons, as I can be a bit "icon-blind" at times, but I realize that's not always feasible. It's fine as long as we keep accessible tooltips.

cc @zedhaque as I think he's thought through this for the mobile-optimized design

app/backend/approaches/chatreadretrieveread.py
jeannotdamoiseaux: Implemented in #2164

PR #2134, opened 2024-11-10
checksum support , improve control with new cm line swicthes

cforce:
This PR leverages the MD5 checksum available through local, blob, or Data Lake storage metadata fields, based on the discussion in [https://github.com/Azure\-Samples/azure\-search\-openai\-demo/pull/2113](https://github.com/Azure-Samples/azure-search-openai-demo/pull/2113).

By default, the PR introduces a check during ingestion (for both local and Azure Data Lake Storage sources) to see if each file already exists in the target storage. The MD5 checksum is saved in the index as a verification method, enabling a lookup that ensures at least one record with the same category/filename is found. This also permits the same file to be copied to multiple categories if needed.

Additionally, a new command-line parameter, `--folder <name>`, allows specifying a dedicated subfolder within the data folder or Data Lake container for each script run. This enables isolated runs with explicitly controlled parameters and a specific subset of files.

The new `--enablevision true/false` option allows per-folder and per-category control over whether vision embeddings should be generated. This option is beneficial when dealing with folders containing predominantly heavy image content, providing control over resource use.

Example usage:

```bash
./.venv/bin/python ./app/backend/prepdocs.py './data/createindex_dummy/*' --category newtest --folder test --enablevision false --verbose
```

The PR also improves progress reporting by displaying the number of files processed and providing updates every 10 files. This enhances visibility into the processing speed, which is particularly useful when working with partially processed data. Additionally, the log level of some messages has been adjusted from INFO to DEBUG to reduce console noise and retain progress updates longer.

Finally, MD5 checksum comparisons are introduced at various stages: a top-level check (to avoid redundant processing), and checks for blobs, text, and image embeddings. A flag to bypass the top-level check for known MD5 values in the index may be introduced later (e.g., `--ignore_checksum`).

* * *


app/backend/prepdocslib/blobmanager.py
```diff
@@ -45,29 +45,63 @@ def __init__(
...
+        md5_check : int = 0 # 0= not done, 1, positive,. 2 negative
+        # Get existing blob properties
+        blob_properties = await blob_client.get_blob_properties()
+        blob_metadata = blob_properties.metadata
+        
+        # Check if the md5 values are the same
+        file_md5 = file.metadata.get('md5')
+        blob_md5 = blob_metadata.get('md5')
+        
+        # Remove md5 from file metadata if it matches the blob metadata
```
pamelafox, line 66: This comment doesn't seem to describe the next line?

app/backend/prepdocslib/blobmanager.py
```diff
@@ -45,29 +45,63 @@ def __init__(
...
+    #async def upload_blob(self, file: File, container_client:ContainerClient) -> Optional[List[str]]:
+        
+    async def _create_new_blob(self, file: File, container_client:ContainerClient) -> BlobClient:
+        with open(file.content.name, "rb") as reopened_file:
+                                blob_name = BlobManager.blob_name_from_file_name(file.content.name)
+                                logger.info("Uploading blob for whole file -> %s", blob_name)
+                                return await container_client.upload_blob(blob_name, reopened_file, overwrite=True, metadata=file.metadata)
+
+    async def _file_blob_update_needed(self, blob_client: BlobClient, file : File) -> bool:
+        md5_check : int = 0 # 0= not done, 1, positive,. 2 negative
```
pamelafox, line 57: Did you end up not using this variable?

app/backend/prepdocslib/blobmanager.py
```diff
@@ -45,29 +45,63 @@ def __init__(
         self.subscriptionId = subscriptionId
         self.user_delegation_key: Optional[UserDelegationKey] = None
 
+    #async def upload_blob(self, file: File, container_client:ContainerClient) -> Optional[List[str]]:
```
pamelafox, line 48: This line can be rm'ed now?

app/backend/prepdocslib/blobmanager.py
```diff
@@ -45,29 +45,63 @@ def __init__(
...
         ) as service_client, service_client.get_container_client(self.container) as container_client:
             if not await container_client.exists():
                 await container_client.create_container()
-
-            # Re-open and upload the original file
+            
+             # Re-open and upload the original file
+            md5_check : int = 0 # 0= not done, 1, positive,. 2 negative
+            
+            # upload  the file local storage zu azure  storage
```
pamelafox, line 82:
```suggestion
            # upload the file from local storage to Azure  storage
```

app/backend/prepdocslib/blobmanager.py
```diff
@@ -45,29 +45,63 @@ def __init__(
...
+                    if self._blob_update_needed(blob_client, file):
+                        md5_check = 2
+                        # Upload the file with the updated metadata
+                        with open(file.content.name, "rb") as data:
+                            await blob_client.upload_blob(data, overwrite=True, metadata=file.metadata)
+                    else:
+                        md5_check = 1                        
+                file.url = blob_client.url
+                
+            if md5_check!=1 and self.store_page_images:
```
pamelafox, line 99: Please install the pre-commit (see CONTRIBUTING guide) as it'll fix Python formatting issues for you, like the whitespace around operators. Thanks!

app/backend/prepdocslib/blobmanager.py
```diff
@@ -93,7 +131,20 @@ async def upload_pdf_blob_images(
 
         for i in range(page_count):
             blob_name = BlobManager.blob_image_name_from_file_page(file.content.name, i)
-            logger.info("Converting page %s to image and uploading -> %s", i, blob_name)
+
+            blob_client = container_client.get_blob_client(blob_name)
+            if await blob_client.exists():
```
pamelafox, line 136: Do we need this additional check here, given the check that happens in the function that calls this code?

app/backend/prepdocslib/blobmanager.py
```diff
@@ -120,15 +171,15 @@ async def upload_pdf_blob_images(
             new_img.save(output, format="PNG")
             output.seek(0)
 
-            blob_client = await container_client.upload_blob(blob_name, output, overwrite=True)
+            await blob_client.upload_blob(data=output, overwrite=True, metadata=file.metadata)
             if not self.user_delegation_key:
                 self.user_delegation_key = await service_client.get_user_delegation_key(start_time, expiry_time)
 
-            if blob_client.account_name is not None:
+            if container_client.account_name is not None:
```
pamelafox, line 178: Am curious why the need to go from blob_client to container_client here? Does it matter?

app/backend/prepdocslib/filestrategy.py
```diff
@@ -22,11 +22,11 @@ async def parse_file(
     if processor is None:
         logger.info("Skipping '%s', no parser found.", file.filename())
         return []
-    logger.info("Ingesting '%s'", file.filename())
+    logger.debug("Ingesting '%s'", file.filename())
```
pamelafox, line 25: Looks like you changed the levels to debug, does the current output feel too verbose? I find it helpful.

app/backend/prepdocslib/filestrategy.py
```diff
@@ -77,29 +79,67 @@ async def run(self):
...
+                processed_count += 1
+                if processed_count % 10 == 0:
+                    remaining = max(doc_count - processed_count, 1)
+                    logger.info("%s removed, %s documents remaining", processed_count, remaining)
+
         elif self.document_action == DocumentAction.RemoveAll:
             await self.blob_manager.remove_blob()
             await search_manager.remove_content()
 
+    async def process_file(self, file, search_manager):
```
pamelafox, line 124: These two functions seem to be unused currently- perhaps your thought was to refactor the code above to call these two functions. I'll remove them for now to make the diff smaller.

app/backend/prepdocslib/listfilestrategy.py
```diff
@@ -168,10 +195,33 @@ async def list(self) -> AsyncGenerator[File, None]:
...
                     except Exception as file_delete_exception:
                         logger.error(f"\tGot an error while deleting {temp_file_path} -> {file_delete_exception}")
+
+    def count_docs(self) -> int:
+        """
+        Return the number of blobs in the specified folder within the Azure Blob Storage container.
+        """
+
+        # Create a BlobServiceClient using account URL and credentials
+        service_client = BlobServiceClient(
```
pamelafox, line 215: Can you use BlobServiceClient to interact with a DataLake Storage account? That's what you seem to be doing here, but I didn't realize that was possible.

app/backend/prepdocslib/blobmanager.py
```diff
@@ -45,29 +47,65 @@ def __init__(
...
+            logger.info("Uploading blob for whole file -> %s", blob_name)
+            return await container_client.upload_blob(blob_name, reopened_file, overwrite=True, metadata=file.metadata)
+
+    async def _file_blob_update_needed(self, blob_client: BlobClient, file: File) -> bool:
+        # Get existing blob properties
+        blob_properties = await blob_client.get_blob_properties()
+        blob_metadata = blob_properties.metadata
+
+        # Check if the md5 values are the same
+        file_md5 = file.metadata.get("md5")
```
pamelafox, line 62: I don't see logic in the main prepdocs for setting the md5 of the local file metadata, I only see that in ADLS2. How does this work when not using the ADLS2 strategy?

app/backend/prepdocslib/searchmanager.py
```diff
@@ -70,92 +72,95 @@ async def create_index(self, vectorizers: Optional[List[VectorSearchVectorizer]]
...
-                            key=True,
-                            sortable=True,
-                            filterable=True,
-                            facetable=True,
-                            analyzer_name="keyword",
-                        )
-                    ),
-                    SearchableField(
-                        name="content",
+            fields = [
```
pamelafox, line 75: How necessary is it to store md5 as a field in the search index? I was thinking we'd also need to store it in the blob/file metadata? @mattgotteiner What do you think of storing md5 in search index, have you seen this approach before?

pamelafox:
I've gone in and made the changes I recommended, but I have some higher-level questions about the approach, which I've left as comments.

PR #2136, opened 2024-11-10
let env var overrule .dot env setting

cforce:
Overriding the env vars in then vscode launch.,json ""env": .." section does not work because of load_azd_env.py
always gives the dot env presence. Therefore i changed load_dotenv(env_file_path, override=True) to
load_dotenv(env_file_path, override=False) to let the env var win in favor to dot env file setting

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -14,7 +14,9 @@
             "env": {
                 "QUART_APP": "main:app",
                 "QUART_ENV": "development",
-                "QUART_DEBUG": "0"
+                "QUART_DEBUG": "0",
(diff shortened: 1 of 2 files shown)


pamelafox:
Very interesting. I definitely understand why you made this change, but I worry then we'll end up with devs in the opposite situation: they accidentally have an environment variable in their global environment with the same name as the azd environment variable. I've run into that in the past, and that's why I do the override.
What if we have an environment variable for allowing override, and you could set that in launch.json when needed? Would that work?

cforce:
Yes, that shall be sufficient

pamelafox:
I've added the environment variable, along with an example in launch.json and comment about how to enable the mode you desire.

PR #2197, opened 2024-11-27
Set AZURE_USE_AUTHENTICATION correctly

rgrey:
## Purpose

Fix the setting of the internal AZURE_USE_AUTHENTICATION variable from the correct environment variable equivalent.

```
[ ] Yes
[x] No
```

```
[ ] Yes
[x] No
```

## Type of change

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## Code quality checklist


scripts/prepdocs.sh
```diff
@@ -4,7 +4,7 @@
 
 echo 'Running "prepdocs.py"'
 
-AZURE_USE_AUTHENTICATION=$(azd env get-value AZURE_PUBLIC_NETWORK_ACCESS)
+AZURE_USE_AUTHENTICATION=$(azd env get-value AZURE_USE_AUTHENTICATION)
 AZURE_PUBLIC_NETWORK_ACCESS=$(azd env get-value AZURE_PUBLIC_NETWORK_ACCESS)
 
 if [ -n "$AZURE_USE_AUTHENTICATION" ] && [ "$AZURE_PUBLIC_NETWORK_ACCESS" = "Disabled" ]; then
```
pamelafox, line 10: Ah, good find! I think actually we can remove line 7, and change line 10 to:

```
if [ -n "$AZURE_PUBLIC_NETWORK_ACCESS" ] && [ "$AZURE_PUBLIC_NETWORK_ACCESS" = "Disabled" ]; then
```

That's how it was in the change before my recent refactor:
https://github.com/Azure-Samples/azure-search-openai-demo/blame/106b52be18b5a2a13d03a76a64cb2342d0758cef/scripts/prepdocs.sh

We also should have the same logic here:
https://github.com/Azure-Samples/azure-search-openai-demo/blob/main/scripts/prepdocs.ps1
Here's how it used to look like:
https://github.com/Azure-Samples/azure-search-openai-demo/blob/1837d5f42ab4e9ba1d1c4785b8f00bb91e7b0a17/scripts/prepdocs.ps1#L12

Let me know if you can take on those changes or if you want me to commandeer the PR.
Thanks so much for noticing!

rgrey: If I read you correctly, have removed and modified accordingly :-)

Issue #2071, opened 2024-10-23
submit question  button is disabled


cforce:
The Submit question  button "mouse over  is disabled" -always . No matter  with or without text.
It still works with Return key to send a message
Using latest Chrome

pamelafox:
I'm unable to replicate this, here's what I see in Edge:
![Screenshot 2024-10-25 at 4 02 41 PM](https://github.com/user-attachments/assets/bbd4fb98-a041-4795-8a6e-19f4a8565849)

Is this with user login enabled or other features?

cforce:
AZURE_ENABLE_GLOBAL_DOCUMENT_ACCESS="true"
AZURE_ENABLE_UNAUTHENTICATED_ACCESS="false"
AZURE_ENFORCE_ACCESS_CONTROL="true"

jeannotdamoiseaux:
This doesn't seem logical. @pamelafox, should I create a quick fix with a PR?

pamelafox:
Yes please, thank you @jeannotdamoiseaux

PR #2202, opened 2024-11-29
[Bugfix] The "send" button for chat should be enabled for logged in users

jeannotdamoiseaux:
## Purpose

Currently, the send button is disabled when login is required, even if the user is already logged in (see #2071). While chatting can still be done via the Enter key, the disabled button is redundant since the text input box is already blocked when the user is not logged in. This unnecessary check has been removed to address this bug.

```
[ ] Yes
[ X ] No
```

```
[ ] Yes
[ X ] No
```

## Type of change

```
[ X ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## Code quality checklist

### DIFF
--- a/app/frontend/src/components/QuestionInput/QuestionInput.tsx
+++ b/app/frontend/src/components/QuestionInput/QuestionInput.tsx
@@ -65,7 +65,7 @@ export const QuestionInput = ({ onSend, disabled, placeholder, clearOnSend, init
     };
 
     const disableRequiredAccessControl = requireLogin && !loggedIn;
-    const sendQuestionDisabled = disabled || !question.trim();
+    const sendQuestionDisabled = disabled || !question.trim() || disableRequiredAccessControl;
 
     if (disableRequiredAccessControl) {
         placeholder = "Please login to continue...";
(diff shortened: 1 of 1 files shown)


app/frontend/src/components/QuestionInput/QuestionInput.tsx
```diff
@@ -65,7 +65,7 @@ export const QuestionInput = ({ onSend, disabled, placeholder, clearOnSend, init
     };
 
     const disableRequiredAccessControl = requireLogin && !loggedIn;
-    const sendQuestionDisabled = disabled || !question.trim() || requireLogin;
+    const sendQuestionDisabled = disabled || !question.trim();
```
pamelafox, line 68: Hm, I wonder if it would be more prudent to have `|| disableRequiredAccessControl`? Would that also resolve the bug, since it should only be true when not logged in.

app/frontend/src/components/QuestionInput/QuestionInput.tsx
```diff
@@ -65,7 +65,7 @@ export const QuestionInput = ({ onSend, disabled, placeholder, clearOnSend, init
     };
 
     const disableRequiredAccessControl = requireLogin && !loggedIn;
-    const sendQuestionDisabled = disabled || !question.trim() || requireLogin;
+    const sendQuestionDisabled = disabled || !question.trim() || disableRequiredAccessControl;
```
pamelafox, line 68: I added "disableRequiredAccessControl" to this boolean condition, to align the textfield and button disabled state.

PR #2209, opened 2024-12-02
Move check for AZURE_PUBLIC_NETWORK_ACCESS

pamelafox:
## Purpose

This PR moves the check for AZURE_PUBLIC_NETWORK_ACCESS into prepdocs.py itself, to avoid having duplicate (and currently incorrect) logic in the shell scripts.
The drawback is that it will take another second before it realizes it shouldn't run prepdocs, but in this case, since the majority of developers do not have AZURE_PUBLIC_NETWORK_ACCESS disabled, that seems fine.

This PR is based on https://github.com/Azure-Samples/azure-search-openai-demo/pull/2197 which I wasn't able to update for some reason, so I made a new PR.

```
[ ] Yes
[X] No
```

```
[ ] Yes
[X] No
```

## Type of change

```
[X] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## Code quality checklist


app/backend/app.py
```diff
@@ -102,7 +102,6 @@
 
 @bp.route("/")
 async def index():
-    # TODO: use msal loginRedirect on a blankish page
```
pamelafox, line 105: Meant to remove this TODO in the built-in auth PR but forgot to push the removal commit before merging. This TODO isn't actually needed, since built-in auth works.

app/backend/prepdocs.py
```diff
@@ -241,8 +241,7 @@ async def main(strategy: Strategy, setup_index: bool = True):
 
 if __name__ == "__main__":
     parser = argparse.ArgumentParser(
-        description="Prepare documents by extracting content from PDFs, splitting content into sections, uploading to blob storage, and indexing in a search index.",
-        epilog="Example: prepdocs.py '.\\data\*' -v",
```
pamelafox, line 245: This epilog causes a SyntaxWarning in Python 3.12, and also isnt entirely accurate. The docs for data ingestion are the most accurate.

scripts/prepdocs.sh
```diff
@@ -4,14 +4,6 @@
 
 echo 'Running "prepdocs.py"'
 
-AZURE_USE_AUTHENTICATION=$(azd env get-value AZURE_PUBLIC_NETWORK_ACCESS)
```
pamelafox, line 7: prepdocs.ps1 was already missing this check, so we just remove it from sh and add it to prepdocs.py.

mattgotteiner (approved):

PR #2233, opened 2024-12-13
WIP: Bring evaluation more tightly into the repo

pamelafox:
## Purpose

This PR makes it easier to run evaluations by bringing the evaluation SDK and tools directly into the repo. These scripts still use the [ai-rag-chat-evaluator](https://github.com/Azure-Samples/ai-rag-chat-evaluator) for its custom evaluation metrics and evaluation review CLI tools, but I've moved the ground truth generation directly into the repo, as I've found that it is often very specific to the needs of the repo.

This PR currently uses the new Simulator from azure-ai-evaluation, but I might try RAGAS as well. It's hard to get good ground truth data that sufficiently covers the whole knowledge base.

```
[ ] Yes
[X] No
```

```
[X] Yes - We'd update the evaluation tutorial!
[ ] No
```

## Type of change

```
[ ] Bugfix
[X] Feature
[ ] Other... Please describe:
```

## Code quality checklist


docs/evaluation.md
```diff
@@ -0,0 +1,84 @@
...
+## Deploy a GPT-4 model
+
+
+1. Run this command to tell `azd` to deploy a GPT-4 level model for evaluation:
+
+    ```shell
+    azd env set USE_EVAL true
+    ```
+
+2. Set the capacity to the highest possible value to ensure that the evaluation runs quickly.
```
mattgotteiner, line 21: should we state that it's expected that evals take a longer time to run in the documentation?

docs/evaluation.md
```diff
@@ -0,0 +1,84 @@
...
+    ```shell
+    azd provision
+    ```
+
+## Setup the evaluation environment
+
+Install all the dependencies for the evaluation script by running the following command:
+
+```bash
+pip install -r evals/requirements.txt
```
mattgotteiner, line 40: Why do we create a 2nd environment? Because we don't need it in production?

evals/generate_ground_truth.py
```diff
@@ -0,0 +1,143 @@
...
+    context: Optional[Dict[str, Any]] = None,
+):
+    messages_list = messages["messages"]
+    latest_message = messages_list[-1]
+    query = latest_message["content"]
+    headers = {"Content-Type": "application/json"}
+    body = {
+        "messages": [{"content": query, "role": "user"}],
+        "stream": stream,
+        "context": {
```
mattgotteiner, line 35: Does this need to be read from config?

evals/ground_truth.jsonl
```diff
@@ -0,0 +1,24 @@
+{"question": "Could you provide details on the insurance coverage Contoso offers for work-related injuries? Additionally, I'd like to know more about the overall benefits package available to employees.", "truth": "Contoso offers Workers' Compensation Insurance coverage through Northwind Health for work-related injuries or illnesses [Northwind_Standard_Benefits_Details.pdf#page=101][Northwind_Health_Plus_Benefits_Details.pdf#page=106]. This coverage includes medical care, wage replacement, vocational rehabilitation, and death benefits [Northwind_Standard_Benefits_Details.pdf#page=101]. Employees should report any injuries or illnesses to their supervisor as soon as possible and ensure that the appropriate paperwork is on file to be eligible for Workers' Compensation Insurance coverage [Northwind_Health_Plus_Benefits_Details.pdf#page=107]. As for the overall benefits package, please provide more specific details or refer to the employee handbook."}
```
mattgotteiner, line 1: How did you generate this?

docs/evaluation.md
pamelafox: The requirements can be a bit hairy to install, so I don't want to put them in the main requirements. You dont need a second venv though.
I won't merge this until the eval SDK removes the promptflow dependency, and that'll make the install even smoother.

evals/generate_ground_truth.py
pamelafox: Great question! This is what is very awkward about how our app currently works: the frontend is the source of truth for the parameter values. If you don't send anything to the API, then you may very well end up with a set of parameters that do *not* match what your app uses. 
Ideally, the backend would have its own default set of values, send those to the frontend, and the frontend settings UI would update based on those. Then you could send a bare request to /chat and have it inherit the right default values.
For now, you have to copy/paste what gets sent over the wire from the frontend, which is awkward.

Oh, but also: I'm going to remove this file entirely, as I didn't find this to be a useful way to generate the ground truth. RAGAS is easier.

This discussion is still relevant for the evaluation flow though, as you currently have to copy/paste parameters into evaluate_config.json

evals/ground_truth.jsonl
pamelafox: This looks like the ground truth generated from the azure-ai-eval simulator. I didn't think this was as useful, so as long as RAGAS looks good, I'll remove this file and the non-RAGAS generation.

docs/evaluation.md
pamelafox: Yep sounds good, can do

evals/requirements.txt
```diff
@@ -0,0 +1,4 @@
+dotenv-azd==0.2.0
+azure-ai-evaluation==1.0.1
+rich
+git+https://github.com/Azure-Samples/ai-rag-chat-evaluator
```
pamelafox, line 4: TODO: Add a version/branch

evals/generate_ground_truth.py
```diff
@@ -0,0 +1,143 @@
...
+    load_azd_env()
+
+    azure_credential = get_azure_credential()
+
+    with open(CURRENT_DIR / "generate_config.json") as f:
+        generate_config = json.load(f)
+        simulations = generate_config["simulations"]
+        num_per_task = generate_config.get("num_per_task", 2)
+
+    asyncio.run(generate_ground_truth(azure_credential, simulations, num_per_task))
```
pamelafox, line 143: TODO: gist

evals/generate_ground_truth.py
pamelafox: Moving this file into a gist:
https://gist.github.com/pamelafox/0d2e8e98dd8fc3788b6a0a7ff2ccb8a1
And deleting this file.

Filed issue about moving API parameter defaults into the backend:
https://github.com/Azure-Samples/azure-search-openai-demo/issues/2335

evals/requirements.txt
pamelafox: Adding tag

docs/evaluation.md
pamelafox: Unfortunately, azure-ai-evaluation still brings in promptflow, so I'm keeping the requirements separate.

PR #2284, opened 2025-01-18
Seed shall be an Int

cforce:
seed is an optional parameter, which can be set to an integer or null.


pamelafox:
Hm, I thought I'd played with this, and that setting this to an int forces the seed to be a non-null value, but I'll check out your branch and see for myself. Thanks for making the PR!

PR #2312, opened 2025-01-27
Improve schema of CosmosDB chat history to handle long conversations

pamelafox:
## Purpose

The original approach to CosmosDB chat history was to store both the session information and all the chat messages inside the same document. However, we received feedback from the CosmosDB team that such an architecture is risky because a single session could have so many messages that it exceeds the 2MB document size limit. The safer approach is to store the sessions and messages separately, in one document each.

This PR migrates to that schema, and is based off the same schema used in this Cosmos DB sample:
https://github.com/AzureCosmosDB/cosmosdb-nosql-copilot

It should be just as performant, as we're using hierarchical partition keys (user_id/session_id) for all the operations.

However, this PR is a breaking change: if a developer already had this deployed with history and deployed the new version, users would lose their current history and start over (in a new container). If this is an issue for some folks, we could write a migration script to migrate over the data from the old container to the new one. I'm not sure how many folks are using history in production yet, since it's a fairly new feature.

```
[X] Yes
[ ] No
```

```
[ ] Yes
[X] No
```

## Type of change

```
[X] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## Code quality checklist


app/backend/chat_history/cosmosdb.py
```diff
@@ -34,23 +34,50 @@ async def post_chat_history(auth_claims: Dict[str, Any]):
...
+        message_pairs = request_json.get("answers")
+        first_question = message_pairs[0][0]
+        title = first_question + "..." if len(first_question) > 50 else first_question
         timestamp = int(time.time() * 1000)
 
-        await container.upsert_item(
-            {"id": id, "entra_oid": entra_oid, "title": title, "answers": answers, "timestamp": timestamp}
-        )
+        # Insert the session item:
+        session_item = {
```
mattgotteiner, line 44: Should we introduce  a version field in case we want to keep migrating schemas in the  future?

app/backend/chat_history/cosmosdb.py
```diff
@@ -34,23 +34,50 @@ async def post_chat_history(auth_claims: Dict[str, Any]):
...
+            "title": title,
+            "timestamp": timestamp,
+        }
+
+        message_pair_items = []
+        # Now insert a message item for each question/response pair:
+        for ind, message_pair in enumerate(message_pairs):
+            message_pair_items.append(
+                {
+                    "id": f"{session_id}-{ind}",
```
mattgotteiner, line 58: Rather than formatting this as a string - should we store the message ordinal as a separate field to enable sorting by history?

app/backend/chat_history/cosmosdb.py
```diff
@@ -150,10 +192,22 @@ async def delete_chat_history_session(auth_claims: Dict[str, Any], item_id: str)
         return jsonify({"error": "User OID not found"}), 401
 
     try:
-        await container.delete_item(item=item_id, partition_key=entra_oid)
-        return jsonify({}), 204
+        res = container.query_items(
+            query="SELECT * FROM c WHERE c.session_id = @session_id",
```
mattgotteiner, line 196: Do we need to select * if we are just going to delete all the items? can we just get the id?

infra/main.bicep
```diff
@@ -799,26 +799,31 @@ module cosmosDb 'br/public:avm/res/document-db/database-account:0.6.1' = if (use
         containers: [
           {
             name: chatHistoryContainerName
+            kind: 'MultiHash'
```
mattgotteiner, line 802: This probably will not be last revision to chat history, should we introduce a env var for version and set it on deployment so we know we need to run a migration?

mattgotteiner (approved):

app/backend/chat_history/cosmosdb.py
pamelafox: Ah that's a clever idea, I can add that!
Note that I did change the container name in the Bicep itself, so that the old ones will stick around, but I like the version field too.

app/backend/chat_history/cosmosdb.py
pamelafox: I think I need an ID regardless, for CosmosDB, right? So that was my attempt at a unique ID.
But yes I can also add as a separate field. @markjbrown says that the default sorting is based on insertion order, but I like the clarity of having a sequence marker too. Note that it'd be null for "session" types.

app/backend/chat_history/cosmosdb.py
pamelafox: Good point! I copied from above. I'll change to just id.

infra/main.bicep
pamelafox: Currently we would know based off the container name. I could make that more explicitly look like a version like "chat-history-v1". I think that for migrating, you'd first run the provision, then run a migration script from the local machine, then do a full up/deploy.

pamelafox: Or I can add this:

```
output AZURE_CHAT_HISTORY_VERSION string = 'cosmosdb-v2'
```

I don't actually need to set it on the app, as the app code only supports the v2 version, but a migration script could check that version.

pamelafox: Actually we should set it on the app if we want to store it in the items as well. Adding.

app/backend/chat_history/cosmosdb.py
pamelafox: Added version field, based off what's in the env variable

graememel-kap:
This is a great change, any chance you can publish a migration script? We've been running this is production for the last few months and have roughly 3000 chat history items in our cosmos container.

pamelafox:
@graememel-kap Sure, we can work on one this week, thanks for letting us know you need it.

markjbrown:
@pamelafox let me know if you have questions on this. The approach I generally recommend is to create an Azure Function with a Cosmos Trigger and Output binding and set the start time to beginning and have it iterate through the collection. You can also create a simple Jupyter Notebook or python script and implement Change Feed Processor natively too. 

For in-place updates you need a way to prevent the trigger from reentry. This sample shows the Azure Function route and how to prevent reentry. https://github.com/AzureCosmosDB/cosmos-embeddings-generator

pamelafox:
We opted to write a simple script (and test), in this PR:
https://github.com/Azure-Samples/azure-search-openai-demo/pull/2442
We're avoiding adding a function since this should only need to be run 1-2 times, and adding a function to the azd template would add a lot of complexity.
@graememel-kap Let me know if this meets your needs.

graememel-kap:
@pamelafox Thanks! We'll test this out next week in DEV.

eg-johnny:
This is a great change!! However, we had already been using the cosmos history before this change was made.  Was a migration script ever created for moving data from the v1 to v2 schema?  We definitely could use this.

PR #2424, opened 2025-03-12
Update searchmanager.py

saurabhbansal123:
Integrated vectorizer needs a title field in the index which is not defined while creating the initial index..

## Purpose

Fixes the index when integrated vectorization is enabled.

```
[ ] Yes
[X ] No
```

```
[ ] Yes
[X ] No
```

## Type of change

```
[X ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## Code quality checklist

### DIFF
--- a/app/backend/prepdocslib/searchmanager.py
+++ b/app/backend/prepdocslib/searchmanager.py
@@ -141,6 +141,8 @@ class SearchManager:
                 if self.use_int_vectorization:
                     logger.info("Including parent_id field in new index %s", self.search_info.index_name)
                     fields.append(SearchableField(name="parent_id", type="Edm.String", filterable=True))
+                    logger.info("Including title field in new index %s", self.search_info.index_name)
+                    fields.append(SimpleField(name="title", type="Edm.String", searchable=True, retrievable=True))
                 if self.search_images:
                     logger.info("Including imageEmbedding field in new index %s", self.search_info.index_name)
                     fields.append(


pamelafox:
Thanks for the PR, I've discussed with @mattgotteiner. He says that the title should not be strictly required for integrated vectorization, but that some developers may want it. If we support title, then we should also add code in prepdocslib that will set the title field. Could you make that change?

We may also want to flag behind an environment variable.
@mattgotteiner What is the drawback to adding title to all search indexes, index size?

saurabhbansal123:
In the code which I have updated, the title field is only added to the index if the user is using integrated vectorization. I have updated /app/backend/prepdocslib/searchmanager.py file 

`
if self.use_int_vectorization:
                    logger.info("Including parent_id field in new index %s", self.search_info.index_name)
                    fields.append(SearchableField(name="parent_id", type="Edm.String", filterable=True))
                    logger.info("Including title field in new index %s", self.search_info.index_name)
                    fields.append(SimpleField(name="title", type="Edm.String", searchable=True, retrievable=True))
`


This is not strictly required for the vectorizer to work but the indexer (defined in /app/backend/prepdocslib/integratedvectorizerstrategy.py) needs the title field to work in order to show the PDF title in the search results.. 


'        indexer = SearchIndexer(
            name=indexer_name,
            description="Indexer to index documents and generate embeddings",
            skillset_name=f"{self.search_info.index_name}-skillset",
            target_index_name=self.search_info.index_name,
            data_source_name=f"{self.search_info.index_name}-blob",
            # Map the metadata_storage_name field to the title field in the index to display the PDF title in the search results
            field_mappings=[FieldMapping(source_field_name="metadata_storage_name", target_field_name="title")],
        )
'

PR #2440, opened 2025-03-19
Adjust container apps to scale to zero

### LINKED ISSUE #2436: Idle app is significantly more epensive than mentioned in cost calculator
I've deployed the demo and used it once or twice with a few queries, but the costs seem to be quite steep, even with zero usage.

Image

Expected monthly cost is ~180$ with few queries, while the pricing calculator mentions ~100$ for 30K semantic queries.

Image

Am I doing something wrong? Why is this costing so much with no use?
Thanks.

pamelafox:
## Purpose

Fixes #2436 

Thanks to the filed issue, I discovered that we were configuring Container Apps with a higher-than-default CPU core and memory, and a min replica of 1.
Unfortunately, we need the higher CPU and memory, due to our app's needs,
but the app still works fine with a min replica of 0 - it just means a slower start when it needs to scale back to 1 sometimes.

```
[X] Yes - If people are running this in production with these settings, it will reduce the performance of the app. I will note in release notes. I've also added a section to productionizing guide.
[ ] No
```

```
[ ] Yes
[X] No
```

## Type of change

```
[X] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## Code quality checklist

N/A, Bicep change

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ The repo includes sample data so it's ready to try end to end. In this sample ap
 Pricing varies per region and usage, so it isn't possible to predict exact costs for your usage.
 However, you can try the [Azure pricing calculator](https://azure.com/e/e3490de2372a4f9b909b0d032560e41b) for the resources below.
 
-- Azure Container Apps: Default host for app deployment as of 10/28/2024. See more details in [the ACA deployment guide](docs/azure_container_apps.md). Consumption plan with 0.5 CPU core, 1.0 GB RAM, minimum of 0 replicas. Pricing with Pay-as-You-Go. [Pricing](https://azure.microsoft.com/pricing/details/container-apps/)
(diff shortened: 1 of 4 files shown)


pamelafox:
Update: Our app needs more memory than the default. When deployed with these settings, we get errors like:

"2025-03-19T05:09:48.9567715Z stderr F 2025-03-19 05:09:48,952 - ERROR - Worker (pid:152) was sent SIGKILL! Perhaps out of memory?"

pamelafox:
I've updated this so that the only change is scaling down to 0 replicas.

mattgotteiner (approved):
LGTM. Maybe in the future we'll have more variables to control this

pamelafox:
@mattgotteiner True, we have a variable for the workload profile already, so min replicas would make sense as a parameter.. if any developer reading this comments wants that as an azd env variable, please file an issue or send a PR!

PR #2442, opened 2025-03-21
Add migration script for Azure Cosmos DB, old container to new container

pamelafox:
## Purpose

This adds migration script as requested by developer in the PR that introduced the new schema:
https://github.com/Azure-Samples/azure-search-openai-demo/pull/2312/files

Thanks @madebygps for pairing on this.

We structured it as a simple script, as opposed to an Azure Function, since developers should only need to run it once or perhaps twice. A developer can run it once while the old schema is running, switch over to the new schema, and then run it again. It should be idempotent.

```
[ ] Yes
[X] No
```

```
[ ] Yes
[X] No
```

## Type of change

```
[ ] Bugfix
[ ] Feature
[X] Other... Please describe:
```

## Code quality checklist

### DIFF
--- a/scripts/cosmosdb_migration.py
+++ b/scripts/cosmosdb_migration.py
@@ -15,6 +15,7 @@ entra_oid: str
 title: str
 timestamp: int
 type: str (always "session")
+version: str (always "cosmosdb-v2")
 
 For message_pair items:
 id: str
@@ -66,7 +67,6 @@ class CosmosDBMigrator:
         self.database = self.client.get_database_client(self.database_name)
         self.old_container = self.database.get_container_client("chat-history")
         self.new_container = self.database.get_container_client("chat-history-v2")
-        # check to see that both containers actually exist
         try:
             await self.old_container.read()
         except Exception:
@@ -83,14 +83,13 @@ class CosmosDBMigrator:
         if not self.client:
             await self.connect()
 
-        # Migration: read from old_container, transform, and insert into new_container
         query_results = self.old_container.query_items(query="SELECT * FROM c")
 
         item_migration_count = 0
         async for page in query_results.by_page():
             async for old_item in page:
                 batch_operations = []
-                # Build session item operation
+                # Build session item
                 session_item = {
                     "id": old_item["id"],
                     "version": "cosmosdb-v2",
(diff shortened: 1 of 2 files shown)


scripts/cosmosdb_migration.py
```diff
@@ -0,0 +1,162 @@
...
+        self.old_container = self.database.get_container_client("chat-history")
+        self.new_container = self.database.get_container_client("chat-history-v2")
+        try:
+            await self.old_container.read()
+        except Exception:
+            raise ValueError(f"Old container {self.old_container.id} does not exist")
+        try:
+            await self.new_container.read()
+        except Exception:
+            raise ValueError(f"New container {self.new_container.id} does not exist")
```
mattgotteiner, line 77: Does migration script need to create the container?

scripts/cosmosdb_migration.py
```diff
@@ -0,0 +1,162 @@
...
+    async def migrate(self):
+        """
+        Migrate data from old schema to new schema.
+        """
+        if not self.client:
+            await self.connect()
+        if not self.old_container or not self.new_container:
+            raise ValueError("Containers do not exist")
+
+        query_results = self.old_container.query_items(query="SELECT * FROM c")
```
mattgotteiner, line 88: We select * but we are only copying specific items. IMO either:
1. Fetch all columns and copy all columns
2. Only select the columns you want to copy

scripts/cosmosdb_migration.py
```diff
@@ -0,0 +1,162 @@
...
+                        "entra_oid": old_item["entra_oid"],
+                        "type": "message_pair",
+                        "question": question,
+                        "response": response,
+                        "order": idx,
+                        "timestamp": None,
+                    }
+                    batch_operations.append(("upsert", (message_pair,)))
+
+                # Execute the batch using partition key [entra_oid, session_id]
```
mattgotteiner, line 124: comment is wrong? we are using [entra_oid, id]

mattgotteiner (approved):

scripts/cosmosdb_migration.py
pamelafox: The old item represented the session, so old_item["id"] gives us the session_id

scripts/cosmosdb_migration.py
pamelafox: No, the Bicep creates the containers, it has all the right configuration for the partitions and hash and such.

scripts/cosmosdb_migration.py
pamelafox: I think we're basically using all the columns here, so I don't think there'd be a performance benefit to restricting the columns?

PR #2443, opened 2025-03-22
Port to gpt-4o-mini as default

pamelafox:
## Purpose

This PR upgrades to gpt-4o-mini as the default chat completion model.
See my analysis here:
https://blog.pamelafox.org/2025/03/gpt-4o-mini-vs-gpt-35-turbo-for-rag.html
The responses are longer, but it should be overall cheaper,
and it's a more recent model.

```
[X] Yes - With existing environments, it will keep gpt-35-turbo
[ ] No
```

```
[X] Yes - Maybe? It might mention the model
[ ] No
```

## Type of change

```
[ ] Bugfix
[ ] Feature
[X] Other... Please describe:
```

## Code quality checklist


evals/results/baseline/summary.json
```diff
@@ -2,26 +2,26 @@
     "gpt_groundedness": {
         "pass_count": 49,
         "pass_rate": 0.98,
-        "mean_rating": 4.94
```
mattgotteiner, line 5: We may want to add an additional set of prompts for each version of the model

pamelafox:
I have done a test azd up with my old environment and a brand new environment, all seems well. Merging.

Jbrocket:
Default azureaideploymentregion is canadaeast (https://github.com/Azure-Samples/azure-search-openai-demo/blob/c5bfb225378d2322bcf9581c2dc7b377e78b30a5/infra/main.bicep#L80), which doesn't support gpt-4o-mini. https://learn.microsoft.com/en-us/azure/ai-services/openai/concepts/models?tabs=global-standard%2Cstandard-chat-completions#assistants-preview

It doesn't make sense to default to a region that won't work.

pamelafox:
@Jbrocket Ah! Great catch, I'll send a fix for that now.

PR #2533, opened 2025-05-12
Updates the baseline evals with embedding 3 large, renames other folders for clarity

pamelafox:
## Purpose

This PR updates the "baseline" folder to use embedding-3-large, since that's the new default for the repo. Along with that, this PR renames the old baseline to "gpt4omini-ada002" and renames other folders for consistency.

The other option is to not have a baseline folder, and just always use descriptive folder names. That wouldn't be compatible with the current evaluation workflow however, as it assumes there's always a baseline folder to diff against. I wish I could figure out how to symlink in git, but I haven't figured out a way, since then I would just symlink baseline to the most recent folder.

```
[ ] Yes
[X] No
```

```
[ ] Yes
[X] No
```

## Type of change

```
[ ] Bugfix
[ ] Feature
[X] Other... Please describe:
```

## Code quality checklist

N/A


evals/results/baseline/config.json
```diff
@@ -1,6 +1,6 @@
 {
     "testdata_path": "ground_truth.jsonl",
-    "results_dir": "results/gpt-4o-mini",
+    "results_dir": "results/experiment<TIMESTAMP>",
```
mattgotteiner, line 3: does this get replaced automatically?

mattgotteiner (approved):

evals/results/baseline/config.json
pamelafox: Oh yes when I ran this eval, I didnt bother changing the results_dir, so thats why we see it changed in the config back to the default configuration. I renamed the folder after the fact instead. I could pretend that I had specified a folder name, but this config.json does actually reflect how I run it, so its true.

pamelafox: And yes, the evaluator replaces <TIMESTAMP> with current timestamp, so you dont accidentally override a past evaluation, if you forget to change the dir. I usually use that approach to be safe.

Issue #2675, opened 2025-08-13
Broken Authentication Due to MSAL Library Upgrade


ckabalan:
Looks like aa06a96 (Fix Dependabot MSAL package upgrade by updating compatible versions (https://github.com/Azure-Samples/azure-search-openai-demo/pull/2632)) introduced an issue with authentication. When I update a deployed environment with this updated library or run the development server I get:

```
Unexpected Application Error!
uninitialized_public_client_application: You must call and await the initialize function before attempting to call any other MSAL API. For more visit: aka.ms/msaljs/browser-errors
```

According to the msal docs, and the link above this was an update they made to the library which requires an initialize phase. The relevant docs are here: [AzureAD/microsoft-authentication-library-for-js - Errors - uninitialized_public_client_application](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-browser/docs/errors.md#uninitialized_public_client_application)

I've taken the `2025-08-07` code and rolled back the library versions to the original versions (3.26.1 and 2.2.0) and the issue resolved immediately, rolled it back forward (4.16.0 and 3.0.16) and it broke again.

I'm not an authentication or security expert, but [cline](https://github.com/cline/cline) recommended:

> A minimal fix is to keep your single PublicClientApplication in layoutWrapper.tsx but defer all MSAL API calls until after you’ve called and awaited initialize(). Here’s the outline:
> 
> 1. Inside layoutWrapper.tsx, after creating msalInstance, add a new state flag: • `const [initialized, setInitialized] = useState(false);`
> 
> 2. Replace your top-level API calls with a single useEffect that does:
> 
>    ```ts
>    useEffect(() => {
>      const init = async () => {
>        await msalInstance.initialize();                 // ← required
>        // Account fallback
>        if (!msalInstance.getActiveAccount() &&
>            msalInstance.getAllAccounts().length > 0) {
>          msalInstance.setActiveAccount(
>            msalInstance.getAllAccounts()[0]
>          );
>        }
>        // Event callback
>        msalInstance.addEventCallback(…);
>        // Check and set loggedIn
>        setLoggedIn(await checkLoggedIn(msalInstance));
>        setInitialized(true);
>      };
>      init();
>    }, []);
>    ```
> 
> 3. In your render, show nothing or a spinner while `!initialized`, then once true wrap in `<MsalProvider instance={msalInstance}>…`.
> 
[...]

PR #2685, opened 2025-08-18
Initialize MSAL before use to fix auth regression

pamelafox:
## Purpose

Fix MSAL auth regression after the MSAL 4.x upgrade that caused "uninitialized_public_client_application" when MSAL APIs were called before initialization. This change:

- Creates a stable PublicClientApplication instance and awaits initialize() before use
- Adds exception handling and console logging to initialization
- Adds a tiny loading placeholder during initialization so the app still renders

Fixes Azure-Samples/azure-search-openai-demo#2675.

```text
[ ] Yes
[x] No
```

```text
[ ] Yes
[x] No
```

## Type of change

```text
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## Code quality checklist


app/frontend/src/layoutWrapper.tsx
```diff
@@ -1,45 +1,75 @@
...
 import { MsalProvider } from "@azure/msal-react";
 import { LoginContext } from "./loginContext";
 import Layout from "./pages/layout/Layout";
 
 const LayoutWrapper = () => {
     const [loggedIn, setLoggedIn] = useState(false);
     if (useLogin) {
-        var msalInstance = new PublicClientApplication(msalConfig);
+        // Create a stable MSAL instance (avoid re-init/duplicate listeners; single shared client for MsalProvider).
+        const msalInstance = useMemo(() => new PublicClientApplication(msalConfig), []);
```
pamelafox: Cline's suggested fix did not use useMemo, but Copilot+GPT5 thought it was a good idea, to avoid re-construction across renders.

app/frontend/src/layoutWrapper.tsx
```diff
@@ -1,45 +1,75 @@
...
+        const [initialized, setInitialized] = useState(false);
+        // Track mount state so we don't call setState after unmount if async init resolves late
+        const mounted = useRef<boolean>(true);
```
pamelafox, line 12: The new mounted ref is to avoid setState calls inside the async init() below if the component unmounts during all that. Don't know how common that is.

app/frontend/src/layoutWrapper.tsx
```diff
@@ -1,45 +1,75 @@
...
-        // Listen for sign-in event and set active account
-        msalInstance.addEventCallback(event => {
-            if (event.eventType === EventType.LOGIN_SUCCESS && event.payload) {
-                const account = event.payload as AccountInfo;
-                msalInstance.setActiveAccount(account);
-            }
-        });
+                    // Default to using the first account if no account is active on page load
+                    if (!msalInstance.getActiveAccount() && msalInstance.getAllAccounts().length > 0) {
+                        msalInstance.setActiveAccount(msalInstance.getAllAccounts()[0]);
```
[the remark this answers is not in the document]
pamelafox: Added original comment back

app/frontend/src/layoutWrapper.tsx
DanWahlin: The [docs](https://learn.microsoft.com/en-us/entra/msal/javascript/react/getting-started) show this general type of pattern for accomplishing that (without useMemo()), but useMemo() should achieve the goal of ensuring it's only created once. I don't know enough about the app to understand the overall flow and how often the component is rerendered. But, assuming it is, useMemo should help there.

mattgotteiner: It's an SPA - my understanding is that this wrapper holds all the content displayed and shouldn't be re-rendered unless the page is refreshed

mattgotteiner (approved):
LGTM

Recommend switching to singleton pattern to make it easier

https://learn.microsoft.com/en-us/entra/msal/javascript/react/getting-started#initialization

pamelafox:
@DanWahlin @mattgotteiner I've revised the PR to use an approach closer to the documentation, with the MsalProvider moved into index.tsx, and then Layout uses useMsal.

DanWahlin:
I like that approach better. Eliminates any rerenders which would get rid of the singleton challenge.

PR #2770, opened 2025-10-16
Add 'australiaeast' to allowed regions

kachihro:
## Purpose
We need to have this - within the allowed Locations - for DocumentIntelligence

```
[ ] Yes
[ x ] No
```

```
[ ] Yes
[ x ] No
```

## Type of change

```
[ x ] Bugfix
[ ] Feature
[ ] Other... Please describe:
```


pamelafox:
Is it available in many more regions now? That's what this claims:
https://learn.microsoft.com/en-us/azure/ai-services/document-intelligence/model-overview?view=doc-intel-4.0.0#regional-availability

But I'm not sure if our use of the new figures mode is limited to some regions, I haven't found a good updated list of regions.
Anyway I will trust that you verified australiaeast and merge this, thank you!

PR #2771, opened 2025-10-20
Switch from security filter to built-in ACL enforcement

mattgotteiner:
## Purpose

* Use the built-in document access control from Azure Search instead of manual security filters

Note that global document access won't work the same, it will now work as the built-in access control does

```
[ ] Yes
[X] No
```

```
[ ] Yes
[X] No
```

## Type of change

```
[ ] Bugfix
[X] Feature
[ ] Other... Please describe:
```

## Code quality checklist


mattgotteiner:
need to update docs

mattgotteiner:
need to update UX to remove security filter options

app/backend/core/authentication.py
```diff
@@ -32,7 +32,7 @@ def __str__(self) -> str:
 
 
 class AuthenticationHelper:
-    scope: str = "https://graph.microsoft.com/.default"
+    scope: str = "https://search.azure.com/.default"
```
pamelafox, line 35: This change may make it slightly harder for people to extend the sample for arbitrary Graph API access, like to read emails, as I once demo'ed. Not sure who's actually doing that though. Just noting.

app/backend/approaches/approach.py
```diff
@@ -292,7 +290,8 @@ async def run_agentic_retrieval(
                         filter_add_on=filter_add_on,
                     )
                 ],
-            )
+            ),
+            x_ms_query_source_authorization=access_token,
```
pamelafox, line 294: Note to self: Ensure that tests were added to verify that token is passed in for all calls to search()/retrieve()

app/backend/core/authentication.py
```diff
@@ -138,128 +135,46 @@ def get_token_auth_header(headers: dict) -> str:
 
         raise AuthError(error="Authorization header is expected", status_code=401)
 
-    def build_security_filters(self, overrides: dict[str, Any], auth_claims: dict[str, Any]):
```
pamelafox, line 141: Yay code removal!

app/backend/core/authentication.py
```diff
@@ -138,128 +135,46 @@ def get_token_auth_header(headers: dict) -> str:
...
-
-        return groups
-
     async def get_auth_claims_if_enabled(self, headers: dict) -> dict[str, Any]:
         if not self.use_authentication:
             return {}
         try:
             # Read the authentication token from the authorization header and exchange it using the On Behalf Of Flow
-            # The scope is set to the Microsoft Graph API, which may need to be called for more authorization information
+            # The scope is set to Azure Search for authentication
```
pamelafox, line 143: @mattgotteiner Is there an Azure AI Search documentation about the ACLs? Might be good place to link it

app/backend/core/authentication.py
```diff
@@ -138,128 +135,46 @@ def get_token_auth_header(headers: dict) -> str:
...
-                auth_claims["groups"] = await AuthenticationHelper.list_groups(graph_resource_access_token)
+            if "error" in search_resource_access_token:
+                raise AuthError(error=str(search_resource_access_token), status_code=401)
+
+            id_token_claims = search_resource_access_token["id_token_claims"]
+            auth_claims = {"oid": id_token_claims["oid"]}
+            # Only pass on the access token if access control is required
+            if self.require_access_control:
+                access_token = search_resource_access_token["access_token"]
+                auth_claims["access_token"] = access_token
```
pamelafox, line 163: How come we don't need groups anymore?

app/backend/core/authentication.py
```diff
@@ -138,128 +135,46 @@ def get_token_auth_header(headers: dict) -> str:
...
             return auth_claims
         except AuthError as e:
             logging.exception("Exception getting authorization information - " + json.dumps(e.error))
-            if self.require_access_control and not self.enable_unauthenticated_access:
+            if not self.enable_unauthenticated_access:
```
pamelafox, line 167: Can you remind me why this check can be simplified?

app/backend/core/authentication.py
```diff
@@ -272,11 +187,13 @@ async def check_path_auth(self, path: str, auth_claims: dict[str, Any], search_c
         # Replace ' with '' to escape the single quote for the filter
         # https://learn.microsoft.com/azure/search/query-odata-filter-orderby-syntax#escaping-special-characters-in-string-constants
         path_for_filter = path.replace("'", "''")
-        filter = f"{security_filter} and ((sourcefile eq '{path_for_filter}') or (sourcepage eq '{path_for_filter}'))"
+        filter = f"(sourcefile eq '{path_for_filter}') or (sourcepage eq '{path_for_filter}')"
 
         # If the filter returns any results, the user is allowed to access the document
         # Otherwise, access is denied
-        results = await search_client.search(search_text="*", top=1, filter=filter)
+        results = await search_client.search(
```
pamelafox, line 195: Should we do auth_claims["access_token"] to ensure that the token exists?

app/backend/prepdocslib/filestrategy.py
```diff
@@ -69,6 +69,7 @@ def __init__(
         category: Optional[str] = None,
         use_content_understanding: bool = False,
         content_understanding_endpoint: Optional[str] = None,
+        require_access_control: bool = False,
```
pamelafox: @mattgotteiner Did you want to rename this to match the env var or nah?

app/backend/prepdocslib/searchmanager.py
```diff
@@ -433,6 +446,29 @@ async def create_index(self):
                             self.search_info,
                         )
 
+                if self.use_acls:
+                    logger.info("Enabling permission filtering on index %s", self.search_info.index_name)
```
pamelafox: This log is confusing since it sometimes disables it, per the next line

docs/login_and_acl.md
```diff
@@ -86,14 +88,14 @@ The easiest way to setup the two apps is to use the `azd` CLI. We've written scr
...
-  To allow users to search on documents that have no access controls assigned, even when access control is required, run the following command:
+  To allow upload of documents that have global access when there are no document-specific access controls assigned, run the following command:
 
     ```shell
     azd env set AZURE_ENABLE_GLOBAL_DOCUMENT_ACCESS true
     ```
 
 1. (Optional) **Allow unauthenticated access**
-  To allow unauthenticated users to use the app, even when access control is enforced, run the following command:
+  To allow unauthenticated users to use the app, run the following command:
```
pamelafox, line 98: You changed that comment, so does that mean access control is not enforced in that case?

docs/login_and_acl.md
```diff
@@ -146,11 +148,17 @@ The following instructions explain how to setup the two apps using the Azure Por
...
-- Select **API Permissions** in the left hand menu. By default, the [delegated `User.Read`](https://learn.microsoft.com/graph/permissions-reference#user-permissions) permission should be present. This permission is required to read the signed-in user's profile to get the security information used for document level access control. If this permission is not present, it needs to be added to the application.
+- Select **API Permissions** in the left hand menu. By default, the [delegated `User.Read`](https://learn.microsoft.com/graph/permissions-reference#user-permissions) permission should be present. This permission is required to read the signed-in user's profile.
   - Select **Add a permission**, and then **Microsoft Graph**.
   - Select **Delegated permissions**.
   - Search for and and select `User.Read`.
   - Select **Add permissions**.
+- Select **API Permissions** in the left hand menu. The server app will use the `user_impersonation` permission from Azure AI Search to issue a token for security filtering on behalf of the logged in user.
+  - Select **Add a permission**, and then **APIs my organization uses**.
+  - Search for and select **Azure Cognitive Search**.
```
pamelafox, line 158: So it never got renamed to AI Search there aye?

docs/login_and_acl.md
```diff
@@ -335,14 +340,25 @@ To run this script with a Data Lake Storage Gen2 account, first set the followin
...
-- `AZURE_ENABLE_UNAUTHENTICATED_ACCESS`: Allows unauthenticated users to access the chat app, even when `AZURE_ENFORCE_ACCESS_CONTROL` is enabled. `AZURE_ENABLE_GLOBAL_DOCUMENT_ACCESS` should be set to true to allow unauthenticated users to search on documents that have no access control assigned. Unauthenticated users cannot search on documents with access control assigned.
+- `AZURE_ENABLE_GLOBAL_DOCUMENT_ACCESS`: Enables prepdocs upload code to support setting user ids and group ids to `["all"]` when uploading documents that have no access control assigned. This will enable the built-in document level access control to return these documents if `AZURE_ENFORCE_ACCESS_CONTROL` is enabled. If you are migrating from a previous version where this was not required, you'll have to perform a [one-time migration](#migrate-to-built-in-document-access-control) to enable global document access.
+- `AZURE_ENABLE_UNAUTHENTICATED_ACCESS`: Allows unauthenticated users to access the chat app. If `AZURE_ENFORCE_ACCESS_CONTROL` is enabled, unauthenticated users cannot search on documents.
```
pamelafox, line 361: So basically you'd use it when you want a login but dont want ACLs? Have you checked this is working correctly with the user upload feature?

scripts/auth_init.py
```diff
@@ -129,7 +136,14 @@ def server_app_permission_setup(server_app_id: str) -> Application:
                     # Graph profile
                     ResourceAccess(id=uuid.UUID("{14dad69e-099b-42c9-810b-d002981feec1}"), type="Scope"),
                 ],
-            )
+            ),
+            RequiredResourceAccess(
+                resource_app_id="880da380-985e-4198-81b9-e05b1cc53158",
```
pamelafox, line 143: Add a comment specifying that its the Azure Search resource ID?

scripts/auth_init.py
```diff
@@ -177,6 +191,87 @@ def server_app_known_client_application(client_app_id: str) -> Application:
...
+    principal_id: str
+    resource_app_id: str
+    scopes: list[str]
+    target_label: str
+
+    def scope_string(self) -> str:
+        return " ".join(self.scopes)
+
+
+async def grant_application_admin_consent(graph_client: GraphServiceClient, client_app_id: str, server_app_id: str):
```
pamelafox, line 209: Can you link to the docs about this that we were looking at?

scripts/auth_update.py
```diff
@@ -9,9 +9,11 @@
 from msgraph.generated.models.web_application import WebApplication
 
 from auth_common import get_application, test_authentication_enabled
+from load_azd_env import load_azd_env
 
 
 async def main():
+    load_azd_env()
```
pamelafox, line 16: Hm how come I didn't need that before?

tests/snapshots/test_app/test_chat_vision_user/auth_client0/result.json
```diff
@@ -8,8 +8,8 @@
                 "westbrae_jun28.pdf#page=2"
             ],
             "images": [
-                "data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAADUlEQVR42mP8z/C/HgAGgwJ/lK3Q6wAAAABJRU5ErkJggg==",
-                "data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAADUlEQVR42mP8z/C/HgAGgwJ/lK3Q6wAAAABJRU5ErkJggg=="
+                "data:image/png;base64,dGVzdCBjb250ZW50",
```
pamelafox: Hm do you know why these changed?

app/backend/core/authentication.py
mattgotteiner: I see - unfortunately, we cannot request both scopes here. We can give guidance to do this twice though in the custom code

app/backend/approaches/approach.py
mattgotteiner: I do believe i wrote tests for this

e.g. https://github.com/Azure-Samples/azure-search-openai-demo/pull/2771/files#diff-a67cb1853203a6f1956991a9d9881d231c4d43557f5baffd45cc672a87e41cc6R205

app/backend/core/authentication.py
mattgotteiner: will add a doc link now

app/backend/core/authentication.py
mattgotteiner: taken care of by enforcement

https://learn.microsoft.com/en-us/azure/search/search-query-access-control-rbac-enforcement

app/backend/core/authentication.py
mattgotteiner: We should fail here regardless of if we require access control or not
The only reason we may succeed is if unauthenticated access was enabled

app/backend/core/authentication.py
mattgotteiner: yes, good idea. we already check for require_access_control enabled at the top so this should not crash

app/backend/prepdocslib/filestrategy.py
mattgotteiner: yes, good idea

app/backend/prepdocslib/searchmanager.py
mattgotteiner: will update to "Updating" instead of "Enabling

mattgotteiner: actulaly, let's log specifically what we did, good idea

docs/login_and_acl.md
mattgotteiner: if a user is not authenticated and access control is enabled, there's no way for them to see the global files anymore, we still need to pass in that access token.
You can put up a login screen though without access control, so the comment is updated here, correct

docs/login_and_acl.md
mattgotteiner: This Entra app didn't change, correct

docs/login_and_acl.md
mattgotteiner: Should we tie user upload to requiring acls?

scripts/auth_update.py
mattgotteiner: I think it was an oversight as we only tested this in the "full azd up" scenario where we loaded the env vars

tests/snapshots/test_app/test_chat_vision_user/auth_client0/result.json
mattgotteiner: no, good question, we'll have to look more deeply

docs/login_and_acl.md
mattgotteiner: done

tests/snapshots/test_app/test_chat_vision_user/auth_client0/result.json
mattgotteiner: codex figured it out - we were passing the literal "test content" through our blob transport, we are now wiring up PNG bytes. this also changed the output of a few more tests (for the better!)

PR #2846, opened 2025-11-26
Increase embedding TPM capacity and add note in cloud ingestion guide

pamelafox:
## Purpose

Some developers have tried out the cloud ingestion approach already and ran into errors due to rate limits with the embedding model. This PR increases the embedding TPM default capacity to 200 and adds a note to the ingestion guide about how to increase even farther. It seems that this is a safe capacity to request, as Azure OpenAI quotas seem to default quite high for embedding models, at least in my account.

```
[ ] Yes
[X] No
```

```
[ ] Yes
[X] No
```

## Type of change

```
[ ] Bugfix
[ ] Feature
[X] Other... Please describe:
```

## Code quality checklist

### DIFF
--- a/.github/workflows/lint-markdown.yml
+++ b/.github/workflows/lint-markdown.yml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout repo
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
       - name: Run markdownlint-cli2
         uses: DavidAnson/markdownlint-cli2-action@v21
         with:
--- a/.github/workflows/azure-dev-validation.yaml
+++ b/.github/workflows/azure-dev-validation.yaml
@@ -17,7 +17,7 @@ jobs:
       security-events: write
     steps:
       - name: Checkout
-        uses: actions/checkout@v5
+        uses: actions/checkout@v6
 
       - name: Build Bicep for linting
         uses: azure/CLI@v2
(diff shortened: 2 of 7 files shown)


.github/workflows/lint-markdown.yml
```diff
@@ -14,9 +14,11 @@ jobs:
     steps:
       - name: Checkout repo
         uses: actions/checkout@v5
-      - name: Run markdownlint
-        uses: articulate/actions-markdownlint@v1
+      - name: Run markdownlint-cli2
+        uses: DavidAnson/markdownlint-cli2-action@v21
```
pamelafox, line 18: We suddenly started getting markdownlint errors, and in debugging, I realized that our markdownlint action was deprecated in favor of this one. David Anson also authors the VS Code extension that we recommend in the repo configuration, so this makes CI consistent with VS Code errors, in theory.

mattgotteiner: sounds good

.github/workflows/lint-markdown.yml
```diff
@@ -14,9 +14,11 @@ jobs:
...
         with:
-          config: .github/workflows/markdownlint-config.json
-          files: '**/*.md'
-          ignore: data/
+          config: .markdownlint-cli2.jsonc
+          globs: |
+            **/*.md
+            !data/**
+            !.github/**
```
mattgotteiner, line 24: ;)

pamelafox: The default markdown files like SECURITY.md are riddled with issues, and we seemed to ignore them before, so I ignored them again here. Could fix em up in future.

PR #2868, opened 2025-12-10
Allow for EML and MSG parsing / also check+replace if fileName has "#" character

kachihro:
## Purpose
We had a requirement to ingest EMAIL files, EML and MSG.   This required a new 'parser' - as well as a new component to display the email, when selected.

Another issue was the "#" character in some fileNames - they would ingest OK, but couldn't use the HREF due to the # within the fileName/Url - rather than DOC.PDF#page1

This PR includes an additional package for the 'extract-msg' - and some CSS tweaks also (for showing MAIL)

Have been able to deploy via "AZD DEPLOY" and the processing of EML/MSG files is working - and can view them after a chat or ask response.    ✅

[ X ] Yes
[ ] No
```

No changes to deployment instructions - perhaps need to include details that EML and MSG are supported types.

[ X ] Yes
[ ] No
```

## Type of change

[ X ] Bugfix - with fileName containing "#"
[ X ] Feature - able to ingest EML and MSG
[ ] Other... Please describe:
```

### DIFF
--- a/app/backend/prepdocslib/emailparser.py
+++ b/app/backend/prepdocslib/emailparser.py
@@ -1,0 +1,137 @@
+"""
+Email parser for MSG and EML files.
+Compatible with the prepdocs pipeline structure.
+"""
+import logging
+import email
+from email import policy
+from typing import AsyncGenerator, IO, Union, Optional
+import extract_msg
+from io import BytesIO
+import re
+
+from .page import Page
+
+logger = logging.getLogger(__name__)
+
+
+class EmailParser:
+    """Parser for EML email files."""
+    
+    async def parse(self, content: IO) -> AsyncGenerator[Page, None]:
+        """Parse EML file content."""
+        try:
+            # Read bytes from IO object
+            content_bytes = content.read()
+            # Parse email from bytes
+            msg = email.message_from_bytes(content_bytes, policy=policy.default)
+            
+            # Extract email metadata and body
+            text_parts = []
+            
+            # Add headers
+            text_parts.append(f"From: {msg.get('From', 'Unknown')}")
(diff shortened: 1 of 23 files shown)

### CODE REVIEW

File: scripts/prepdocs.ps1 [hunk 1]
```diff
@@ -26,4 +26,8 @@ $argumentList = "./app/backend/prepdocs.py $dataArg --verbose $additionalArgs"
 
 $argumentList
 
-Start-Process -FilePath $venvPythonPath -ArgumentList $argumentList -Wait -NoNewWindow
+$process = Start-Process -FilePath $venvPythonPath -ArgumentList $argumentList -Wait -NoNewWindow -PassThru
```
pamelafox, line 29: Are these changes orthogonal to the EML change?

app/backend/prepdocslib/integratedvectorizerstrategy.py
```diff
@@ -170,6 +170,28 @@ async def run(self):
             files = self.list_file_strategy.list()
             async for file in files:
                 try:
+                     # Check if filename contains # and rename on disk if it does
```
pamelafox, line 173: We are going to deprecate integratedvectorizerstrategy.py in favor of cloud ingestion strategy only, so no changes should be needed to this file.

app/backend/prepdocslib/emailparser.py
```diff
@@ -0,0 +1,137 @@
+"""
+Email parser for MSG and EML files.
+Compatible with the prepdocs pipeline structure.
+"""
+import logging
+import email
+from email import policy
+from typing import AsyncGenerator, IO, Union, Optional
+import extract_msg
```
pamelafox, line 9: Please run pre-commit on all files to fix formatting, see CONTRIBUTING.md for instructions.

PR #2959, opened 2026-02-06
Fix ty type checking issues

pamelafox:
## Description

Fixes two ty type checking issues found after upgrading to ty 0.0.15:

- **blobmanager.py**: Remove unused `type: ignore[arg-type]` comment that is no longer needed
- **approach.py**: Add `type: ignore[no-matching-overload]` for the `**params` spread in `create()` call, and only add tools to params when not None

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)


pamelafox:
Fixed in separate PR

PR #3067, opened 2026-05-30
Fix locale key parity: remove stale i18n keys and add guard test

jluocsa:
Removes stale i18n keys that no longer exist in the English (en) base locale and adds a guard test to prevent future drift.

## Changes

- Removed helpTexts.llmInputs from 8 locales (no longer present in en).
- Removed labels.maxSubqueryCount and labels.vector.options.both from the pl locale.
- Added 	ests/test_frontend_locales.py to assert that every locale has key parity with en.

## Result

All 9 locales now pass key parity with en.


pamelafox:
Thank you, merging!

PR #3069, opened 2026-05-30
Heal LLM-simplified citations to canonical reference

### LINKED ISSUE #2950: LLM simplifies citations for filenames with special characters, breaking document links
## Description

When uploading documents via the user upload feature (with `USE_USER_UPLOAD=true` and `AZURE_ENFORCE_ACCESS_CONTROL=true`), the `sourcefile` field in the Azure AI Search index gets a `"- "` prefix that shouldn't be there.

## Steps to Reproduce

1. Enable user upload and access control:
   ```
   azd env set USE_USER_UPLOAD true
   azd env set AZURE_ENFORCE_ACCESS_CONTROL true
   azd provision && azd deploy
   ```
2. Log in and upload a document (e.g., `PyCon US 2025.pdf`)
3. Query the search index to inspect the indexed document

## Expected Behavior

The `sourcefile` field should match the original filename:
```
sourcefile: "PyCon US 2025.pdf"
```

## Actual Behavior

The `sourcefile` field has a `"- "` prefix:
```
sourcefile: "- PyCon US 2025.pdf"
```

## Impact

This causes the `/content/<path>` route to return 403 Forbidden when trying to view the document, because `check_path_auth()` filters by `sourcefile eq 'PyCon US 2025.pdf'` but the indexed value is `'- PyCon US 2025.pdf'`.

## Environment

- User upload enabled with ADLS Gen2
- Access control enforcement enabled
- Local development (may also affect deployed environments)

jluocsa:
## Purpose

Fixes #2950.

When a document filename contains leading special characters (e.g. a leading
hyphen, as produced for some user-uploaded files: `- PyCon US 2025.pdf`), the LLM
tends to "simplify" the citation it emits, dropping the leading characters. For
example, given the source `- PyCon US 2025.pdf#page=1`, the model writes
`[PyCon US 2025.pdf#page=1]` in its answer.

The frontend citation parser validated the cited text with `endsWith`, which
correctly recognized it as a valid citation, but then used the **simplified text**
as the citation reference. The resulting `/content/<path>` request used
`PyCon US 2025.pdf` instead of the indexed `- PyCon US 2025.pdf`, so
`check_path_auth()` filtered by the wrong `sourcefile` value and the document
returned 403/404.

This change resolves the cited text back to the **canonical** citation from
`data_points.citations` (exact match preferred, otherwise the `endsWith` match)
and uses that canonical value for the reference and for the
`citation_activity_details` lookup. Links and lookups now use the indexed value,
so the content route resolves correctly.

```
[ ] Yes
[x] No
```

Citations that already match exactly continue to resolve exactly (the exact match
is preferred). Only previously-broken simplified citations change behavior.

```
[ ] Yes
[x] No
```

## Type of change

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## Code quality checklist

- [x] `tsc` type-checks cleanly (`npx tsc --noEmit`).
- [x] `prettier` applied to the changed file.


jluocsa:
Updated this PR to implement the maintainer's suggested **Solution #1 (post-process the LLM response)** server-side, in addition to the original frontend safety net.

**What changed**
- `app/backend/approaches/approach.py` — added `heal_citation` / `heal_citations` helpers that map a simplified citation token (e.g. `[PyCon US 2025.pdf#page=1]`) back to the canonical indexed reference via exact match, else a *unique* `endswith` match. Ambiguous or no match leaves the token unchanged.
- `app/backend/approaches/chatreadretrieveread.py` — applies healing in both the non-streaming and streaming paths (including citations split across token deltas).
- `app/frontend/src/components/Answer/AnswerParser.tsx` — original frontend fix retained as a display-layer safety net.
- `tests/test_chatapproach.py` — 10 new tests covering exact/unique/ambiguous/no-match cases and a citation split across streaming deltas.

This means corrected citations are now present in the raw response itself (streaming, API consumers, exports), not just the rendered UI. Rebased onto current `main`; all healing tests pass and `ty` is clean.

PR #3072, opened 2026-05-31
docs: add setup verification commands

Zikeyyy:
## Summary

Adds a short checklist for verifying required local setup tools before continuing with the local environment setup.

## Changes

* Added commands to confirm Azure Developer CLI, Python, pip, Node.js, Git, and PowerShell are available from the terminal
* Added a Windows note about restarting the terminal after installing or updating tools so PATH changes are picked up

## Testing

Docs-only change. No application tests were run.

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,6 +140,18 @@ A related option is VS Code Dev Containers, which will open the project in your 
     - [Powershell 7+ (pwsh)](https://github.com/powershell/powershell) - For Windows users only.
       - **Important**: Ensure you can run `pwsh.exe` from a PowerShell terminal. If this fails, you likely need to upgrade PowerShell.
 
+    Before continuing, confirm the required tools are available from your terminal:
+
+    ```shell
+    azd version
+    python --version
+    pip --version
+    node --version
+    git --version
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -140,6 +140,18 @@ A related option is VS Code Dev Containers, which will open the project in your
     - [Powershell 7+ (pwsh)](https://github.com/powershell/powershell) - For Windows users only.
       - **Important**: Ensure you can run `pwsh.exe` from a PowerShell terminal. If this fails, you likely need to upgrade PowerShell.
 
+    Before continuing, confirm the required tools are available from your terminal:
+
+    ```shell
+    azd version
+    python --version
```
pamelafox, line 147: It is not as simple as this, as we use different commands for python based off the OS. For example, `python --version` actually fails on my current machine, but the repo works fine as it uses `python3` in the sh script.
And we have even other things we try for Windows.

pamelafox:
I appreciate the PR, but it's actually a bit tricky to do a simple check that works across operating systems, so I think it's better to rely on errors from the sh/ps1 scripts themselves.

PR #3161, opened 2026-07-10
Fix azd target environment loading in hooks

### LINKED ISSUE #3160: load_azd_env() ignores AZURE_ENV_NAME, loading the default azd env instead of the target env in hooks

```
- [ ] feature request
```

### Minimal steps to reproduce

The local `load_azd_env()` helper (`app/backend/load_azd_env.py`, duplicated in `scripts/load_azd_env.py`) resolves the azd environment by picking the entry flagged `IsDefault` from `azd env list -o json`. It never consults `AZURE_ENV_NAME`, even though azd sets `AZURE_ENV_NAME` in the environment of every provision/deploy hook and honors `-e <env>` for the command itself.

```python
for entry in env_json:
    if entry["IsDefault"]:
        env_file_path = entry["DotEnvPath"]
```

To reproduce:

1. Create two azd envs, e.g. `env-a` and `env-b`, and populate each with different `AZURE_STORAGE_ACCOUNT` / `AZURE_SEARCH_SERVICE` values.
2. Select `env-a` as the default: `azd env select env-a`.
3. Deploy the *other* env: `azd deploy -e env-b` (or `azd up -e env-b`).
4. The `postdeploy`/`postprovision` hooks (`prepdocs.py`, `setup_cloud_ingestion.py`) call `load_azd_env()`, which loads **`env-a`'s** `.env` (the default) rather than `env-b`'s, because it matches on `IsDefault` and ignores the `AZURE_ENV_NAME=env-b` that azd set for the hook.

This surfaced concretely while testing multiple envs across git worktrees that share a single `.azure` directory (one `defaultEnvironment`): a cloud-ingestion deploy loaded a *different* env's storage account and the Search managed identity failed with `ClientAuthenticationError` because it legitimately can't read the wrong storage account.

### Any log messages given by the failure

```
azure.core.exceptions.ClientAuthenticationError: ... (from the cloud-ingestion setup hook)
Loading azd env from /path/to/.azure/<WRONG-ENV>/.env
```

(The exact downstream error varies by which config value diverges between the two envs — storage auth, wrong Search service, etc.)

### Expected/desired behavior

When `AZURE_ENV_NAME` is set (as it always is inside azd hooks, and when a user passes `-e`), `load_azd_env()` should load **that** env's `.env`, falling back to the `IsDefault` env only when `AZURE_ENV_NAME` is unset (e.g. a manual local run). This matches azd's own resolution contract.

Notably, the PyPI package `dotenv_azd` (already used by `evals/*` via `dotenv-azd==0.3.0`) does not have this bug: it uses `azd env get-values`, which resolves the environment from `AZURE_ENV_NAME` and falls back to the default. Our two in-repo copies use `azd env list` + `IsDefault` instead.

Suggested fix (either):

1. **Preferred:** adopt `dotenv_azd` at all call sites and delete both local copies (`app/backend/load_azd_env.py`, `scripts/load_azd_env.py`), adding `dotenv-azd` to `app/backend/requirements.in`. Note the backend copy currently defaults to `override=True` (via `LOADING_MODE_FOR_AZD_ENV_VARS`) whereas `dotenv_azd` defaults to `override=False`, so call sites must pass `override=` to preserve existing semantics.
2. **Minimal:** in both local copies, prefer `AZURE_ENV_NAME` (match `entry["Name"]`) and fall back to `IsDefault` when it's unset.

### OS and Version?

macOS (observed); platform-independent.

### azd version?

Reproduces with current azd; not version-specific (the helper is app code, not azd itself).

### Versions

- `dotenv-azd==0.3.0` (present in `evals/requirements.txt` only; not in `app/backend/requirements.*`).

- Affected call sites (all use the buggy local copies): `app/backend/prepdocs.py`, `app/backend/setup_cloud_ingestion.py`, `app/backend/main.py`, and most `scripts/*` (`auth_init.py`, `auth_update.py`, `manageacl.py`, `manage_indexer.py`, `adlsgen2setup.py`, `cosmosdb_migration.py`, `verify_search_index_acls.py`).
- `evals/*` already import from `dotenv_azd` and are unaffected.
- In a normal single-env `azd up`, the deploying env *is* the default, so this doesn't bite typical users — it only affects multi-env / worktree workflows where a non-default env is deployed.

pamelafox:
## Purpose

Fixes #3160 by replacing the two duplicated local `load_azd_env()` helpers with [`dotenv-azd`](https://pypi.org/project/dotenv-azd/) 0.3.0, which delegates environment selection to `azd` instead of manually selecting the default environment.

### Why the package implementation works

<table>
<thead>
<tr>
<th>Deleted local implementation</th>
<th><code>dotenv-azd</code> 0.3.0 implementation</th>
</tr>
</thead>
<tbody>
<tr>
<td><pre><code>result = subprocess.run(
    "azd env list -o json",
    ...
)
for entry in json.loads(result.stdout):
    if entry["IsDefault"]:
        env_file_path = entry["DotEnvPath"]

load_dotenv(env_file_path, override=True)</code></pre></td>
<td><pre><code>result = run(
    ["azd", "env", "get-values"],
    ...
)
config = StringIO(result.stdout)

load_dotenv(
    stream=config,
    override=override,
)</code></pre></td>
</tr>
<tr>
<td>Always chooses the entry marked <code>IsDefault</code>, ignoring <code>AZURE_ENV_NAME</code> set by <code>azd up -e</code>, <code>azd deploy -e</code>, and azd hooks.</td>
<td><code>azd env get-values</code> performs azd's normal environment resolution, so it honors <code>AZURE_ENV_NAME</code> and falls back to the selected default when it is unset.</td>
</tr>
</tbody>
</table>

All backend, script, and eval call sites now import `load_azd_env` from `dotenv_azd` and consistently honor `LOADING_MODE_FOR_AZD_ENV_VARS`. The existing default remains `override`; setting the variable to `no-override` lets process environment variables take precedence over azd values everywhere.

```
[ ] Yes
[x] No
```

```
[ ] Yes
[x] No
```

## Type of change

```
[x] Bugfix
[ ] Feature
[ ] Other... Please describe:
```

## Code quality checklist

- [x] I ran `ty check` to check for type errors


app/backend/prepdocs.py
```diff
@@ -157,7 +157,7 @@ async def main(strategy: Strategy, setup_index: bool = True):
         # to avoid seeing the noisy INFO level logs from the Azure SDKs
         logger.setLevel(logging.DEBUG)
 
-    load_azd_env()
+    load_azd_env(override=os.getenv("LOADING_MODE_FOR_AZD_ENV_VARS") != "no-override")
```
mattgotteiner, line 160: nice :)

mattgotteiner (approved):

PR #3182, opened 2026-08-03
docs: note OpenAI client base_url for multi-model gateways

seven7763:
## Summary

Document that the OpenAI-compatible `OPENAI_BASE_URL` pattern works with this sample and with other OpenAI-compatible multi-model gateways.

DaoXE (`https://api.daoxe.com/v1`) is **one concrete example** among OpenAI-compatible gateways — no pricing/uptime claims.

Docs only — optional tip under the local OpenAI-compatible API section.

## Test plan
- [ ] Local dev docs still render
- [ ] Note is optional / non-breaking

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>


pamelafox:
This is likely an automated PR given the poor placement and the affiliate query args, closing.

